PR #66, opened 2017-01-25
Alert component

HHogg:
screen shot 2017-01-25 at 08 41 57


pmsorhaindo:
Will these be permanent until removed, or will we have temporary ones that are removed on setTimeout  too.

HHogg:
@pmsorhaindo Yup, adding and removing these should be handled outside of Axiom.

PR #68, opened 2017-01-25
Logo improvments

HHogg:


src/components/logo/LogoFlag.scss
```diff
@@ -0,0 +1,11 @@
+@import '../../design-patterns/colors/vars';
+@import '../../design-patterns/spacing/vars';
+
+$logo-flag-top-padding: $spacing-grid-size * 6;
+$logo-flag-padding: $spacing-grid-size * 4;
+
+.ax-logo-flag {
```
stefanpearson: A small quirk of CSS is that if you don't set the vertical-alignment, it throws the baseline off and does weird things, so might be worth putting a default of `vertical-align: top;` here

src/components/logo/LogoFlag.scss
```diff
@@ -0,0 +1,11 @@
...
+  display: inline-block;
+  padding: $logo-flag-top-padding $logo-flag-padding $logo-flag-padding;
```
stefanpearson: I'm not sure whether the padding should be absolute, regardless of scale. Perhaps this should be the containers responsibility to set this instead?

stefanpearson: Unless this is a one-off component… in which case, ignore

src/components/logo/LogoFlag.scss
HHogg: Thanks, good to know. Added 👍

src/components/logo/LogoFlag.scss
HHogg: Yeah, I was unsure about this, the original SVGs had this padding included so would scale with the logo... doing it this way is actually a bad idea if we wanted different sizes of the 'flag' logo (like at the bottom of analytics) ... rethinking about it I might work out the size and add it back into the SVG 👍

HHogg: Yup I agree, I'll move this to the login composite styles.

PR #76, opened 2017-02-13
IconIndicator

HHogg:
screen shot 2017-02-13 at 12 23 27
screen shot 2017-02-13 at 12 23 20
screen shot 2017-02-13 at 12 23 10

### DIFF
--- a/src/components/alert/example/icon.js
+++ b/src/components/alert/example/icon.js
@@ -1,0 +1,34 @@
+import React, { Component, PropTypes } from 'react';
+import { Example, Snippet } from 'style-guide';
+import { Alert, Strong } from 'bw-axiom';
+
+export default class AlertExample extends Component {
+  static propTypes = {
+    components: PropTypes.shape({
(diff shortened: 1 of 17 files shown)


src/components/alert/example/icon.js
```diff
@@ -0,0 +1,34 @@
...
+
+  render() {
+    const { components } = this.props;
+    const { Alert: { color } } = components;
+
+    return (
+      <Example name="Colored Alerts with IconIndicator">
+        <Snippet>
+          { color.values.map((color) =>
+            <Alert color={ color } key={ color } withIcon={ true }>
```
thatguynamedandy: On the example files I still get confused by the duplicate name...

`import { Alert, Strong } from 'bw-axiom';`
`const { Alert: { color } } = components;`

HHogg: This structure is generated by [react-docgen](https://github.com/reactjs/react-docgen). The duplication is kind of inherent because it's the properties that directly relates to that component.

PR #78, opened 2017-02-13
 Added Image fallback pattern

HHogg:
Added an Image fallback pattern where if the image fails to load it renders it's children.

### DIFF
--- a/src/components/image/example/falling-back.js
+++ b/src/components/image/example/falling-back.js
@@ -1,0 +1,25 @@
+import React, { Component } from 'react';
+import { Example, Snippet } from 'style-guide';
+import { Image, Paragraph } from 'bw-axiom';
+
+export default class ImageExample extends Component {
+  render() {
+    return (
+      <Example name="Base Image">
+        <Paragraph>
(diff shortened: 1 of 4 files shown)


src/components/image/example/falling-back.js
```diff
@@ -0,0 +1,25 @@
...
+          as many times as you like.
+        </Paragraph>
+
+        <Snippet>
+          <Image src="assets/WRONG.jpg">
+            <Image src="assets/WRONG_AGAIN.jpg">
+              <Image src="assets/axiom-bg.jpg" />
+            </Image>
+          </Image>
+        </Snippet>
```
thatguynamedandy: Not sure if it's necessary but a non image fallback example might help. I also wonder if we should provide some default functionality?

```js
<Snippet>
  <Image src="assets/WRONG.jpg">
    <p>Wrong image fool</p>
  </Image>
</Snippet>

HHogg: I'll add a Candytar example once it's merged :)

PR #84, opened 2017-02-16
feat(Progress): adds new Progress component

thatguynamedandy:
Closes https://github.com/BrandwatchLtd/axiom/projects/1#card-1407363

progress-screenshot


src/components/progress/ProgressFinite.js
```diff
@@ -0,0 +1,51 @@
...
+    const {
+      diameter,
+      innerRadius,
+      strokeWidth,
+      strokeDasharray,
+      strokeDashoffset,
+    } = getArcProperties({ percent });
+
+    return (
+      <RadialProgress className="ax-progress-finite" { ...rest }>
```
HHogg: Just need to place `{ ...rest }` at the start of the component to stop users defining their own `className` and overriding `className="ax-progress-finite"`

src/components/progress/ProgressInfinite.js
```diff
@@ -0,0 +1,53 @@
...
+    const {
+      diameter,
+      innerRadius,
+      strokeWidth,
+      strokeDasharray,
+      strokeDashoffset,
+    } = getArcProperties({ percent: 75 });
+
+    return (
+      <RadialProgress className={ classes } { ...this.props }>
```
HHogg: Same here.

HHogg: Also you need to `...rest` because of the `color` prop that shouldn't get passed to `RadialProgress`

src/components/progress/Progress.test.js
```diff
@@ -0,0 +1,33 @@
...
+  );
+}
+
+function matchSnapshot(props = {}) {
+  const component = getComponent(props);
+  const tree = component.toJSON();
+  expect(tree).toMatchSnapshot();
+}
+
+describe('ProgressFinite', () => {
```
HHogg: This should be `Progress`?

src/components/progress/__snapshots__/Progress.test.js.snap
```diff
@@ -0,0 +1,119 @@
...
+            "strokeDashoffset": 31.41592653589793,
+            "strokeWidth": 4,
+          }
+        } />
+    </g>
+  </svg>
+</div>
+`;
+
+exports[`ProgressFinite renders a success indicator 1`] = `
```
HHogg: I don't think this test is correct, it does not render the `<IconIndicator>` in the output template?

src/components/progress/__snapshots__/Progress.test.js.snap
```diff
@@ -0,0 +1,119 @@
...
+          }
+        } />
+    </g>
+  </svg>
+</div>
+`;
+
+exports[`ProgressFinite renders an infinite progress indicator 1`] = `
+<div
+  className="ax-radial-progress ax-radial-progress--small ax-progress-finite">
```
HHogg: I don't think this test is correct either, should it not render the infinite progress?

src/components/progress/Progress.test.js
thatguynamedandy: so should the top level import, hence the below failures.

`import { Progress } from 'bw-axiom';`

Copy and paste fail.

pmsorhaindo:
Crazy cool! 😄

PR #86, opened 2017-02-16
ImageCircle component

HHogg:


src/components/image/ImageCircle.scss
```diff
@@ -0,0 +1,70 @@
+@import '../../design-patterns/animations/mixins';
+@import '../../design-patterns/colors/vars';
+@import '../../design-patterns/spacing/mixins';
+@import '../vars';
+
+// Avatar Base
```
thatguynamedandy: Couldn't be any more minor, but is this comment relevant?

HHogg: It really isn't ... and it's incorrect too :D removing

PR #97, opened 2017-02-23
fix(ImageCircle): Fixes issue with image not sizing to circle

thatguynamedandy:


src/components/image/ImageCircle.js
```diff
@@ -24,7 +24,7 @@ export default class ImageCircle extends Component {
     return (
       <Base className={ classes } style={ style }>
         <div className="ax-image-circle__image">
-          <Image { ...rest } />
+          <Image { ...rest } style={ style } />
```
HHogg: This is squishing the images because of the set dimensions on the parent, the examples on the style guide show this. A set `width: 100%` on the <Image> will scale it up to the available space by the parent.

HHogg: I can't say I'm overly bothered in this situation of it being inline style object. 🤔

PR #105, opened 2017-02-24
Fiddly Card Spacing. 

HHogg:


HHogg:
Thanks @thatguynamedandy. It's the first time I've approached spacing like this, so we'll see how it goes. It will probably be revisited sometime in the future.

PR #107, opened 2017-02-28
feat(Table): Added Table component

siggysamson:
screen shot 2017-02-28 at 10 25 10


src/components/_vars.scss
```diff
@@ -15,10 +15,12 @@ $component-border-radius: 0.1875rem;
 $component-padding-vertical-small: $spacing-grid-size - $component-border-width-small;
 $component-padding-vertical-medium: ($spacing-grid-size * 2) - $component-border-width-small;
 $component-padding-vertical-large: ($spacing-grid-size * 3) - $component-border-width-small;
+$component-padding-vertical-huge: ($spacing-grid-size * 6) - $component-border-width-small;
 
 $component-padding-horizontal-small: $spacing-grid-size * 3;
 $component-padding-horizontal-medium: $spacing-grid-size * 3;
 $component-padding-horizontal-large: $spacing-grid-size * 4;
+$component-padding-horizontal-huge: $spacing-grid-size * 6;
```
siggysamson: Added `huge` padding options - not sure if a factor of 6 is ideal. This matches the designs provided

src/components/table/Table.scss
```diff
@@ -0,0 +1,42 @@
...
+  width: 100%;
+}
+
+.ax-table__cell {
+  border-bottom: $component-border-width-medium solid $color-ui-light-0;
+  padding-right: $component-padding-horizontal-huge;
+}
+
+.ax-table__head {
+  .ax-table__cell {
```
siggysamson: Nesting looks kinda odd. Another option would be to have a property on `TableCell` which defines it either as part of `head` or as a part of `body`. That would make it more explicit, currently I went for the implicit route because in order to get a proper markup you have to nest it like this anyways.

src/components/table/Table.scss
```diff
@@ -0,0 +1,42 @@
...
+  border-spacing: 0;
+}
+
+.ax-table__cell {
+  padding-right: $component-padding-horizontal-huge;
+  border-bottom: $component-border-width-medium solid $color-ui-light-0;
+}
+
+.ax-table__head {
+  .ax-table__cell {
```
siggysamson: Nesting looks kinda odd. Another option would be to have a property on `TableCell` which defines it either as part of `head` or as a part of `body`. That would make it more explicit, currently I went for the implicit route because in order to get a proper markup you have to nest it like this anyways.

src/components/table/Table.scss
```diff
@@ -0,0 +1,42 @@
...
+  .ax-table__cell:last-child {
+    padding-right: 0;
+  }
+}
+
+.ax-table__row--compact {
+  .ax-table__cell {
+    padding-top: 0;
+    padding-bottom: 0;
+    border: 0;
```
HHogg: Do you think this should be something at the Table level? I'm not sure there would be a mixture of compact styles within a single table?

Also I wonder if there should be separate variations to allow for 

* Full size - with borders
* Full size - without borders
* Compact - with borders
* Compact - without borders

So a table component with 

```jsx
<Table borders={ true|false } compact={ true|false } />
```

src/components/_vars.scss
HHogg: So the small, medium and large sizing are really for the components that are linked to this sizing patterns (they have a small/medium/large variation. 

For example `Button` has a small, medium and large variation. `Label` has just a small and medium. `TextInput` has just a medium and large. It keeps them all aligned up. 

The table cells are independent of this pattern so can live within the table module :)

PR #109, opened 2017-03-01
chore(Jest): Updated to version 19.0.2

siggysamson:
I've updated Jest to version 19.0.2.

Besides introducing an improved watch mode this includes a breaking change that is related to the Snapshot format: From now on snapshots have a specific version so that if you use a different version of Jest you don't accidentally run into false-positives etc.

See https://facebook.github.io/jest/blog/2017/02/21/jest-19-immersive-watch-mode-test-platform-improvements.html for more info about the update


package.json
```diff
@@ -31,6 +31,7 @@
   },
   "dependencies": {
     "classnames": "^2.2.1",
+    "jest": "^19.0.2",
```
HHogg: Should be a dev dependency 😬

PR #111, opened 2017-03-01
feat(Chart): Added ChartHeading component

HHogg:


HHogg:
@thatguynamedandy I do agree with you. I even think having this under the Header component opens it up for abuse... I think I'll move this into the chart module under a ChartLabel component instead 👍

PR #116, opened 2017-03-06
fix(Table): Allow passing additional classNames

siggysamson:


HHogg:
Hey @siggysamson 👋 I've been avoiding this as it's something that I thought might be quite beneficial. By not allow custom class names it helps keep the patterns as they are designed. It may also lead to users getting themselves into a specificity war.

Is there an additional variation that needs to be added? What's the use case here?

siggysamson:
Thanks for the explanation @HHogg! I actually wanted to add a comment to ask about this as I had the impression that custom class names shouldn't be allowed. We experienced this because we wanted to add a custom `qa` class to improve ATs. We are currently using data attributes but I'm happy to continue doing so.

Regarding `not allowing any other class names` - maybe there is a special use case behind this but I've found the following components where a class name currently is allowed:
- [/src/components/list/List.js](https://github.com/BrandwatchLtd/axiom/blob/bb3564b71a2c51b6dc3f7eec389d6b3f99c0d7e5/src/components/list/List.js)
- [/src/components/progress/RadialProgress.js](https://github.com/BrandwatchLtd/axiom/blob/667c52a1e858f9341cbd251187c0e168af66c9d2/src/components/progress/RadialProgress.js)

(I've ignored the ones under `/style-guide`)
We might want to consider updating those?

thatguynamedandy:
Hi @siggysamson, it looks like @HHogg beat me to responding on this one, and probably said it a bit clearer than me. 

On the AT use case, one pattern we have adopted is to use custom data attributes, so we use something like `<Alert data-tid="error-message" />`. This approach can keep the class property clean and is potentially less brittle, the only downside is that it's exposed to users, but I don't forsee that being a huge problem (it could be mitigated with a post build step for prod).

All of the Axiom classes spread the custom properties onto the component, so in theory no additional changes should be required. Looking to the future I was hoping each Axiom component could potentially have a base AT class which knows how to drive that specific component and could be shared across each project.

siggysamson:
Thanks @thatguynamedandy! I agree that using data attributes is fine for ATs. I don't see a need to remove them in the production build - ideally there aren't too many of them anyways.

Will close this PR :)

thatguynamedandy:
I've raised https://github.com/BrandwatchLtd/axiom/issues/117 to discuss wether Axiom should supply it's own automation hooks.

PR #118, opened 2017-03-09
Word cloud chart

pmsorhaindo:


src/charts/wordcloud/WordCloud.js
```diff
@@ -0,0 +1,50 @@
...
+    return (
+      <Base className="svg-container">
+        <svg
+            className="svg-content"
+            version="1.1"
+            viewBox={ `0 0 ${width} ${height}` }>
+
+          <g transform={ `translate(${translateX},${translateY})scale(${scaleX},${scaleY})` }>
+            { do {
+              words.map((options, key) => <Word key={ key } { ...options } />);
```
HHogg: You should let the user do the mapping themselves, because it might be the case they want to add `onClick` handlers or `onMouseOver` or `data-attributes` etc.... By hiding the `<Word>` mapping you can't account for the endless custom behaviour the user might want to add. So essentially, just change this to children. 

```
<g>
  { children }
</g>
```

src/charts/wordcloud/Word.js
```diff
@@ -0,0 +1,30 @@
...
+      y: PropTypes.number.isRequired,
+    }).isRequired,
+    text: PropTypes.string,
+    textAnchor: PropTypes.string,
+  };
+/* eslint-disable max-len */
+  render() {
+    const { color, font, fontSize, position, text, textAnchor } = this.props;
+    return (
+      <text style={ { fontSize: `${fontSize}`, fontFamily: `${font}`, fill: `${color}` } } textAnchor={ textAnchor } transform={ `translate(${position.x},${position.y})` }>
```
HHogg: You don't need to interpolate the strings if they are already strings 😉. Also if you break apart into a style object and break the lines you can get rid of the `eslint` comment. 

```jsx
const style = {
  fontSize, 
  fontFamily, 
  fill: color, 
  transform: `translate(${position.x},${position.y})`,
};

<text { ...rest }
    style={ style }  />
```

src/charts/wordcloud/Word.js
```diff
@@ -0,0 +1,30 @@
...
+  static propTypes = {
+    color: PropTypes.string, // TODO oneOf
+    font: PropTypes.oneOf([
+      'Roboto',
+      'Helvetica',
+      'Arial',
+      'sans-serif',
+    ]),
+    fontSize: PropTypes.string,
+    position: PropTypes.shape({
```
HHogg: Does this need to be an object, it could be flattened out to just `x="" y=""`

src/charts/wordcloud/WordCloud.js
```diff
@@ -0,0 +1,50 @@
...
+    } = this.props;
+
+    return (
+      <Base className="svg-container">
+        <svg
+            className="svg-content"
+            version="1.1"
+            viewBox={ `0 0 ${width} ${height}` }>
+
+          <g transform={ `translate(${translateX},${translateY})scale(${scaleX},${scaleY})` }>
```
HHogg: You could use `style` for this.

src/charts/wordcloud/WordCloud.js
```diff
@@ -0,0 +1,50 @@
...
+      translateX = 0,
+      translateY = 0,
+      width = 256,
+      words,
+    } = this.props;
+
+    return (
+      <Base className="svg-container">
+        <svg
+            className="svg-content"
```
HHogg: These classNames don't follow the Axiom standard, or match up with the classes in the CSS file?

src/charts/wordcloud/WordCloud.scss
```diff
@@ -0,0 +1,15 @@
+.ax-word-cloud {
+  display: inline-block;
```
HHogg: Unlikely something as big as a word cloud will be inline next to anything. Block would be better

src/charts/wordcloud/WordCloud.scss
```diff
@@ -0,0 +1,15 @@
...
+  padding-bottom: 100%; 
+  vertical-align: middle; 
+  overflow: hidden;
+}
+
+.ax-word-cloud__content {
+  display: inline-block;
+  position: absolute;
+  top: 0;
+  left: 0;
```
HHogg: Are any of these styles needed if the example is working with the incorrect classes?

HHogg:
There's also quite a few errors in the console. 
screen shot 2017-03-09 at 11 46 42

siggysamson:
Any specific reason the files were moved?
`src/components/chart/ChartPanel.scss → src/charts/chart-panel/ChartPanel.scss`

I think the component should still remain underneath `src/components`? Unless I'm missing something here

pmsorhaindo:
@siggysamson the reasoning was that from a design perspective chart panels are used to to display charts and this folder is grouping these chart components and any future chart components together.

PR #122, opened 2017-03-16
feat(Table): add hGutters prop so horizontal gutters can be adjusted

benji6:
This is my first Axiom PR! So if I've done anything wrong here or the commit message formatting is off please let me know! The change is required to meet design requirements for a table in Audiences. Thanks :D

### DIFF
--- a/src/components/table/Table.js
+++ b/src/components/table/Table.js
@@ -11,10 +11,12 @@ export default class Table extends Component {
     borders: PropTypes.bool,
     children: PropTypes.node,
     compact: PropTypes.bool,
+    hGutters: PropTypes.oneOf(['small', 'medium']),
   };
 
   static defaultProps = {
     borders: true,
+    hGutters: 'medium',
   };
 
   render() {
@@ -22,12 +24,14 @@ export default class Table extends Component {
       children,
       compact,
       borders,
+      hGutters,
       ...rest
     } = this.props;
 
     const classes = classnames('ax-table', {
       'ax-table--compact': compact,
       'ax-table--borders': borders === true,
+      [`ax-table--gutters-${hGutters}-h`]: true,
     });
 
     return (
--- a/src/components/table/Table.scss
+++ b/src/components/table/Table.scss
@@ -5,7 +5,8 @@
 $table-padding-vertical-small: $spacing-grid-size * 2;
 $table-padding-vertical-large: $spacing-grid-size * 5;
 
-$table-padding-horizontal: $spacing-grid-size * 8;
+$table-padding-horizontal-small: $spacing-grid-size * 3;
+$table-padding-horizontal-medium: $spacing-grid-size * 8;
 
 // Table styling
 .ax-table {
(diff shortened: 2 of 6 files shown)


src/components/table/example/index.js
```diff
@@ -1,4 +1,5 @@
 module.exports = [
   require('./table').default,
   require('./compact').default,
+  require('./hGutters').default,
```
thatguynamedandy: quite minor but I think the preferred convention is lowercase with dashes not camelcase, e.g. 'h-gutter'.
I would actually be tempted to just go with `gutters`, and vertical gutter changes could live in the same example.

src/components/table/Table.js
```diff
@@ -11,23 +11,27 @@ export default class Table extends Component {
     borders: PropTypes.bool,
     children: PropTypes.node,
     compact: PropTypes.bool,
+    hGutters: PropTypes.oneOf(['small', 'medium']),
```
thatguynamedandy: Is it worth putting a large in too? I know we don't currently need it but could be useful in the future if it doesn't look too wild.

src/components/table/Table.scss
```diff
@@ -5,16 +5,21 @@
...
 
 // Table styling
 .ax-table {
   width: 100%;
   border-spacing: 0;
 }
 
-.ax-table__cell {
-  padding-right: $table-padding-horizontal;
+.ax-table--hgutters-small .ax-table__cell {
```
thatguynamedandy: I found this convention in `ax-grid--gutters-none-h` https://github.com/BrandwatchLtd/axiom/blob/master/src/components/grid/Grid.js#L66

I wonder if similar would fir for this use case.. 
`ax-table__cell--gutters-small-h`
`ax-table__cell--gutters-medium-h`

src/components/table/Table.js
benji6: I don't think I'll bother with this as I have no idea what size large should be and it probably needs design input. I know what size the small gutters should be at least

src/components/table/example/index.js
benji6: Yeah that's pretty confusing actually because the components use PascalCase and the examples use kebab-case - will change the name to be consistent. I'll use `h-gutters` just to keep it specific

benji6: it's also pretty confusing because the prop is called `hGutters` not `h-gutters` but yeah will stick with convention

src/components/table/Table.scss
benji6: surely should be `ax-table--gutters-small-h` though as it targets the block not a child of it? but yeah I'll use the existing convention

src/components/table/Table.js
thatguynamedandy: no probs 👍

benji6:
@thatguynamedandy I've amended the commit with comments addressed

src/components/table/example/index.js
thatguynamedandy: yes, might go worth us cleaning up the conventions on that. Feel free to raise an issue to have it discussed 🤡

PR #123, opened 2017-03-16
feat(TableCell): add expand prop so table columns can be stretched

benji6:
This change is required to meet a design requirement on an axiom table. As far as I'm aware this is the way to achieve an expanded column in a table, but open to other suggestions if anyone has them. Thanks!

### DIFF
--- a/src/components/table/Table.scss
+++ b/src/components/table/Table.scss
@@ -17,6 +17,10 @@ $table-padding-horizontal: $spacing-grid-size * 8;
   padding-right: $table-padding-horizontal;
 }
 
+.ax-table__cell--grow {
+  width: 100%;
+}
+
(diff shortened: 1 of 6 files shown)


pmsorhaindo:
looks good, just unsure of whether the property name `expand` follows convention.

benji6:
yeah I'm not really sure what convention is on this. There is a `shrink` property in the Grid component so maybe the opposite is `expand`?

benji6:
Had a chat with @pmsorhaindo maybe we rename `expand` to `grow`? Then we have `shrink` and `grow` which is the same terminology used by flexbox. Thoughts?

pmsorhaindo:
That feels better to me. :)

PR #127, opened 2017-03-22
Table: extended column and row padding props

HHogg:


src/components/table/Table.js
```diff
@@ -10,28 +10,22 @@ export default class Table extends Component {
   static propTypes = {
     borders: PropTypes.bool,
     children: PropTypes.node,
-    compact: PropTypes.bool,
-    hGutters: PropTypes.oneOf(['small', 'medium']),
+    columPadding: PropTypes.oneOf(['none', 'small', 'medium', 'large']),
```
pmsorhaindo: `columnPadding` ?

pmsorhaindo:
LGTM otherwise.

src/components/table/Table.js
HHogg: Good spot 👍

PR #128, opened 2017-03-22
Design Session tweaks

HHogg:
Mostly minor colour and style changes here and there. Notable breaking change includes the removal of IconLink as it no longer makes sense to have it. Please see commit message.


src/components/button/Button.test.js
```diff
@@ -75,6 +75,30 @@ describe('Button', () => {
...
+    expect(tree).toMatchSnapshot();
+  });
+
+  it('renders circular size large', () => {
+    const component = getComponent({ circular: 'large' });
+    const tree = component.toJSON();
+    expect(tree).toMatchSnapshot();
+  });
+
+  it('renders circular size small', () => {
```
thatguynamedandy: tiny copy and paste error here.

HHogg: good spot 👍

PR #131, opened 2017-03-27
feat(Sass): Migrated over to CSS. See breaking changes

HHogg:
### Sass Removal 
**Variables are no longer available**: Where possible the old Sass variables have been mapped over to Custom Properties (CSS Variables). 

**Mixins are no longer available**: Likely `transition()` is the only concern, simply specify the transition properties with the CSS vars. 

### Structure shuffle
 **`/deisgn-patterns`**: Everything that Axiom provides are design patterns so the contents of this folder has been moved to `/materials`. Where files were originally in folders it has now become a flat structure. This only affects those targeting imports directly. _Note the folder still exists for the style guide this will change with a new style guide design/structure is decided_.

**`/components/vars`**: Have been moved into `/materials/sizing`

### Theming 
**PostCSS**: For IE11/Edge (currently) support you will need to PreProcess the theming variables. See README for more details 

**Theme Files**: Are now JSON files inside `/materials` 

### Global Variables 
**`__INCLUDE_CSS__`**: No longer needs defining in Webpack. However, if you want to skip the CSS files being imported you need to handle that. 

### Components

**`<Base>`**: No longer has a `theme` property, therefore all of the components have lost it (with the exceptions defined below. This is due to lack of CSS variable scoping in IE11/Edge. 

**`<TextInput>`**: The `style` property (which took a `translucent` value) has been removed, and ` theme` property has been added in it's place. 

**`<Context>`**: Retains it's `theme` prop but simply maps text color and background color.

**`<Word>`**: Now takes one of the brand color values instead of a raw color value, due to not being able to use the `getCSSVar` util.

### Composites 

**`<Login>`**: Retains it's `theme` prop to map the text color and translucent inputs colors correctly.


src/charts/wordcloud/Word.test.js
```diff
@@ -14,4 +14,76 @@ describe('Word', () => {
     const tree = component.toJSON();
     expect(tree).toMatchSnapshot();
   });
+
+  it('renders with color rose', () => {
+    const component = getComponent({ color: 'rose' });
+    const tree = component.toJSON();
+    expect(tree).toMatchSnapshot();
```
siggysamson: Having a test for each color makes the tests more explicit but this looks quite repetitive. Couldn't we do something like this:
```javascript
it('renders with color property', () => {
  const colors = ['rose', 'pink'];
  colors.forEach(color => {
    const component = getComponent({ color });
    expect(tree).toMatchSnapshot(`color ${color}`);
  });
});
```

src/components/badge/Badge.css
```diff
@@ -0,0 +1,62 @@
...
+.ax-badge--orange .ax-badge__background { background-color: var(--color-brand-orange); }
+.ax-badge--brown .ax-badge__background { background-color: var(--color-brand-brown); }
+.ax-badge--grey .ax-badge__background { background-color: var(--color-brand-grey); }
+
+.ax-badge--full {
+  width: 100%;
+}
+
+@media (--screen-less-than-small) {
+  .ax-badge--full--small { width: 100%; }
```
siggysamson: Might not be one for this PR but is that a "double" modifier? Shouldn't this be `.ax-badge--full.ax-badge--small`?

src/components/grid/Grid.test.js
```diff
@@ -127,6 +119,50 @@ describe('Grid', () => {
     });
   });
 
+  describe('hGutters', () => {
+    it('renders with value false', () => {
```
siggysamson: Same as above - maybe you could consider an array and loop over it instead of having a couple of separate `it`

README.md
```diff
@@ -16,83 +16,61 @@ yarn add bw-axiom
...
+};
 ```
-
 
 ### Using just the CSS
 
 The React components provide an abstraction of the templates and CSS, which greatly reduces any needed maintenance when CSS classes or the HTML structure to a component changes. They also encapsulate the behaviour of more interactive components, so components that do require Javascript will not work and will require you to plumb it all together.
 
-Just using the CSS is highly unrecommended but it is possible. The compiled CSS file is stored on the Brandwatch Google Cloud Platform CDN or target individual Sass files from the node_modules.
+The compiled CSS file is stored on the Brandwatch Google Cloud Platform CDN or target individual Sass files from the node_modules.
```
siggysamson: Looks like you already fixed that

src/charts/chart-heading/ChartHeading.css
```diff
@@ -1,14 +1,14 @@
...
-  font-weight: $font-weight-regular;
+  font-weight: var(--font-weight-regular);
   letter-spacing: 0.00125rem;
   line-height: 1rem;
   text-transform: uppercase;
 
   &.ax-text--strong,
-  .ax-text--strong {
-    font-weight: $font-weight-bold;
+  & .ax-text--strong {
```
siggysamson: This is CSS Nesting - http://tabatkins.github.io/specs/css-nesting/ Wasn't aware of that :)

HHogg: Yeah, good spot though! With CSS nesting the parent-child relationship needs explicitly defining with the `&` parent selector.

src/components/badge/Badge.css
HHogg: So the double modifier on here refers to where the `full` style is applied at a breakpoint. 

```css
/* This is the badge size */
.ax-badge--small { } 

/* This is the badge at full width */ 
.ax-badge--full { } 

/* This is the badge at full width at the small breakpoint */
.ax-badge--full--small { } 
```

Perhaps the classname is not that clear? It's a pattern throughout Axiom, which we can easily change without introducing any breaking changes, perhaps we should bring up some alternatives on a separate PR?

src/charts/wordcloud/Word.test.js
HHogg: Yeah that sounds good, will make it easier when adding more colors in the future. I had to change it slightly, because Jest needed unique `it` descriptions to test the snapshots so have went for 

```js
describe('renders with color', () => {
    [
      'rose',
      'pink',
      'purple',
      'lilac',
      'blue',
      'teal',
      'green',
      'chartreuse',
      'amber',
      'orange',
      'brown',
      'grey',
    ].forEach((color) => {
      it(color, () => {
        const component = getComponent({ color });
      });
    });
  });
```

HHogg: I'll also refactor the other tests to follow the same pattern.

src/charts/chart-panel/example/heading.js
```diff
@@ -1,17 +0,0 @@
-import React, { Component } from 'react';
```
pmsorhaindo: I should have removed this before, sorry! 😩

src/components/badge/BadgeGroup.css
```diff
@@ -0,0 +1,15 @@
+@import '../../materials/sizing';
+
+:root {
+  --badge-spacing: calc(var(--spacing-grid-size) * 2);
+}
+
+.ax-badge-group {
+  margin-bottom: -var(--badge-spacing);
```
pmsorhaindo: negating a these variables looks weird.. more `-`'s

thatguynamedandy: The negative sign threw me for a moment, not sure there is a clearer way though

src/components/base/Base.css
```diff
@@ -0,0 +1,179 @@
...
+    display: none !important;
+  }
+}
+/* stylelint-enable */
+
+.ax-text--left   { text-align: left; }
+.ax-text--center { text-align: center; }
+.ax-text--right  { text-align: right; }
+
+@media (--screen-more-than-small) {
```
pmsorhaindo: I'm probably wrong, but aren't these rules in the following media queries redundant?

src/components/badge/BadgeGroup.css
HHogg: I'm not sure agree it looks weird... an alternative would be to multiply them by -1 like 

```cs
calc(var(--foo) * -1);
```

but that seems a little unnecessary. How would you define them?

src/components/base/Base.css
HHogg: I don't think so. The classnames have breakpoint suffixes to differentiate them. 

```css
@media (--screen-more-than-small) { ...--small { } }
@media (--screen-more-than-medium) { ...--medium { } }
@media (--screen-more-than-large) { ...--large { } }
```

... or do you mean something else?

src/components/form/TextInput.css
```diff
@@ -0,0 +1,110 @@
+@import '../../materials/animations';
+@import '../../materials/colors';
+@import '../../materials/sizing';
+@import '../../materials/typography';
+@import './vars';
+
+/* stylelint-disable no-descending-specificity */
```
pmsorhaindo: It's not clear to me why the `ax-input__container` rules can't be applied after the `ax-input` wouldn't that rid the need for this lint dodge.

HHogg: It's not anything to do with the `ax-input__container` or `border-styles`... it was because of the specificity of the `&:-webkit-autofill` styles, which made sense to keep with the input styles. 

I guess the less inline linting overrides the better, so I've unnested them and moved them to the bottom of the file.

src/components/badge/BadgeGroup.css
HHogg: Not really, the only other alternative I can think of is 

```css
calc(var(--foo) * -1);
```

pmsorhaindo: No, I didn't have a suggestion. I agree, I also prefer `-var(...` over `...foo) * -1);`

src/components/base/Base.css
pmsorhaindo: I mean, doesn't the class `ax-text--right` provide the same rules that these breakpoint suffix'd rules apply?

HHogg: No, the breakpoint suffix'd ones apply themselves at the breakpoints they are defined in. 

```html
<span class="ax-text--center ax-text--right--medium"></span>
```

This will apply `text-align: right;` when `> medium-breakpoint` and `text-align: center;` all of the time.

src/components/context/ContextMenuItem.css
```diff
@@ -0,0 +1,27 @@
+@import '../../materials/animations';
+@import '../../materials/colors';
+@import '../../materials/sizing';
+
+.ax-context-menu__item {
+  display: block;
+  width: 100%;
+  padding: calc(var(--spacing-grid-size) * 3) calc(var(--spacing-grid-size) * 6);
```
stefanpearson: as you've broken down other properties, it would kind of make sense to split the compound padding/margins etc to make it easier to read, especially when there's css-next-bracket-soup™

src/components/form/ChedioButtox.js
```diff
@@ -1,10 +1,7 @@
 import React, { Component, PropTypes } from 'react';
 import classnames from 'classnames';
 import { Base } from 'bw-axiom';
-
-if (__INCLUDE_CSS__) {
-  require('./ChedioButtox.scss');
-}
+import './ChedioButtox.css';
```
stefanpearson: I would really like to find out more about this component

src/components/dialog/Dialog.css
```diff
@@ -1,45 +1,32 @@
...
-
-  &:first-child { padding-top: $page-padding-v; }
-  &:last-child  { padding-bottom: $page-padding-v; }
+  filter: drop-shadow(var(--drop-shadow));
 }
 
 .ax-dialog__header,
 .ax-dialog__footer {
   flex: 0 1 auto;
+  padding: var(--page-padding-v) var(--page-padding-h);
```
stefanpearson: I've noticed there are both `horizontal` and `h` in var names etc, this could end up being confusing?

src/components/form/vars.css
```diff
@@ -0,0 +1,17 @@
+@import '../../materials/colors';
+
+:root {
+  --input-border-color: var(--color-context-border);
+  --input-background-color: var(--color-context-background);
+  --input-placeholder-color: var(--color-context-text-disabled);
+
+  --input-disabled-background-color: var(--color-context-background-disabled);
```
stefanpearson: ok being _really_ pedantic here but it feels like the state should go at the end of the name, as per `--color-context-background-disabled`

src/components/tabset/Tabset.css
```diff
@@ -0,0 +1,86 @@
...
+
+  &:disabled {
+    color: var(--color-context-text-disabled);
+    cursor: default;
+  }
+}
+
+.ax-tabset--medium .ax-tabset__button {
+  padding:
+    calc(var(--component-padding-vertical-medium) - var(--component-border-width))
```
stefanpearson: this is a good example of where I feel it should be split out

src/components/typography/Heading.css
```diff
@@ -0,0 +1,60 @@
...
+
+.ax-heading--headline {
+  font-size: var(--font-size-headline);
+  font-weight: var(--font-weight-headline-standard);
+  letter-spacing: var(--letter-spacing-headline);
+  line-height: var(--line-height-headline);
+}
+
+.ax-heading--title {
+  font-size: var(--font-size-title);
```
stefanpearson: as a heads up, you may want to look at `@apply` to easily mixin some of these repetitions

```css
--font-headline: {
    font-size: var(--font-size-headline);
    line-height: var(--line-height-headline);
    letter-spacing: var(--letter-spacing-headline);
};

.my-el {
    @apply --font-headline;
}
```

style-guide/components/Layout/vars.css
```diff
@@ -0,0 +1,13 @@
...
+
+:root {
+  --header-height-sm: calc(var(--spacing-grid-size) * 12);
+  --header-height-lg: calc(var(--spacing-grid-size) * 25);
+
+  --sidebar-width: 15rem;
+  --sidebar-active-color: var(--color-ui-primary-1);
+  --sidebar-border-width: 0.0625rem;
+
+  --z-index-layout-header: 400;
```
stefanpearson: the best z-index

style-guide/components/Navigation/NavListItem.css
```diff
@@ -0,0 +1,28 @@
...
+.dm-nav__list-item--active > .dm-nav__link {
+  color: var(--sidebar-active-color);
+}
+
+.dm-nav__list-item--open {
+  margin-bottom: calc(var(--spacing-grid-size) * 2);
+}
+
+.dm-nav > .dm-nav__list > .dm-nav__list-item {
+  & > .dm-nav__list > .dm-nav__list-item > .dm-nav__link {
```
stefanpearson: is this nesting needed? the element should be explicit enough, shouldn't it?

src/components/button/Button.css
```diff
@@ -0,0 +1,182 @@
...
+  font-weight: var(--font-weight-bold);
+  letter-spacing: 0.001875rem;
+  line-height: var(--component-line-height);
+  text-align: center;
+  text-transform: uppercase;
+  white-space: nowrap;
+  vertical-align: middle;
+  transition-property: background-color, border-color, color;
+  transition-duration: var(--transition-time-base);
+  transition-timing-function: var(--transition-func);
```
HHogg: Every little helps ...   haha good point, super unnecessary, will change it.

src/components/base/Base.css
```diff
@@ -0,0 +1,179 @@
...
+
+html {
+  font-size: 100%;
+}
+
+body {
+  font-family: var(--font-family-body);
+  font-size: var(--font-size-body);
+  -webkit-font-smoothing: antialiased;
+  -moz-osx-font-smoothing: grayscale;
```
HHogg: That is pro.

src/components/card/CardList.css
```diff
@@ -0,0 +1,61 @@
+@import '../../materials/animations';
+@import '../../materials/colors';
+@import '../../materials/sizing';
+
+:root {
+  --card-medium-padding: calc(var(--spacing-grid-size) * 4);
```
HHogg: Yeah I like that idea 👍

src/components/dialog/Dialog.css
HHogg: Yeah there's definitely some inconsistencies around. I simply mapped the names from Sass, but now would definitely be the perfect time to align them all. 👍

style-guide/components/Navigation/NavListItem.css
HHogg: It's to target the second level, but you're right it doesn't need to be _this_ explicit. 👍

style-guide/components/Layout/vars.css
HHogg: Not as good as `9999`

PR #134, opened 2017-03-31
feat(Colors): Adds interaction brand color variations

HHogg:


src/materials/colors.css
```diff
@@ -20,18 +20,53 @@
...
-  --color-brand-teal: #31A6AF;
-  --color-brand-green: #77B232;
-  --color-brand-chartreuse: #C1B816;
-  --color-brand-amber: #F9B800;
-  --color-brand-orange: #F38700;
-  --color-brand-brown: #AA6130;
-  --color-brand-grey: #78909C;
+  --color-brand-rose-0: #E884A6;
+  --color-brand-rose-1: #F498B7;
+  --color-brand-rose-2: #FFABC7;
```
thatguynamedandy: Having discussed I think I would rather go for...

```css
--color-brand-rose--light: #E884A6;
--color-brand-rose: #F498B7;
--color-brand-rose--dark: #FFABC7;
```

which I know doesn't work too well for things which already have `light` and `dark` in them, but at least having it as a modifier makes it slightly more palatable. If eventually we end up with additional shades I wonder if we could then get away with..

```css
--color-brand-rose--light-2 #E782A4;
--color-brand-rose--light-1: #E884A6;
--color-brand-rose--light: var(--color-brand-rose--light-1);
--color-brand-rose: #F498B7;
--color-brand-rose--dark: #FFABC7;
```

PR #141, opened 2017-04-04
fix(DataPoint): Rendering in Firefox

HHogg:


src/charts/data-point/DataPoints.js
```diff
@@ -42,14 +42,16 @@ export default class DataPoints extends Component {
   };
 
   render() {
-    const { children, type, ...rest } = this.props;
+    const { children, height, type, ...rest } = this.props;
     const arrayChildren = Children.toArray(children);
     const vbWidth = getWidth(type, arrayChildren.length);
+    const style = { height };
```
pmsorhaindo: smooth! 😎

thatguynamedandy: what does Firefox do differently? Just for future reference.

HHogg: It was ignoring the height attribute and scaling the height up to 100% width.

PR #142, opened 2017-04-04
feat(WordCloud): Adds basic animation

thatguynamedandy:
Tested on:
IE11
Edge
Firefox
Chrome
Safari

### DIFF
--- a/src/charts/wordcloud/Word.js
+++ b/src/charts/wordcloud/Word.js
@@ -4,6 +4,7 @@ import './Word.css';
 
 export default class Word extends Component {
   static propTypes = {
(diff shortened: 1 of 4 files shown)


src/charts/wordcloud/Word.js
```diff
@@ -32,18 +33,31 @@ export default class Word extends Component {
   };
 
   static defaultProps = {
+    animationDelay: 0,
```
benji6: just want to check this doesn't log a warning because it's a number instead of a string which is the type defined in the propTypes

thatguynamedandy: good spot, it should be a string 👍

PR #143, opened 2017-04-05
fix(Tooltip): Not showing when cursor has already entered the target …

HHogg:
…element. For example due to another tooltip blocking the cursor before being removed.


src/components/tooltip/TooltipTarget.js
```diff
@@ -13,10 +13,10 @@ export default class TooltipTarget extends Component {
   handleMouseEnter(...args) {
```
benji6: should prob be renamed to `handleMouseMove` I guess

HHogg: Good catch 👍

PR #144, opened 2017-04-05
feat(Typography): Adds Superscript and Subscript components

HHogg:


HHogg:
Yeah, I was a bit unsure on this as well 😕 the designs are specific about where they should be positioned... 

screen shot 2017-04-05 at 16 40 14

screen shot 2017-04-05 at 16 42 25

... not being above the ascender line. I think I'll remove the `<Subscript>` component as what it's essentially doing at the moment can be achieved with a `<Small>` component.

HHogg:
I'm going to close this for the moment, as it doesn't work for 100% of the use cases and as @stefanpearson pointed out doesn't really follow what a sub/sup should _actually_ do. 

_Thank you @stefanpearson for pointing this out, it needs a better solution_

PR #146, opened 2017-04-06
feat(Reveal): Adds Reveal component

HHogg:
_The animations are actually smoother than in the GIF_
reveal

### DIFF
--- a/src/components/reveal/Reveal.js
+++ b/src/components/reveal/Reveal.js
@@ -1,0 +1,94 @@
+import React, { Component, PropTypes } from 'react';
+import { findDOMNode } from 'react-dom';
+import classnames from 'classnames';
+import { Base } from 'bw-axiom';
+import './Reveal.css';
+
+export default class Reveal extends Component {
+  static propTypes = {
+    children: PropTypes.node.isRequired,
+    space: PropTypes.string,
+    visible: PropTypes.bool.isRequired,
+  };
+
+  constructor(props) {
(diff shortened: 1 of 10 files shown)


src/components/reveal/Reveal.js
```diff
@@ -0,0 +1,94 @@
...
+export default class Reveal extends Component {
+  static propTypes = {
+    children: PropTypes.node.isRequired,
+    space: PropTypes.string,
+    visible: PropTypes.bool.isRequired,
+  };
+
+  constructor(props) {
+    super(props);
+    this.state ={
```
thatguynamedandy: `={` accidental emoji.

src/components/reveal/example/reveal.js
```diff
@@ -0,0 +1,92 @@
...
+    super(props);
+    this.state = {
+      visible: false,
+    };
+  }
+
+  handleOnClick() {
+    const { visible } = this.state;
+
+    this.setState({ visible: !visible });
```
thatguynamedandy: I'm starting to favour the functional approach for this kind of change, what do you think?
```
this.setState(({ visible }) => ({ visible: !visible }));
```

HHogg: Yeah I prefer that too :+1: Ok to follow up with a whole library change separately?

thatguynamedandy: of course 👌🏻

PR #147, opened 2017-04-06
fix(Context): Rendering when filter prop is not supported

thatguynamedandy:
screen shot 2017-04-06 at 12 44 58

 The above example is a screenshot from IE11 which does not support `filter: drop-shadow`

### DIFF
--- a/src/components/context/Context.css
+++ b/src/components/context/Context.css
@@ -7,25 +7,40 @@
 
 .ax-context {
   position: relative;
-  filter: drop-shadow(var(--drop-shadow));
+
+  @supports (filter: drop-shadow(var(--drop-shadow))) {
+    filter: drop-shadow(var(--drop-shadow));
+  }
 }
 
 .ax-context__content {
+  border-width: var(--component-border-width);
+  border-style: solid;
   border-radius: var(--component-border-radius);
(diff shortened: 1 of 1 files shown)


src/components/context/Context.css
```diff
@@ -7,7 +7,13 @@
 
 .ax-context {
   position: relative;
-  filter: drop-shadow(var(--drop-shadow));
+  border-width: var(--component-border-width);
+  border-style: solid;
+
+  @supports (filter: drop-shadow(var(--drop-shadow))) {
```
HHogg: The MDN docs says you can proceed the expression with `not`... so could keep the fallback stuff within the `@supports` 

```css
@suppers not (filter: drop-shadow(var(--drop-shadow))) {
  border-width: var(--component-border-width);
  border-style: solid;
}
```

The plus side is it wouldn't require any overrides... the negative side I guess it goes against "progressive enhancement" instead of "graceful degradation" 🤔

thatguynamedandy: yes I must admit I almost went down that route, but I thought progressive enhancement is probably a slightly better convention. Especially if we ever decide to target more bleeding edge features of other browsers.

PR #150, opened 2017-04-10
feat(Pagination): Adds Pagination composite

HHogg:
pagination

### DIFF
--- a/src/components/button/Button.css
+++ b/src/components/button/Button.css
@@ -188,3 +188,11 @@
   border-width: var(--component-border-width-huge);
 }
 
+.ax-button--stadium-small {
+  min-width: var(--component-round-size-small);
+  height: var(--component-round-size-small);
+  padding: 0 calc(var(--component-round-size-small) / 4);
+  border-width: var(--component-border-width-small);
+  border-radius: var(--component-round-size-small);
+  line-height: 0;
+}
--- a/src/components/button/Button.css
+++ b/src/components/button/Button.css
@@ -115,6 +115,14 @@
   &:disabled { color: var(--color-context-text-disabled); }
 }
 
+.ax-button--quaternary {
+  color: var(--color-context-text-subtle);
+
+  &:hover    { color: var(--color-ui-primary--light); }
+  &:active   { color: var(--color-ui-primary--dark); }
+  &:disabled { color: var(--color-context-text-disabled); }
+}
+
 .ax-button--small {
   padding: var(--component-padding-vertical-small) var(--component-padding-horizontal-small);
   font-size: var(--font-size-small);
--- a/src/components/button/Button.js
+++ b/src/components/button/Button.js
@@ -17,7 +17,7 @@ export default class Button extends Component {
       PropTypes.oneOf(['small', 'medium', 'large']),
     ]),
     size: PropTypes.oneOf(['small', 'medium', 'large']),
-    style: PropTypes.oneOf(['primary', 'secondary', 'tertiary']),
+    style: PropTypes.oneOf(['primary', 'secondary', 'tertiary', 'quaternary']),
   };
 
   static contextTypes = {
--- a/src/components/button/Button.js
+++ b/src/components/button/Button.js
@@ -17,6 +17,7 @@ export default class Button extends Component {
       PropTypes.oneOf(['small', 'medium', 'large']),
     ]),
     size: PropTypes.oneOf(['small', 'medium', 'large']),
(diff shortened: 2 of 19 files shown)


src/composites/pagination/Pagination.test.js
```diff
@@ -0,0 +1,99 @@
...
+        currentPage={ 1 }
+        onPageChange={ () => {} }
+        totalPages={ 10 }
+        { ...props } />
+  );
+}
+
+describe('Composite: Pagination', () => {
+  it('getSetStart', () => {
+    assert.equal(getSetStart(1, 5), 1);
```
siggysamson: I don't think that's our usual pattern of writing tests, isn't it? I might have missed something ...

HHogg: The UX is a bit different than normal paginations so required some functions for calculating the numbering. Did it TDD style with these tests, so thought I would leave them in just incase.

siggysamson: Sorry I wasn't clear with my question! 🙂  The tests are fine, I was talking about `assert.equal(getSetStart(1, 5), 1)` - shouldn't this be `expect(getSetStart(1, 5)).toEqual(1)`? It's not a big deal but just thinking that we should keep it consistent?

HHogg: Ahhhh I see, that is me getting confused between projects 😄 We use nodes assert in Audiences, completely forgot about Jest in Axiom. Will update 👍

siggysamson: No worries, that's what I was assuming 😄  I've found one test file that uses `assert.equal` [as well](https://github.com/BrandwatchLtd/axiom/search?utf8=%E2%9C%93&q=assert.equal&type=). We should probably refactor it at some point - not that important though

src/composites/pagination/PaginationButton.js
```diff
@@ -0,0 +1,21 @@
+import React, { Component, PropTypes } from 'react';
+import { Button } from 'bw-axiom';
+
+export default class PaginationButton extends Component {
+  static propTypes = {
+    active: PropTypes.bool,
+    page: PropTypes.number,
+    onClick: PropTypes.func,
```
thatguynamedandy: Are either 'page' or 'onClick' required?

HHogg: To keep the styling the same across all the buttons this component is also used for the `...` ellipsis separators which don't require page or onClick. Also it's purely an internal component.

src/components/button/Button.js
```diff
@@ -17,7 +17,7 @@ export default class Button extends Component {
       PropTypes.oneOf(['small', 'medium', 'large']),
     ]),
     size: PropTypes.oneOf(['small', 'medium', 'large']),
-    style: PropTypes.oneOf(['primary', 'secondary', 'tertiary']),
+    style: PropTypes.oneOf(['primary', 'secondary', 'tertiary', 'quaternary']),
```
thatguynamedandy: I'm not a big fan of `quaternary`, but equally I can't off much in a way of an alternative.

src/components/button/Button.js
```diff
@@ -17,7 +17,8 @@ export default class Button extends Component {
       PropTypes.oneOf(['small', 'medium', 'large']),
     ]),
     size: PropTypes.oneOf(['small', 'medium', 'large']),
-    style: PropTypes.oneOf(['primary', 'secondary', 'tertiary']),
```
thatguynamedandy: I get it now, because it's the shape of a stadium 😀 . I think we should probably question any additional top level variants creeping in.

src/components/button/Button.css
```diff
@@ -115,6 +115,14 @@
   &:disabled { color: var(--color-context-text-disabled); }
 }
 
+.ax-button--quaternary {
```
thatguynamedandy: I'm not a massive fan of `quaternary`, but unfortunately I can't offer anything more sensible.

src/components/button/Button.js
HHogg: 😄  it's actually a [legit geometrical shape](https://en.wikipedia.org/wiki/Stadium_(geometry))... but you're right, we're slowly approaching a large amount of variations. I'm hoping as the design system is maturing this will start levelling out.

src/components/button/Button.css
HHogg: Exactly how me and Christian were felling when adding it 😕

PR #165, opened 2017-04-24
feat(BarChart): Adds new Bar Chart type

lpoulter:
Adds a new bar chart and supporting components. 

screen shot 2017-05-02 at 14 33 58

### DIFF
--- a/src/charts/bar-chart/Bar.js
+++ b/src/charts/bar-chart/Bar.js
@@ -1,0 +1,52 @@
+import React, { Component } from 'react';
+import PropTypes from 'prop-types';
+import { Small } from 'bw-axiom';
+import classnames from 'classnames';
+import './Bar.css';
+
+export default class Bar extends Component {
+  static propTypes = {
+    color: PropTypes.oneOf([
+      'rose',
+      'pink',
+      'purple',
+      'lilac',
+      'blue',
+      'teal',
+      'green',
+      'chartreuse',
+      'amber',
+      'orange',
+      'brown',
+      'grey',
+    ]).isRequired,
+    percentage: PropTypes.number.isRequired,
+  };
+
+  render() {
+    const {
+      color,
+      percentage,
+      ...rest
+    } = this.props;
+
+    const classes = classnames('ax-bar-chart__bar-rect', {
+      [`ax-bar-chart__bar-rect--${color}`]: color,
+    });
+
+    const style = {
+      flexBasis: `${percentage}%`,
+    };
+
+    return (
+      <div { ...rest } className="ax-bar-chart__bar-container">
+        <div className="ax-bar-chart__bar">
+          <div className="ax-bar-chart__bar-label">
+            <Small textColor="subtle">{ percentage }%</Small>
+          </div>
+          <div className={ classes } style={ style }></div>
+        </div>
+      </div>
+    );
+  }
+}
--- a/src/charts/bar-chart/BarChart.js
+++ b/src/charts/bar-chart/BarChart.js
@@ -1,0 +1,25 @@
+import React, { Component } from 'react';
+import PropTypes from 'prop-types';
+import { Base } from 'bw-axiom';
+import './BarChart.css';
+
+export default class BarChart extends Component {
+  static propTypes = {
+    children: PropTypes.node.isRequired,
+    height: PropTypes.string.isRequired,
+  };
+
+  render() {
+    const {
+      children,
+      height,
+      ...rest
+    } = this.props;
+
+    return (
+      <Base { ...rest } className="ax-bar-chart" style={ { height: height } }>
+        { children }
+      </Base>
+    );
+  }
+}
--- a/src/charts/bar-chart/BarGroup.js
+++ b/src/charts/bar-chart/BarGroup.js
@@ -1,0 +1,35 @@
+import React, { Component } from 'react';
+import PropTypes from 'prop-types';
+import { Small, Weak, Heading, Strong, Paragraph } from 'bw-axiom';
+import './BarGroup.css';
+
+export default class BarGroup extends Component {
+  static propTypes = {
+    children: PropTypes.node.isRequired,
+    label: PropTypes.string.isRequired,
+    ratio: PropTypes.string,
+  };
+
+  render() {
+    const {
+      children,
+      label,
+      ratio,
+      ...rest
+    } = this.props;
+
+    return (
+      <div { ...rest } className="ax-bar-chart__group">
+        <div className="ax-bar-chart__group-bars">
+          { children }
+        </div>
+        <div className="ax-bar-chart__group-label">
+          { ratio && <Paragraph>
+            <Small><Weak>{ ratio }%</Weak></Small>
+          </Paragraph> }
+          <Heading><Strong>{label}</Strong></Heading>
+        </div>
+      </div>
+    );
+  }
+}
--- a/src/charts/bar-chart/Bar.css
+++ b/src/charts/bar-chart/Bar.css
@@ -1,0 +1,44 @@
+@import '../../materials/colors';
+@import '../../materials/sizing';
+
+:root {
+  --ax-bar-bar-label-spacing: calc(var(--spacing-grid-size) * 2);
+}
+
+.ax-bar-chart__bar-container {
+  position: relative;
+  flex: 1 1 0%;
+}
+
+.ax-bar-chart__bar {
+  display: flex;
+  position: absolute;
+  top: 0;
+  left: 0;
+  flex-direction: column;
+  justify-content: flex-end;
+  width: 100%;
+  height: 100%;
+}
+
+.ax-bar-chart__bar-label {
+  flex: 0 0 auto;
+  margin-bottom: var(--ax-bar-bar-label-spacing);
+}
+
+.ax-bar-chart__bar-rect {
+  flex: 0 1 0%;
+}
+
+.ax-bar-chart__bar-rect--rose { background-color: var(--color-brand-rose); }
+.ax-bar-chart__bar-rect--pink { background-color: var(--color-brand-pink); }
+.ax-bar-chart__bar-rect--purple { background-color: var(--color-brand-purple); }
+.ax-bar-chart__bar-rect--lilac { background-color: var(--color-brand-lilac); }
+.ax-bar-chart__bar-rect--blue { background-color: var(--color-brand-blue); }
+.ax-bar-chart__bar-rect--teal { background-color: var(--color-brand-teal); }
+.ax-bar-chart__bar-rect--green { background-color: var(--color-brand-green); }
+.ax-bar-chart__bar-rect--chartreuse { background-color: var(--color-brand-chartreuse); }
+.ax-bar-chart__bar-rect--amber { background-color: var(--color-brand-amber); }
+.ax-bar-chart__bar-rect--orange { background-color: var(--color-brand-orange); }
+.ax-bar-chart__bar-rect--brown { background-color: var(--color-brand-brown); }
+.ax-bar-chart__bar-rect--grey { background-color: var(--color-brand-grey); }
--- a/src/charts/bar-chart/Bar.test.js
+++ b/src/charts/bar-chart/Bar.test.js
@@ -1,0 +1,53 @@
+import React from 'react';
+import renderer from 'react-test-renderer';
+import { Bar } from 'bw-axiom';
+
+function getComponent(props = { percentage: 43 }) {
+  return renderer.create(
+    <Bar { ...props } />
+  );
+}
+
+describe('Bar', () => {
+  it('renders with defaultProps', () => {
+    const component = getComponent();
+    const tree = component.toJSON();
+    expect(tree).toMatchSnapshot();
+  });
+
+  describe('renders with color', () => {
+    [
+      'rose',
+      'pink',
+      'purple',
+      'lilac',
+      'blue',
+      'teal',
+      'green',
+      'chartreuse',
+      'amber',
+      'orange',
+      'brown',
+      'grey',
+    ].forEach((color) => {
+      it(color, () => {
+        const component = getComponent({ color });
+        const tree = component.toJSON();
+        expect(tree).toMatchSnapshot();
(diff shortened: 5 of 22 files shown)


src/charts/bar-chart/bar.css
```diff
@@ -0,0 +1,37 @@
+@import '../../materials/colors';
+
+.bar-chart {
+    display:flex;
+    padding: 15px;
+    height: 200px;
```
HHogg: This would be better as a configurable property on the BarChart component.

src/charts/bar-chart/bar.css
```diff
@@ -0,0 +1,37 @@
...
+}
+
+.column {
+    display:flex;
+    flex: 1;
+    margin-right: 2%;
```
HHogg: Again with the spacing variables.

src/charts/bar-chart/bar.css
```diff
@@ -0,0 +1,37 @@
+@import '../../materials/colors';
+
+.bar-chart {
```
HHogg: Also should this go into a CSS file for the `BarChart.css`? As this is the `Bar.css` file?

src/charts/bar-chart/bar.css
```diff
@@ -0,0 +1,37 @@
...
+
+.bar-container {
+    width: 50%;
+    display: flex;
+    flex-direction: column;
+    justify-content: flex-end;
+}
+
+.label-column {
+    margin-top: 20px;
```
HHogg: Variable.

src/charts/bar-chart/bar.css
```diff
@@ -0,0 +1,37 @@
...
+    flex-direction: column;
+    text-align: center;
+}
+
+.column:last-child {
+    margin-right: 0;
+}
+
+.bar-container {
+    width: 50%;
```
HHogg: What's the purpose of this width? If all the Bars are flexed should they need a width?

src/charts/bar-chart/BarChart.js
```diff
@@ -0,0 +1,47 @@
...
+
+  render() {
+    const {
+      data,
+    } = this.props;
+
+    const columns = [];
+    const max = findMax(data.series);
+
+    data.series.map((series, index) => {
```
HHogg: [Array.map](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/map?v=example) returns an array created from the values that are returned from the callback... so you could just do... 

```js
const columns =  data.series.map((series, index) =>
  <Column key={ index } max={ max } series={ series }></Column>
);
```

src/charts/bar-chart/BarChart.js
```diff
@@ -0,0 +1,47 @@
...
+      columns.push(
+        <Column key={ index } max={ max } series={ series }></Column>
```
HHogg: The problem with this kind of API is it doesn't allow users to make use of passing standard React props to the elements. For example they might want to add an `onClick` or a `data-attribute` to the elements. So it's best to let the user take care of the mapping themselves.

src/charts/bar-chart/BarChart.js
```diff
@@ -0,0 +1,47 @@
...
+      );
+    });
+
+    return (
+      <div>
```
HHogg: There doesn't seem to be a need for this blank div?

src/charts/bar-chart/BarChart.js
```diff
@@ -0,0 +1,47 @@
...
+        <div className="ax-bar-chart">
```
HHogg: Use the `Base` component here as it will provide top level styling ... for example vertical margins.

src/charts/bar-chart/Column.js
```diff
@@ -0,0 +1,43 @@
...
+    const bars =[];
+
+    const colors = ['ax-ui-light--dark','ax-ui-primary--dark' ];
+
+    series.data.map((dataPoint, index) => {
+      const barHeight = this.getRatio(dataPoint);
+
+      bars.push(
+        <Bar
+            color={ colors[index % colors.length] }
```
HHogg: This should be a configurable property by the user.

src/charts/bar-chart/bar.css
```diff
@@ -0,0 +1,36 @@
+@import '../../materials/colors';
+
+.ax-bar-chart {
+  display: flex;
+  height: 25rem;
```
HHogg: This should be a configurable property by the user, as different situations may require different heights.

src/charts/bar-chart/bar.css
```diff
@@ -0,0 +1,36 @@
...
+}
+
+.ax-column {
+  display: flex;
+  flex: 1;
+  margin-right: 2%;
```
HHogg: This seems a little bit random, probably want to use one [the sizing variables in materials](https://github.com/BrandwatchLtd/axiom/blob/master/src/materials/sizing.css#L3) and ask Design how much it should be spaced by. [(Calc example)](https://github.com/BrandwatchLtd/axiom/blob/master/src/components/alert/Alert.css#L6)

src/charts/bar-chart/bar.css
```diff
@@ -0,0 +1,36 @@
...
+
+.ax-column:last-child {
+  margin-right: 0;
+}
+
+.ax-bar-container {
+  display: flex;
+  flex-direction: column;
+  justify-content: flex-end;
+  width: 50%;
```
HHogg: What's the purpose of this width?

src/charts/bar-chart/bar.css
```diff
@@ -0,0 +1,36 @@
...
+}
+
+.ax-label-column {
+  margin-top: 1rem;
```
HHogg: Sizing variable.

src/charts/bar-chart/bar.css
```diff
@@ -0,0 +1,36 @@
...
+}
+
+.ax-ui-light--dark { background-color: var(--color-ui-dark--light); }
+.ax-ui-primary--dark { background-color: var(--color-ui-primary--dark); }
```
HHogg: Following from the other visualisations, they tend to all use the branding colours. Take a look at the [Word component](https://github.com/BrandwatchLtd/axiom/blob/master/src/charts/wordcloud/Word.css#L19)

src/charts/bar-chart/Column.js
```diff
@@ -0,0 +1,43 @@
...
+        </Bar>
+      );
+    });
+
+    return (
+      <div className="ax-column ax-column-label">
+        <div className="ax-column">
+          { bars }
+        </div>
+        <div className="ax-label-column">{series.label}</div>
```
HHogg: This structure looks a little bit confusing, not really sure what's a column and what's a label as every classname has both words in it?

src/charts/bar-chart/Bar.js
```diff
@@ -0,0 +1,34 @@
+import React, { Component, PropTypes } from 'react';
+import classnames from 'classnames';
+
+export default class Bar extends Component {
+  static propTypes = {
+    color: PropTypes.string.isRequired,
```
lpoulter: Any idea what colours i should use? I have designs for the final facebook component but hat uses different colour to axiom.

src/charts/bar-chart/BarChart.js
```diff
@@ -0,0 +1,47 @@
...
+    const {
+      data,
+    } = this.props;
+
+    const columns = [];
+    const max = findMax(data.series);
+
+    data.series.map((series, index) => {
+      columns.push(
+        <Column key={ index } max={ max } series={ series } suffix={ data.labels.units }></Column>
```
HHogg: The problem with this kind of API is it doesn't allow users to make use of passing standard React props to the elements. For example they might want to add an `onClick` or a `data-attribute` to the elements. So it's best to let the user take care of the mapping themselves.

src/charts/bar-chart/bar.css
lpoulter: Without it the 2nd bar child shrinks to fit its label i would have thought flex one would fix this but it does not. 

screen shot 2017-04-25 at 12 06 08

HHogg: That's because you have set display flex on the parent but not any flex properties on the children so they are defaulting to shrinking to their contents width...

```css
flex: 1 0 0%;
```

... is more along the lines of what you need

src/charts/bar-chart/Bar.js
```diff
@@ -0,0 +1,46 @@
...
+      [`ax-bar--${color}`]: color,
+    });
+
+    const barStyle = {
+      height: `${percentage}%`,
+    };
+
+    return (
+      <div { ...rest } className="ax-bar__container">
+        <div>{ percentage }%</div>
```
HHogg: The `<div>` here isn't needed, it's got no styling or anything. Looking at the other charts, the percentage labels all use the subtle coloring. Like ...

```jsx
<Small textColor="subtle"> ... </Small>
```

src/charts/bar-chart/BarChart.js
```diff
@@ -0,0 +1,25 @@
+import React, { Component, PropTypes } from 'react';
+import { Base } from 'bw-axiom';
+import './bar.css';
```
HHogg: The CSS file names should be TitleCamelCase like the Javascript component files. Also the CSS could do with being separated so each components CSS belongs with the component. 

```
BarChart.css
BarGroup.css
Bar.css
```

src/charts/bar-chart/BarChart.js
```diff
@@ -0,0 +1,25 @@
...
+
+export default class BarChart extends Component {
+  static propTypes = {
+    children: PropTypes.node.isRequired,
+    minimumHeight: PropTypes.string.isRequired,
```
HHogg: Minimum height probably doesn't make sense for a chart because the contents is percentage based so would never be larger than the defined height. So probably just `height`.

src/charts/bar-chart/BarChart.js
```diff
@@ -0,0 +1,25 @@
...
+  render() {
+    const {
+      children,
+      minimumHeight,
+      ...rest
+    } = this.props;
+
+    return (
+      <Base { ...rest } className="ax-bar__chart" style={ { minHeight: minimumHeight } }
+          width="100%">
```
HHogg: This width property is invalid. Base would produce a `<div>` here so a width attribute won't do anything.

src/charts/bar-chart/BarGroup.js
```diff
@@ -0,0 +1,25 @@
...
+
+  render() {
+    const {
+      children,
+      label,
+      ...rest
+    } = this.props;
+
+    return (
+      <div { ...rest } className="ax-bar__column">
```
HHogg: The classes of the elements is still a little bit off. Mainly the block should be the top level like and the naming should match up to the components. 

This is what the output HTML will look like at the moment. 

```html
<div class="ax-bar__chart"> <!-- <BarChart> -->
  <div class="ax-bar__column"> <!-- <BarGroup> -->
    <div class="ax-bar__group">
      <div class="ax-bar__container"> <!-- <Bar> -->
        <div></div>
        <div class="ax-bar ax-bar--rose"></div>
      </div> <!-- </Bar> -->
    </div> 

    <div class="ax-bar__column-label"></div>
  </div> <!-- </BarGroup> -->
</div> <!-- </BarChart> -->
```

Where as a template like this, you can see that the all of the elements are part of the `ax-bar-chart` block. 

```html
<div class="ax-bar-chart"> <!-- <BarChart> -->
  <div class="ax-bar-chart__group"> <!-- <BarGroup> -->
    <div class="ax-bar-chart__group-bars">
      <div class="ax-bar-chart__bar"> <!-- <Bar> -->
        10%
        <div class="ax-bar-chart__bar-rect ax-bar-chart__bar-rect--rose"></div>
      </div> <!-- </Bar> -->
    </div> 

    <div class="ax-bar-chart__group-label"></div>
  </div> <!-- </BarGroup> -->
</div> <!-- </BarChart> -->
```

src/charts/bar-chart/bar.css
```diff
@@ -0,0 +1,54 @@
+@import '../../materials/colors';
+@import '../../materials/sizing';
+
+.ax-bar__chart {
+  display: flex;
+  flex-wrap: wrap;
+  padding: var(--component-padding-vertical-medium);
```
HHogg: It feels like you might have randomly selected these values 😄 Because this will apply the vertical padding to the left and right sides too... Double check with design what these multiples of the grid size these should, and you can then scope them to the top of the file and give them a bit of context like...

```css
:root {
  --cmp-bar-chart-group-spacing: calc(var(--spacing-grid-size) * 3));
}

.ax-bar-chart__group {
  margin-right: var(--cmp-bar-chart-group-spacing);

  &:last-child { 
    margin-right: 0 
  }
}
```

HHogg: There's a few other instances below.

src/charts/bar-chart/bar.css
```diff
@@ -0,0 +1,54 @@
...
+
+.ax-bar__container {
+  display: flex;
+  flex: 1;
+  flex-direction: column;
+  justify-content: flex-end;
+}
+
+.ax-bar__container:hover {
+  opacity: 0.7;
```
HHogg: By using an opacity here it will produce a colour that doesn't match any of the colours that design provide. please use only colours that are available in the palette. I guess bars that don't have interactable behaviour should have feedback. So you'll probably have to do something along the lines of.... 

```js
const classes = classnames('ax-bar-chart__bar', `ax-bar-chart__bar--${color}`, {
  'ax-bar-chart__bar--clickable': onClick,
});
```
```css
.ax-bar-chart__bar--clickable {
  &.ax-bar-chart__bar-rect--rose { 
    &:hover { background-color: var(--color-brand-rose--light) }
    &:active { background-color: var(--color-brand-rose--dark) }
  }

  &.ax-bar-chart__bar-rect--pink { 
    &:hover { background-color: var(--color-brand-pink--light) }
    &:active { background-color: var(--color-brand-pink--dark) }
  }

  ...etc
}
```

Issue #161, opened 2017-04-19
Landing page for style guide


ahume:
As a developer that wants to shout about Brandwatch being awesome, I want to Tweet a link to Axiom, but it doesn't look like a thing yet.

I'd like to link to https://brandwatchltd.github.io/axiom/ but what appears is a strange looking page with "Chart Heading" as a title and some funny looking SGML. Which makes me hesitant about sharing the top quality work we're doing in the area of pattern libraries and style guides.

On the landing page could there be a heading that says "Axiom", and a small paragraph that explains what Axiom is, and how it is used.? Does the pattern library include such components?

HHogg:
I'm on it 😉

PR #169, opened 2017-04-26
docs(StyleGuide): Custom structure

HHogg:
The original way of reading from the folder structure put the docs for the styleguide in a not great order. We're also going to get a newly designed style guide from design at some point which will contain pages that don't exist in the docs, so this gives us a little more flexibility in ordering the docs.  

It's currently up on [github.io/axiom](https://brandwatchltd.github.io/axiom) if you want to check out the order.


thatguynamedandy:
Just for reference this links to https://github.com/BrandwatchLtd/axiom/issues/161

PR #170, opened 2017-04-26
feat(StatCard): Adds new StatCard component

HHogg:
screen shot 2017-04-27 at 00 01 37


src/components/statcard/example/utils.js
```diff
@@ -0,0 +1,5 @@
+import { tinyNumber } from 'bw-axiom';
+
+export function n(l = 30000, u = 999999) {
```
benji6: maybe a more descriptive name like `randomStatNumber`?

PR #171, opened 2017-04-28
DotPlot Chart & Some other charting components 

HHogg:
screen shot 2017-04-28 at 12 09 27

**Browser Checks**:
- [X] IE11
- [X] Edge
- [X] Chrome
- [X] Firefox
- [X] Safari


src/charts/chart-grid/ChartGrid.css
```diff
@@ -0,0 +1,26 @@
+@import '../../materials/colors';
+@import '../../materials/layout';
```
benji6: presumably this import isn't required?

src/charts/chart-grid/ChartGrid.js
```diff
@@ -0,0 +1,36 @@
...
+import './ChartGrid.css';
+
+export default class ChartGrid extends Component {
+  static propTypes = {
+    count: PropTypes.number.isRequired,
+  };
+
+  render() {
+    const { count, ...rest } = this.props;
+    const increments = Array.from({ length: count }, (_, i) => i * 10);
```
benji6: Looks familiar :D

src/charts/dot-plot/Dot.js
```diff
@@ -0,0 +1,18 @@
...
+import { DataPoint } from 'bw-axiom';
+import omit from 'lodash.omit';
+
+export default class Dot extends Component {
+  static propTypes = {
+    percent: PropTypes.number.isRequired,
+  };
+
+  render() {
+    const { ...rest } = this.props;
```
benji6: surely no point in shallow cloning `props` here?

src/charts/chart-grid/ChartGrid.css
HHogg: This was my mistake, it was needed for making the grid lines responsive... just forgot to add that feature back in 😄. Done.

src/charts/chart-grid/ChartGrid.js
jdelight: nice little pattern 👍

benji6: OK we can't actually use `Array.from` if we want to support IE11. You can do this instead to keep it ES5 compliant:

```js
Array.apply(null, { length: count }).map((_, i) => i * 10)
```

PR #173, opened 2017-04-28
Remove experimental js

benji6:
takes us from `babel-preset-stage-0` to `babel-preset-stage2`! 🎉

### DIFF
--- a/.babelrc
+++ b/.babelrc
@@ -3,7 +3,7 @@
     "development": {
       "presets": [
         "es2015",
-        "stage-0",
+        "stage-2",
         "react",
         "react-hmre",
       ],
@@ -11,7 +11,7 @@
     "production": {
       "presets": [
         "es2015",
-        "stage-0",
+        "stage-2",
         "react",
       ],
     }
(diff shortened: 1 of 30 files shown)


siggysamson:
👏  I really like seeing that. Any reason you didn't use [babel-preset-env](https://github.com/babel/babel-preset-env)? I don't remember exactly what it includes or doesn't but maybe that gives a bit more flexibility as we move forward? New features would be coming in as we update babel-preset-env and also the compiled output would change based on our supported browser list

HHogg:
@siggysamson makes a good point, we also have the browsers set in the [postcss.config.js](https://github.com/BrandwatchLtd/axiom/blob/master/postcss.config.js#L5) and AFAIK it's the same format. Could move these to a central `browsers.js`?

benji6:
@siggysamson love that idea but we're still dependent on some experimental transforms. I guess I could use `babel-preset-env` with the experimental transforms the code relies on (e.g. `babel-plugin-transform-class-properties`) specified separately, but currently it amounts to the same thing. Maybe something to look at down the line... Personally I'd be quite up for sticking with spec-compliant js all the way, but this'll do for now!

PR #176, opened 2017-05-03
fix(CSS): Removed normalize.css

HHogg:
Removed normalize.css due to issues with inserting Axiom into an existing application. Applied needed rules, some scoped to Axiom classes

### DIFF
--- a/src/components/base/Base.css
+++ b/src/components/base/Base.css
@@ -1,4 +1,3 @@
-@import 'normalize.css/normalize.css';
 @import '../../materials/colors';
 @import '../../materials/layout';
 @import '../../materials/sizing';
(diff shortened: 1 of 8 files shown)


siggysamson:
Good move 👍  This makes it more solid to use in any application as we don't change the global CSS (well, kinda - at least we are moving into that direction. There are still some `button`, `html` and `body` declarations. We should think about how we are handling those)

src/components/base/Base.css
```diff
@@ -52,9 +51,14 @@ body {
...
+  -ms-text-size-adjust: 100%;
+  -webkit-text-size-adjust: 100%;
+  /* stylelint-enable property-no-vendor-prefix */
 }
 
 body {
+  margin: 0;
   font-family: var(--font-family-body);
   font-size: var(--font-size-body);
   -webkit-font-smoothing: antialiased;
```
larister: ack sorry @HHogg but this is still causing a discrepancy - lemme check with UX though as it seems like a "good" change to me

larister: I checked with @edhesketh and he said it should be fine 👍

PR #177, opened 2017-05-03
Make labels in bar chart optional

lpoulter:
Makes the the X-axis labels optional.

### DIFF
--- a/src/charts/bar-chart/BarGroup.js
+++ b/src/charts/bar-chart/BarGroup.js
@@ -6,7 +6,7 @@ import './BarGroup.css';
 export default class BarGroup extends Component {
   static propTypes = {
     children: PropTypes.node.isRequired,
-    label: PropTypes.string.isRequired,
+    label: PropTypes.string,
     ratio: PropTypes.string,
   };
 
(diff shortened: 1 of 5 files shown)


src/charts/bar-chart/BarGroup.js
```diff
@@ -23,12 +23,15 @@ export default class BarGroup extends Component {
         <div className="ax-bar-chart__group-bars">
           { children }
         </div>
+        { (ratio || label) &&
         <div className="ax-bar-chart__group-label">
           { ratio && <Paragraph>
             <Small><Weak>{ ratio }%</Weak></Small>
           </Paragraph> }
-          <Heading><Strong>{label}</Strong></Heading>
+          { label && <Heading>
```
HHogg: I really think the correct solution to this is to have the xAxis labels as a separate component as we explored. Even with this change trying to get all the components to align to the designs won't be possible. 

This is the intended alignment from the designs....
screen shot 2017-05-03 at 14 37 17

However by hiding the labels in one chart the alignment will become...

screen shot 2017-05-03 at 14 37 34

lpoulter:
Closing as no longer needed.

PR #178, opened 2017-05-03
feat(DataPicker): Adds DataPicker chart component

pmsorhaindo:
screen shot 2017-05-04 at 2 18 48 pm


src/charts/data-picker/DataPickerHeader.js
```diff
@@ -0,0 +1,75 @@
...
+    return (
+      <PanelHeader { ...rest }>
+        <Grid gutters="tiny" verticalAlign="middle">
+          <GridCell shrink={ true }>
+            <DataPoints height="1rem">
+              <DataPoint color={ color } radius="1" />
+            </DataPoints>
+          </GridCell>
+          <GridCell>
+            <Dropdown>
```
HHogg: I think this section may need to be conditional, looking at some of the designs it looks like some of the situations where it's and enforce data series and can't be altered. Basically looks like the header text is output with no chevron on dropdown functionality.

src/charts/data-picker/DataPickerHeader.js
```diff
@@ -0,0 +1,75 @@
...
+              <DropdownTarget>
+                <Link style="subtle">
+                  <Grid gutters="tiny" verticalAlign="midddle">
+                    <GridCell shrink={ true }>
+                      <Heading style="title">{ headerText || defaultText }</Heading>
```
HHogg: I don't think you need to handle the defaultText within this component. The style you've already handled so it could simply be used from an API perspective as 

```jsx
<DataPickerHeader text={ mySelectedText || myDefaultText }>
  ...etc
</DataPickerHeader>
```

src/charts/data-picker/DataPickerHeader.js
```diff
@@ -0,0 +1,75 @@
...
+                    </GridCell>
+                    <GridCell>
+                      <Icon name="chevron-down"/>
+                    </GridCell>
+                  </Grid>
+                </Link>
+              </DropdownTarget>
+
+              <DropdownContent>
+                { findComponent(children, DataPickerMenu) }
```
HHogg: I don't think you need to do a findComponent here... the only children expected will be a single child of DataPickerMenu... following on from the previous comment about conditionally adding the Dropdown behaviour. You could easily conditionally add it by a children check 

```jsx
{ children && (
  <Dropdown>
    ...
  </Dropdown>
) }
```

src/charts/data-picker/example/data-picker.js
```diff
@@ -0,0 +1,127 @@
...
+
+  onThirdMenuItemClick(id) {
+    this.setState({ thirdSelectedId: id });
+  }
+
+  render() {
+    const { firstSelectedId, secondSelectedId, thirdSelectedId } = this.state;
+    const firstSelectedValue = getData().find((query) => query.id === firstSelectedId);
+    const secondSelectedValue = getData().find((query) => query.id === secondSelectedId);
+    const thirdSelectedValue = getData().find((query) => query.id === thirdSelectedId);
```
HHogg: Unfortunately, we can't use `find` anymore :) see the babel-polyfill PR removal.

style-guide/routes.js
```diff
@@ -212,8 +212,8 @@ export default (
         <Route
             path="chart-panel"
             name="Chart Panel"
```
HHogg: These values should probably be updated too? I wonder if it makes more sense for the Panel component to now be moved to Components rather than Charts.

style-guide/routes.js
```diff
@@ -222,6 +222,12 @@ export default (
             examples={ require('../src/charts/data-point/example') }
             component={ Doc } />
         <Route
+            path="data-picker"
```
HHogg: This should probably be programatically done, but the order defined here makes it's way into the menu... which is alphabetically ordered :wink:

src/charts/data-picker/example/data.js
```diff
@@ -0,0 +1,23 @@
+export function getData() {
```
HHogg: This doesn't look like it's dynamic so does it need to be a function?

src/charts/data-picker/example/data-picker.js
```diff
@@ -0,0 +1,127 @@
...
+    return (
+      <Example name="Data  Picker">
+        <Snippet>
+          <Grid gutters="tiny">
+            <GridCell>
+              <DataPicker>
```
HHogg: Is it possible to loop over these three to prevent duplicating the structure in the example? Also may be worthing adding some of those badly names `snippetSkip|Ignore` properties to tidy up the snippet.

src/charts/data-picker/example/data-picker.js
pmsorhaindo: 😭

src/charts/data-picker/example/data.js
pmsorhaindo: I guess it doesn't need to be, but it prevents accessing a modified array as `const` doesn't protect an array from array operations.

src/charts/data-picker/example/data-picker.js
```diff
@@ -0,0 +1,105 @@
...
+      value: secondSelectedValue,
+    }, {
+      color: 'pink',
+      id: thirdSelectedId,
+      onClick: this.onThirdMenuItemClick,
+      value: thirdSelectedValue,
+    }];
+
+    return (
+      <Example name="Data  Picker">
```
benji6: 2 spaces here - not sure if that has any consequences?

src/charts/data-picker/example/data-picker.js
```diff
@@ -0,0 +1,105 @@
...
+        <Snippet>
+          <Grid gutters="tiny" snippetIgnore="ignore">
+            { data.map(({ color, id, onClick, value }, i) => (
+              <GridCell key={ i } snippetIgnore="ignore">
```
benji6: key should probably be `id` right?

src/charts/data-picker/DataPickerMenu.js
```diff
@@ -0,0 +1,22 @@
+import React, { Component, PropTypes } from 'react';
```
benji6: import PropTypes from prop-types package

src/charts/data-picker/DataPickerMenuItem.js
```diff
@@ -0,0 +1,18 @@
+import React, { Component, PropTypes } from 'react';
```
benji6: PropTypes import

src/charts/data-picker/example/data-picker.js
pmsorhaindo: I think either works, the id should be unique but that isn't guaranteed. `i` is guaranteed unique.

benji6: There's probably a bug elsewhere if the `id` is not unique. Generally speaking you can get weird bugs if you use the array index instead of an id, but actually in this case either is fine because the data is static so nbd 👍

PR #179, opened 2017-05-04
Make labels optional in bar-chart. Add new axis-labels component with…

lpoulter:
WIP

Splits the Axis labels on the bar chart out to a separate component.

- [x] Split out labels

- [ ]  Write tests for new component.

- [x] Remove labels from bar-chart components and update examples to use new component


lpoulter:
Closing as no longer needed due #186.

PR #180, opened 2017-05-04
feat(chart-table): add expand/collapse functionality

benji6:
screen shot 2017-05-04 at 14 22 37
screen shot 2017-05-04 at 14 22 50

### DIFF
--- a/src/charts/chart-axis/ChartAxisLabels.css
+++ b/src/charts/chart-axis/ChartAxisLabels.css
@@ -2,6 +2,7 @@
 
 .ax-chart-axis-labels {
   display: flex;
+  width: 100%;
 }
 
 .ax-chart-axis-labels__label {
--- a/src/charts/chart-table/ChartTable.css
+++ b/src/charts/chart-table/ChartTable.css
@@ -1,7 +1,29 @@
+@import '../../materials/animations';
 @import '../../materials/sizing';
 
 :root {
   --cmp-chart-table-row-padding: calc((var(--spacing-grid-size) * 3) / 2);
+}
+
+.ax-chart-table__axis-button-container {
+  flex-shrink: 0;
+}
+
+.ax-chart-table__axis-row-container {
+  display: flex;
+}
+
+.ax-chart-table__collapse-expand-text {
+  margin-left: calc(var(--spacing-grid-size) * 2);
+}
+
+.ax-chart-table__cell {
+  padding-top: var(--cmp-chart-table-row-padding);
+  padding-bottom: var(--cmp-chart-table-row-padding);
+}
+
+.ax-chart-table__cell:first-child {
+  padding-left: calc(var(--spacing-grid-size) * 6);
 }
 
 .ax-chart-table,
@@ -11,7 +33,11 @@
   border-spacing: 0;
 }
 
-.ax-chart-table__cell {
-  padding-top: var(--cmp-chart-table-row-padding);
-  padding-bottom: var(--cmp-chart-table-row-padding);
+.ax-chart-table__rows-container {
+  /* prevents DataPoints from being clipped on right edge */
+  padding-right: calc(var(--spacing-grid-size) * 2);
+  overflow-y: hidden;
+  transition-property: height;
+  transition-duration: var(--transition-time-base);
+  transition-timing-function: var(--transition-function);
 }
--- a/src/charts/chart-table/ChartTable.js
+++ b/src/charts/chart-table/ChartTable.js
@@ -7,20 +7,46 @@ import './ChartTable.css';
 export default class ChartTable extends Component {
   static propTypes = {
     children: PropTypes.node.isRequired,
+    collapsedVisibleRowCount: PropTypes.number,
+    expandButtonSuffix: PropTypes.string,
     labelColumnWidth: PropTypes.string,
   };
 
   static childContextTypes = {
+    collapsedVisibleRowCount: PropTypes.number,
+    collapsible: PropTypes.bool.isRequired,
+    expandButtonSuffix: PropTypes.string.isRequired,
+    isExpanded: PropTypes.bool.isRequired,
     labelColumnWidth: PropTypes.string,
+    rowsCount: PropTypes.number.isRequired,
+    setRowsCount: PropTypes.func.isRequired,
+    toggleExpand: PropTypes.func.isRequired,
   };
 
   static defaultProps = {
     labelColumnWidth: '10rem',
   };
 
+  constructor(props) {
+    super(props);
+    this.state = {
+      isExpanded: this.props.collapsedVisibleRowCount == null,
+      rowsCount: 0,
+    };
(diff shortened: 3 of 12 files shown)


thatguynamedandy:
Minor one but the commit message should probably be camel cased `feat(ChartTable)`

src/charts/chart-table/ChartTableRows.test.js
```diff
@@ -7,20 +8,44 @@ import {
   ChartTableVisual,
 } from 'bw-axiom';
 
-function getComponent(props = {}) {
+class ContextProvider extends Component {
```
benji6: The problem I guess is that you need to define the `childContextTypes` so I wasn't really sure how much time was actually saved by defining a helper especially since this isn't used anywhere else right now. I'm also not sure if there's a better way of doing this... Might be worth looking into how enzyme manages to render with context... I think I'm going to leave it for now and we can factor it out if we need to make more use of context in our tests down the line

src/charts/chart-table/ChartTable.css
```diff
@@ -1,17 +1,43 @@
...
   border-collapse: collapse;
   border-spacing: 0;
 }
 
-.ax-chart-table__cell {
-  padding-top: var(--cmp-chart-table-row-padding);
-  padding-bottom: var(--cmp-chart-table-row-padding);
+.ax-chart-table__rows-container {
+  /* prevents DataPoints from being clipped on right edge */
+  padding-right: calc(var(--spacing-grid-size) * 2);
```
HHogg: I realise this is currently a problem, but this has a couple of issues. 

1) `calc(var(--spacing-grid-size) * 2);` doesn't  really related to the DataPoints size because they are dynamic 
2) The ties ChartTable to being used with the DotPlots when it's also to be used with other chart types. like bars.... 

screen shot 2017-05-08 at 09 04 18

I've had a little think about this issue, and it's not great but it does feel like the `ChartTable` needs a `visualCellInsetProperty`, it's a little more manual but it would make it more detachable from specific charts and extendable in future.

src/charts/chart-table/ChartTable.js
```diff
@@ -7,28 +7,61 @@ import './ChartTable.css';
 export default class ChartTable extends Component {
   static propTypes = {
     children: PropTypes.node.isRequired,
+    collapsedVisibleRowCount: PropTypes.number,
+    collapsible: PropTypes.bool,
```
HHogg: Could these be condensed into one property? ...  _Brace yourself for pseudo code_ 

```
defaultProps.collapsedVisibleRowCount = 0;

if collapsedVisibleRowCount !== 0 
  then it's collapsible 
else 
  it's not been set so not collapsible 
```

src/charts/chart-table/ChartTable.js
```diff
@@ -7,28 +7,61 @@ import './ChartTable.css';
...
   };
 
+  constructor(props) {
+    super(props);
+    this.state = {
+      isExpanded: !this.props.collapsible,
+      rowsCount: 0,
+    };
+    this.setRowsCount = rowsCount => this.setState({ rowsCount });
+    this.toggleExpand = () => this.setState(state => ({ isExpanded: !state.isExpanded }));
```
HHogg: Is there a reason for adding these in the constructor and not class methods? I'm not saying either one is wrong, correct or even different but the pattern in Axiom has been to add them as class methods.

src/charts/chart-table/ChartTable.js
```diff
@@ -7,28 +7,61 @@ import './ChartTable.css';
...
+    isExpanded: PropTypes.bool.isRequired,
     labelColumnWidth: PropTypes.string,
+    rowsCount: PropTypes.number.isRequired,
+    setRowsCount: PropTypes.func.isRequired,
+    toggleExpand: PropTypes.func.isRequired,
   };
 
   static defaultProps = {
-    labelColumnWidth: '10rem',
+    labelColumnWidth: '11rem',
```
HHogg: I'm assuming this was changed for the specific use case in the examples? It might be better to set this property in the example because if it keeps getting tweaked it might unexpectedly change the applications Axiom is being consumed in.

src/charts/chart-table/ChartTable.js
```diff
@@ -7,28 +7,61 @@ import './ChartTable.css';
...
 
     return (
-      <Base { ...omit(rest, ['labelColumnWidth']) }
+      <Base
+          { ...omit(rest, [
+            'collapsedVisibleRowCount',
+            'collapsible',
+            'expandButtonSuffix',
+            'labelColumnWidth',
+          ]) }
```
HHogg: RNABD but it might be nicer to define this up something like ... 

```jsx
const { children, ...rest } = this.props;
const props = omit(rest, [ ... ]);

<Base { ...props } />
```

src/charts/chart-table/ChartTableGrid.css
```diff
@@ -7,5 +9,8 @@
   top: 0;
   right: 0;
   bottom: 0;
+
+  /* prevents DataPoints from being clipped on right edge */
+  padding-right: calc(var(--spacing-grid-size) * 2);
```
HHogg: Same comment above, relating to detaching from the dot plot example. Could this be a `fix()` in a separate PR?

src/charts/chart-table/ChartTableRows.js
```diff
@@ -1,22 +1,62 @@
...
-          Component="table"
-          className="ax-chart-table__rows">
-        <tbody>
-          { children }
-        </tbody>
+          Component="div"
+          className="ax-chart-table__rows-container"
+          style={ { height } }>
+        <table className="ax-chart-table__rows">
+          <tbody ref={ tbody => this.tbody = tbody }>
```
HHogg: As discussed about animations. Either need to fix the responsiveness of setting `auto` after the transition ends or add back in when we do the animation sweep.

src/charts/chart-table/ChartTable.js
benji6: Great point!

src/charts/chart-axis/ChartAxisLabels.css
```diff
@@ -2,6 +2,7 @@
 
 .ax-chart-axis-labels {
   display: flex;
+  flex-grow: 1;
```
benji6: Yeah you have a good point here. I think I can achieve the same effect with `width: 100%`

src/charts/chart-table/ChartTable.js
benji6: I like doing this because they are bound to the instance. It will be quicker to render too as you lose the overhead of either function creation or function binding in the render method. I suppose you could bind in the constructor but then I don't really like having the function defined on the class and in the instance.

src/charts/chart-table/ChartTableGrid.css
HHogg: Will be picked up in the following Bar Charts PR

PR #181, opened 2017-05-05
chore(StyleGuide): minify js in styleguide build

### LINKED ISSUE #72: UglifyJS2 for style-guide errors
It doesn't stop the deploy, but minifying is failing ... find out why.

```
ERROR in ./assets/bundle.4d44980acc7f24e14fa9.min.js from UglifyJs
SyntaxError: Unexpected token: operator (>) [./~/get-own-enumerable-property-symbols/index.js:1,0]
```

benji6:
Fixes #72 

js bundle down from 3mb to 1.2mb before gzip 🎉

### DIFF
--- a/webpack.static.config.js
+++ b/webpack.static.config.js
@@ -99,6 +99,11 @@ module.exports = {
         NODE_ENV: JSON.stringify('production'),
       },
     }),
+    new webpack.optimize.UglifyJsPlugin({
+      mangle: {
+        keep_fnames: true,
+      },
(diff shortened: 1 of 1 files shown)


HHogg:
For example ... 
screen shot 2017-05-08 at 08 26 23

benji6:
Thanks @HHogg great catch - I've amended the commit and the docs are non-mangled now! size is now 1.4mb before gzip which is still awesome!

PR #185, opened 2017-05-09
fix(DataPickerHeader): Fixes issue with overflowing headers

thatguynamedandy:


src/components/grid/Grid.css
```diff
@@ -14,14 +14,16 @@
...
-  flex-wrap: wrap;
 }
 
 .ax-grid__cell {
   flex: 1 1 0%;
   min-width: 0;
 }
 
+.ax-grid--wrap { flex-wrap: wrap; }
+.ax-grid--no-wrap { flex-wrap: nowrap; }
```
HHogg: This class doesn't need adding because the initial value of `flex-wrap` is `nowrap`

PR #186, opened 2017-05-09
Bar Charts

HHogg:
This covers a few variations needed in bar/column charts. 

screen shot 2017-05-09 at 12 07 00

screen shot 2017-05-09 at 12 07 05

screen shot 2017-05-09 at 12 07 13

screen shot 2017-05-09 at 12 07 37


src/charts/bars/Bar.js
```diff
@@ -0,0 +1,77 @@
...
+  };
+
+  static contextTypes = {
+    direction: PropTypes.string.isRequired,
+  };
+
+  render() {
+    const { direction } = this.context;
+    const { color, onClick, minSize, percent, showLabel, size, ...rest } = this.props;
+    const isVertical = direction === 'up' || direction === 'down';
```
thatguynamedandy: I tend to write these like the below, purely anecdotal. I'm actually not sure which I prefer.
`['up','down'].includes(direction)`

src/charts/bars/Bar.js
```diff
@@ -0,0 +1,77 @@
...
+      height: !isVertical && size,
+      minHeight: !isVertical && !size && minSize,
+      width: isVertical && size,
+      minWidth: isVertical && !size && minSize,
+    };
+
+    return (
+      <Base { ...rest }
+          className={ classes }
+          onClick={ onClick }
```
thatguynamedandy: can onClick just be passed with the `rest`?

thatguynamedandy:
Soo many chart variations, looks ace 📊

src/charts/bars/Bar.js
benji6: there is a (very minor) perf hit to creating an array then iterating over it though. Dunno how fussed we would be about that

benji6: I just realized that you can't use `includes` here in either case because of browser support

src/charts/bars/Bar.js
HHogg: It's extracted in the destructure above to add the `clickable` class. This could be changed to just `this.props.onClick`, but currently destructuring props when they are used is the pattern 😕

src/charts/bars/Bars.js
```diff
@@ -0,0 +1,48 @@
...
+  }
+
+  render() {
+    const { children, direction, label, ...rest } = this.props;
+    const classes = classnames('ax-bars', `ax-bars--${direction}`);
+
+    return (
+      <Base { ...rest } className={ classes }>
+        <div className="ax-bars__bars-container">
+          <div className="ax-bars__bars">
```
pmsorhaindo: Not too keen on the block and element having the same name.. But at the same time I don't have a useful alternative name.

src/charts/bars/example/chart-table.js
```diff
@@ -0,0 +1,63 @@
...
+  ChartKeyItem,
+  ChartTable,
+  ChartTableAxis,
+  ChartTableGrid,
+  ChartTableKey,
+  ChartTableRows,
+  ChartTableRow,
+  ChartTableLabel,
+  ChartTableVisual,
+  DataPoints,
```
pmsorhaindo: Alphabetical import order.

HHogg: Updated.

PR #187, opened 2017-05-09
feat(ColorPicker): Adding ColorPicker component

pmsorhaindo:


src/components/color-picker/ColorPicker.js
```diff
@@ -0,0 +1,81 @@
...
+        <DropdownTarget>
+          <ColorPickerOption color={ selected } />
+        </DropdownTarget>
+
+        <DropdownContent>
+          <Context width={ width }>
+            <ContextBox>
+              <Grid gutters="small" horizontalAlign="center">
+                { options.map((color => (
+                  <GridCell shrink={ true }>
```
HHogg: Does this not complain about a missing key?

src/components/color-picker/ColorPicker.js
```diff
@@ -0,0 +1,81 @@
...
+                    <ColorPickerOption
+                        color={ color }
+                        onClick={ () => { onSelectColor(color); } } />
```
HHogg: `onSelectColor` isn't required in the propTypes so calling it when it's not given will throw and error. Should it be required?

src/components/color-picker/ColorPicker.js
```diff
@@ -0,0 +1,81 @@
...
+    width: '10.5rem',
+  };
+
+  render() {
+    const { onSelectColor, options, selected, width, ...rest } = this.props;
+
+    return (
+      <Dropdown position="down" { ... rest }>
+        <DropdownTarget>
+          <ColorPickerOption color={ selected } />
```
HHogg: Could the size of the displayed value be dynamic with a size prop?

src/components/color-picker/ColorPickerOption.css
```diff
@@ -0,0 +1,88 @@
+@import '../../materials/colors';
+@import '../../materials/sizing';
+
+.ax-color-picker__option {
+  border-radius: 50%;
+  height: 1rem;
+  width: 1rem;
+  cursor: point;
```
HHogg: This should be `pointer`?

src/components/color-picker/ColorPickerOption.css
```diff
@@ -0,0 +1,88 @@
+@import '../../materials/colors';
+@import '../../materials/sizing';
+
+.ax-color-picker__option {
+  border-radius: 50%;
+  height: 1rem;
+  width: 1rem;
```
HHogg: As mentioned above could this be dynamic?

src/components/color-picker/ColorPickerOption.css
```diff
@@ -0,0 +1,88 @@
...
+
+.ax-color-picker__option--grey {
+  background-color: var(--color-brand-grey);
+  &:hover  { background-color: var(--color-brand-grey--light); }
+  &:active { background-color: var(--color-brand-grey--dark); }
+}
+
+.ax-color-picker__option--empty {
+  color: transparent;
+  border: 4px solid var(--color-ui-light);
```
HHogg: `0.25rem` 😉  it's probably directly related to the grid sizing though? So maybe `var(--spacing-grid-size)`?

src/components/color-picker/ColorPickerOption.css
```diff
@@ -0,0 +1,88 @@
...
+}
+
+.ax-color-picker__option--grey {
+  background-color: var(--color-brand-grey);
+  &:hover  { background-color: var(--color-brand-grey--light); }
+  &:active { background-color: var(--color-brand-grey--dark); }
+}
+
+.ax-color-picker__option--empty {
+  color: transparent;
```
HHogg: This probably isn't needed because it's `color` not `background-color` and if `background-color` it will appear transparent.

src/components/color-picker/ColorPickerOption.css
```diff
@@ -0,0 +1,88 @@
...
+  border: 4px solid var(--color-ui-light);
+  &:hover  { border-color: var(--color-ui-light--light); }
+  &:active { border-color: var(--color-ui-light--dark); }
```
HHogg: These colors should be using the context colors so they work on the dark theme. Looking at the designs the non-interactive color is `--color-context-background-subtle` ... thought there are no interaction colors given from design. Something to talk to them about.

src/components/color-picker/ColorPickerOption.js
```diff
@@ -0,0 +1,33 @@
...
+      'teal',
+      'green',
+      'chartreuse',
+      'amber',
+      'orange',
+      'brown',
+      'grey',
+    ]),
+    height: PropTypes.string,
+    width: PropTypes.string,
```
HHogg: It looks like the size is dynamic so maybe add the defaults here rather than in the CSS?

HHogg: Though these could probably be combined into a `size` prop as it's a circle 😉

PR #190, opened 2017-05-10
fix(ChartTable): Collapsible behaviour

HHogg:
Fixes:
* Removes lodash as **it's not a dependency**
* Sets height to `auto` when ChartTable is collapsable and is in the expanded state.
* Removes collapsable behaviour when no collapsable props are set. 
* Removes a workaround that was added to pass tests.


HHogg:
*Note to self: Learn to spell ... and trust my spell checker.* 

Thanks @benji6 😉

PR #192, opened 2017-05-10
feat(Chart): Adds Chart layout component

pmsorhaindo:


src/charts/chart/Chart.css
```diff
@@ -0,0 +1,13 @@
+@import '../../materials/colors';
+@import '../../materials/sizing';
+
+.ax-chart__header {
+  border-bottom-width: var(--component-border-width-small);
+  border-bottom-style: solid;
+  border-bottom-color: var(--color-ui-light--dark);
```
HHogg: This won't work on the dark theme. It should probably be the 15% or 20% context colour. Double check with design as it looks like it might be marked up wrong in the designs.

src/charts/chart/Chart.css
```diff
@@ -0,0 +1,13 @@
...
+  padding-bottom: calc(var(--spacing-grid-size) * 2.5);
```
HHogg: Need to query this value to find out why it's not following 4x grid pattern.

src/index.js
```diff
@@ -101,6 +101,10 @@ export { default as Weak } from './components/typography/Weak';
 export { default as Bar } from './charts/bar-chart/Bar';
 export { default as BarChart } from './charts/bar-chart/BarChart';
 export { default as BarGroup } from './charts/bar-chart/BarGroup';
+export { default as Chart } from './charts/chart/Chart';
+export { default as ChartBody } from './charts/chart/ChartBody';
+export { default as ChartHeader } from './charts/chart/ChartHeader';
+export { default as ChartTitle } from './charts/chart/ChartTitle';
```
HHogg: I think it makes sense that these are grouped together, but currently the sections are alphabetised, probably best to keep it that way for the moment.

Issue #110, opened 2017-03-01
Deploy pull request changes to a public url


thatguynamedandy:
In order that we can be sure we are happy with a feature before merging it we should be able to preview it, ideally without the hassle of pulling down forks off different branches.

We should look at trying to deploy pull requests onto Github pages, potentially under a sub directory. We should then be sure any sub directories are cleaned up once a branch is merged/closed.

siggysamson:
That's something we definitely should be working on! I ran into exactly this issue with my recent PR.

I like the idea of using Github Pages for it because that's the easiest to use solution. Some thoughts around this:
- When I was setting up the deploy to Github Pages I ran into a couple issues with the base url so I'm not sure if we might experience issues because we are going to store everything in a subdirectory. I _think_ we should be able to solve this by adding the proper publicPath though so might not be an issue any more.
- We should always make sure that the top level dir contains the latest from master.
- Webhooks should help us detecting closed PRs so removing already deployed branches should be possible

HHogg:
> When I was setting up the deploy to Github Pages I ran into a couple issues with the base url

This also relates to #90, I have a WIP branch at the moment to remove the basepath... I didn't realise it has an evil side to it that causes all sorts of mayhem 😸

thatguynamedandy:
I've come close to getting the first part done with this commit...
https://github.com/thatguynamedandy/axiom/commit/090c5d96175919363aac62bf67b6951f5af6af2b

The idea is that branches are just deployed to a sub directory, it works, however there is one problem. The `gh-pages` module has an `add` command, this needs to be set to prevent the master deploy deleting all of the branch deploys, sadly though it means any removed files will remain on the branch. There doesn't appear to be way currently to get around it, which leads to the following set of options...

- Create a PR into gh-pages which helps to cover our use case
- Live with the fact that there will be orphaned files in the gh-pages branch
- Use something else altogether https://surge.sh/ seems to have gained some traction

thatguynamedandy:
https://stackshare.io/stackups/github-pages-vs-surge-vs-netlify

siggysamson:
Regarding the orphaned files: Could we come up with a script that runs whenever the master branch is build to compare the directories against the set of open PRs?

- Send a request to GH API
- Get a list of all open PRs
- Compare against folders in `pull-request`
- Delete the ones that don't match.

For now we could just live with orphaned files and clean it up occasionally.

[...]

PR #195, opened 2017-05-10
chore(Netlify): Netlify config file

thatguynamedandy:
This is actually all it takes to get going on netlify.

https://www.netlify.com

We would need to get the integration approved for BrandwatchLtd. From my understanding this will deploy PR's and branches. But we don't generally have branches as we generally fork the repo, It shouldn't deploy forked branches, but if people want to they can setup the integration themselves.

#110

### DIFF
--- a/netlify.toml
+++ b/netlify.toml
@@ -1,0 +1,4 @@
+# Global settings applied to the whole site.
+[build]
+  publish = "static"
+  command = "yarn build:static"


thatguynamedandy:
There is still a bit of wrangling over whether we can approve third party access to Netlify, In the meantime this PR could probably still go in. As it is still possible to approve Netlify for your own fork and the relevant branch should get deployed. It's not the slickest but it would still enable sharing in the interim.

HHogg:
I've spoken to Netlify and they've given me some steps to authenticate just this repo.

HHogg:
Unfortunately it looks like Netlify doesn't support deploying from forks. It might be we provide setup instructions for our forks, but am closing this for the moment as further investigations are needed.

PR #197, opened 2017-05-11
fix(Context): Give Context a max-height

benji6:
screen shot 2017-05-11 at 16 55 16
screen shot 2017-05-11 at 16 55 31

### DIFF
--- a/src/components/context/Context.css
+++ b/src/components/context/Context.css
@@ -17,7 +17,11 @@
   border-width: var(--component-border-width);
   border-style: solid;
   border-radius: var(--component-border-radius);
-  overflow: hidden;
+  overflow-y: auto;
+  /* stylelint-disable */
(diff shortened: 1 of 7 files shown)


src/components/context/Context.css
```diff
@@ -17,7 +17,7 @@
   border-width: var(--component-border-width);
   border-style: solid;
   border-radius: var(--component-border-radius);
-  overflow: hidden;
+  overflow-y: auto;
```
HHogg: This will cause an issue with the content overflowing the border-radius (as discussed).

benji6: fixed with the magical `-webkit-mask-image: -webkit-radial-gradient(white, black);`

HHogg: 😄 Pretty magical

PR #204, opened 2017-05-15
fix(Global CSS): Moves globally applied styles to a separate file 

HHogg:
This is to provide a way for those styles to be ignored through import overriding.

Partially linked to #176.

### DIFF
--- a/src/components/form/TextArea.css
+++ b/src/components/form/TextArea.css
@@ -7,7 +7,6 @@
   border: var(--component-border-width) solid var(--cmp-input-border-color);
   border-radius: var(--component-border-radius);
   background-color: var(--cmp-input-background-color);
-  overflow: auto; /* Remove the default vertical scrollbar in IE. */
   resize: none;
   transition-property: border-color;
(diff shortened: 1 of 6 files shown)


src/components/form/TextArea.css
```diff
@@ -7,7 +7,6 @@
   border: var(--component-border-width) solid var(--cmp-input-border-color);
   border-radius: var(--component-border-radius);
   background-color: var(--cmp-input-background-color);
-  overflow: auto; /* Remove the default vertical scrollbar in IE. */
```
larister: I'm guessing you pulled these up as they make more sense as globals?

jdelight:
Yes this worked for me 👍 nice one @HHogg

src/components/form/TextArea.css
HHogg: Yeah, that were moved into the ax scoped classes in #176 , but really affect these elements globally, so to protect future elements not in these classes (for example `<button>` elements are also used in the Tabs, Menu and Button) I moved them to the now optional global styles.

PR #208, opened 2017-05-17
wip add separated prop to bar charts

jdelight:
wip: adds a `separated` property to `Bars` component to determine if the bars should be contiguous.

### DIFF
--- a/src/charts/bars/Bars.css
+++ b/src/charts/bars/Bars.css
@@ -36,6 +36,11 @@
     width: 100%;
     height: 100%;
   }
+
+  &.ax-bars--separated .ax-bars__bar-rect {
+    margin-left: 1px;
+    margin-right: 1px;
+  }
 }
 
 .ax-bars--up,
@@ -56,6 +61,11 @@
 .ax-bars--left {
   & .ax-bars__bars {
     flex-direction: column;
+  }
+
+  &.ax-bars--separated .ax-bars__bar-rect {
+    margin-top: 1px;
+    margin-bottom: 1px;
   }
 }
 
--- a/src/charts/bars/example/vertical-bars.js
+++ b/src/charts/bars/example/vertical-bars.js
@@ -12,7 +12,7 @@ export default class BarsExample extends Component {
             { data.map(({ data }, index) =>
               <GridCell key={ index } snippetIgnore={ true }>
                 <div snippetIgnore={ true } style={ { height: '10rem' } }>
-                  <Bars direction="up" snippetSkip={ index !== 0 }>
+                  <Bars direction="up" separated={true} snippetSkip={ index !== 0 }>
                     { data.map((value, index) =>
                       <Bar
                           color={ labels[index].color }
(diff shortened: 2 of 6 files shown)


src/charts/bars/Bars.css
```diff
@@ -36,6 +36,11 @@
     width: 100%;
     height: 100%;
   }
+
+  &.ax-bars--separated .ax-bars__bar-rect {
```
HHogg: There is a parent element with the class `ax-bars__bar` which would better with this styling on, because `.ax-bars__bar-rect` has a sibling element for the label, and they should probably be shifted together so to keep the label central to the bar.

src/charts/bars/Bars.css
```diff
@@ -36,6 +36,11 @@
...
+    margin-left: 1px;
+    margin-right: 1px;
```
HHogg: This will double up the margins between bars (if 2px isn't intentional) and also add a margin on left side of the first bar and the right side or the last bar. I don't think this is correct because it won't give the appearance of the bars lining up with content that may be above or below. A better alternative to this would be to add it to one side and remove it from the last/first child, or with the sibling selector. Like...

```css
.ax-bars--separated .ax-bars__bar {
  margin-right: 1px;

  &:last-child {
     margin-right: 0 
  }
}

/* OR */

.ax-bars--separated .ax-bars__bar + .ax-bars__bar {
  margin-left: 1px;
}
```

src/charts/bars/Bars.css
```diff
@@ -57,6 +62,11 @@
   & .ax-bars__bars {
     flex-direction: column;
   }
+
+  &.ax-bars--separated .ax-bars__bar-rect {
+    margin-top: 1px;
```
HHogg: We use rems throughout Axiom :) So `1px` will be `0.0625rem` (`1/16`)

src/charts/bars/Bars.css
```diff
@@ -57,6 +62,11 @@
...
+    margin-bottom: 1px;
```
HHogg: As this value is repeated a few times throughout the file, it would be better being defined as a variable at the top of the file for easy tweaking in the future. 

```css
:root {
  --cmp-bars-separation-size: 0.0625rem;
}
```

src/charts/bars/example/vertical-bars.js
```diff
@@ -12,7 +12,7 @@ export default class BarsExample extends Component {
             { data.map(({ data }, index) =>
               <GridCell key={ index } snippetIgnore={ true }>
                 <div snippetIgnore={ true } style={ { height: '10rem' } }>
-                  <Bars direction="up" snippetSkip={ index !== 0 }>
+                  <Bars direction="up" separated={true} snippetSkip={ index !== 0 }>
```
HHogg: Until we have the editable props example (can't wait), it might be nicer to show a separate example with separated styling for all the directions, so as to still show the most basic example without the separation styling.

src/charts/bars/Bars.css
jdelight: Makes sense thanks 👍

src/charts/bars/Bars.css
jdelight: Nice CSS fu - like it 👍

jdelight:
Thanks @HHogg - learned a bunch of stuff :) will get these fixes in

src/charts/bars/Bars.css
HHogg: I promise no fu intended 😉

PR #215, opened 2017-05-24
Rework dot plot api

benji6:
# Dot Plot Overhaul

## What I've Done

Before if a user wanted to create a dot plot chart they would have to manually assemble a large number of heavily nested components. This gave the user a lot of flexibility but also a lot of complexity and a lot of responsibility for implementation. I have now consolidated creating a dot plot chart into a single `DotPlotChart` component that is heavily encapsulated and handles a lot of behaviours automatically. This is less customizable (which may be desirable from the view of explicit design constraints) and also much simpler and easier to use.

If people like this API we should do the same for bar charts too and I imagine we will no longer expose any of the `ChartTable` components.

### The data structure

I have assumed sorting will always be handled by `DotPlotChart` and the user will have no control over this - hence using key-values in an object. This is in keeping with all existing designs. If we wanted the data to be user-sortable the `value` format should probably be an array something like this:

```js
[
  { label: 'Label A', value: 0 },
  { label: 'Label B', value: 90 },
  { label: 'Label C', value: 45 },
]
```

But I don't think we should do it that way unless we _know_ there will be cases where the sorting is not handled by the `DotPlotChart` component.

### The docs

They don't look so good:

screen shot 2017-05-24 at 10 42 22

screen shot 2017-05-24 at 10 41 12

### The tests

The snapshots are thousands of lines! How useful is this?

## Extending the `DotPlotChart` component

In the future we may want to implement popovers. I envisage this happening via a component passed in to `DotPlotChart` (perhaps named something like `popoverComponent`) which will be rendered with props indicating which dot has been clicked on as the user clicks on the chart.

Implementing sorting etc within this component should be trivial :v:


src/charts/dot-plot/DotPlotChart.js
```diff
@@ -0,0 +1,128 @@
...
+        'orange',
+        'brown',
+        'grey',
+      ]).isRequired,
+      name: PropTypes.string.isRequired,
+      values: PropTypes.object.isRequired,
+    })).isRequired,
+    expandButtonSuffix: PropTypes.string,
+    labelColumnWidth: PropTypes.string.isRequired,
+    xLabels: PropTypes.arrayOf(PropTypes.string),
```
HHogg: This might be better renamed to `xAxisLabels`

src/charts/dot-plot/DotPlotChart.js
```diff
@@ -0,0 +1,128 @@
...
+        'teal',
+        'green',
+        'chartreuse',
+        'amber',
+        'orange',
+        'brown',
+        'grey',
+      ]).isRequired,
+      name: PropTypes.string.isRequired,
+      values: PropTypes.object.isRequired,
```
HHogg: Following on from your description (and as we've discussed), you are right that this would better as an array because of the custom sorting scenarios that are required outside of the knowledge of the DotPlotChart.

PR #216, opened 2017-05-24
feat(Position): Adds option to specify where position nodes are appended

larister:
By default this will be document.body, but when Axiom is being used
within a wider application which uses a wrapper node, we need to ensure
the Position node remains within the wrapper.

Open to alternative approaches

### DIFF
--- a/src/components/position/Position.js
+++ b/src/components/position/Position.js
@@ -12,6 +12,7 @@ export default class Position extends Component {
     children: PropTypes.array.isRequired,
     isVisible: PropTypes.bool.isRequired,
     offset: PropTypes.oneOf(['start', 'middle', 'end']),
+    parentNode: PropTypes.object,
     position: PropTypes.oneOf(['top', 'right', 'bottom', 'left']),
     onMaskClick: PropTypes.func,
   };
@@ -67,9 +68,12 @@ export default class Position extends Component {
   }
 
   createReactRootNode() {
+    const parentNode = this.props.parentNode || document.body;
     this._reactRootNode = document.createElement('div');
     this._reactRootNode.classList.add('AxiomPositionRoot');
-    document.body.appendChild(this._reactRootNode);
+
+    parentNode.appendChild(this._reactRootNode);
+
     this.renderSubtree();
   }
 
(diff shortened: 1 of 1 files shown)


src/components/position/Position.js
```diff
@@ -11,6 +11,7 @@ export default class Position extends Component {
   static propTypes = {
     children: PropTypes.array.isRequired,
     isVisible: PropTypes.bool.isRequired,
+    nodeToAppendTo: PropTypes.object,
```
benji6: maybe `parentNode` is a better name?

larister: Hmm maybe, my reasoning was that as this already had a "root node" it might be confusing to have both a root node and a parent node. In hindsight though, maybe `nodeToAppendTo` isn't much better! Will change

larister:
@benji6 forgive me but I rebased; it's just a small PR so thought it'd be ok :)

src/components/position/Position.js
```diff
@@ -24,6 +25,8 @@ export default class Position extends Component {
   constructor(props) {
     super(props);
 
+    this.parentNode = props.parentNode || document.body;
```
benji6: I forgive you for the rebase, but I should apologise because really defaults should be specified on the static `defaultProps` object (as above in this module). Main reason is for the auto-generated API docs, but good for consistency too!

larister: Tried that, but sadly webpack complains. Have taken a different tack which isn't ideal, but keeps Travis happy. Lemme know your thoughts

src/components/position/Position.js
```diff
@@ -12,6 +12,7 @@ export default class Position extends Component {
     children: PropTypes.array.isRequired,
     isVisible: PropTypes.bool.isRequired,
     offset: PropTypes.oneOf(['start', 'middle', 'end']),
+    parentNode: PropTypes.object,
```
jdelight: NABD - is there a `node` propType? perhaps `object` is fine.

HHogg: I think `node` is for React Nodes. Anything that can be rendered by react, numbers, string, elements (React elements).

PR #217, opened 2017-05-24
feat(Scripts): Adds a build:link script for use with npm link

larister:
Just a suggestion, it's fine if a) there's a better way or b) it's not deemed useful enough. It just took me a little while & a little help to figure out.

### DIFF
--- a/package.json
+++ b/package.json
@@ -18,6 +18,7 @@
   "scripts": {
     "build:css": "NODE_ENV=production BABEL_ENV=production webpack --config webpack.css.config.js",
     "build:static": "NODE_ENV=production BABEL_ENV=production webpack --config webpack.static.config.js",
+    "build:local": ". ./deploy/travis_setup.sh && yarn prepublish",
     "ci:build:static": "BASENAME_ENV=\"'/axiom/'\" yarn build:static",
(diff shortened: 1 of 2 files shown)


package.json
```diff
@@ -18,6 +18,7 @@
   "scripts": {
     "build:css": "NODE_ENV=production BABEL_ENV=production webpack --config webpack.css.config.js",
     "build:static": "NODE_ENV=production BABEL_ENV=production webpack --config webpack.static.config.js",
+    "build:link": ". ./deploy/travis_setup.sh && npm run prepublish",
```
benji6: This should probably be `yarn prepublish` instead of `npm run prepublish`

HHogg: I agree with @benji6 comments about renaming from `link` and to use `yarn` 😉 ... also I wonder if `travish_setup.sh` needs renaming to something else like `prebuild.sh` ¯\_(ツ)_/¯ ...

larister:
Brill thanks both, have updated accordingly.

HHogg:
Thanks @larister. All is good to go except for the commit message, a `feat()` will trigger a new release so as it's purely a development enhancement it should really be a `chore()`

larister:
Ah thanks @HHogg, fixed

PR #221, opened 2017-06-01
feat(ColorPicker): Adds a disabled state to ColorPickerOptions

pmsorhaindo:
screen shot 2017-06-02 at 11 44 45 am

### DIFF
--- a/src/components/color-picker/ColorPicker.js
+++ b/src/components/color-picker/ColorPicker.js
@@ -13,6 +13,7 @@ import {
 
 export default class ColorPicker extends Component {
   static propTypes = {
+    disabledOptions: PropTypes.arrayOf(PropTypes.string),
     options: PropTypes.array,
     selected: PropTypes.oneOf([
       'rose',
@@ -33,6 +34,7 @@ export default class ColorPicker extends Component {
   };
 
   static defaultProps = {
+    disabledOptions: [],
     options: [
       'rose',
       'pink',
@@ -51,7 +53,7 @@ export default class ColorPicker extends Component {
   };
 
   render() {
-    const { onSelectColor, options, selected, width, ...rest } = this.props;
+    const { disabledOptions, onSelectColor, options, selected, width, ...rest } = this.props;
 
     return (
       <Dropdown position="bottom" { ... rest }>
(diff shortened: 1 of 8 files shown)


src/components/color-picker/ColorPicker.js
```diff
@@ -70,6 +72,9 @@ export default class ColorPicker extends Component {
                   <GridCell key={ color } shrink={ true }>
                     <ColorPickerOption
                         color={ color }
+                        disabled={ Boolean(disabledColors.find((disabledColor) => {
```
benji6: `find` isn't available in IE11

pmsorhaindo: 😩  IE will be the death of me.

src/components/color-picker/ColorPicker.js
```diff
@@ -66,13 +68,18 @@ export default class ColorPicker extends Component {
...
-                  </GridCell>
-                ) ) ) }
+                { options.map(((color)=> {
+                  return (
+                    <GridCell key={ color } shrink={ true }>
+                      <ColorPickerOption
+                          color={ color }
+                          disabled={ disabledColors.indexOf(color) === -1 }
+                          onClick={ () => {
+                            return !(disabledColors.indexOf(color) === -1) && onSelectColor(color);
```
HHogg: Minor, but you could avoid the extra `!` by `disabledColors.indexOf(color)  >= 0`. 

It might also be a bit nicer to read with an implicit return 

```jsx
onClick={ () => disabledColors.indexOf(color) >= 0 && onSelectColor(color) }
```

benji6: creating an intermediate `isColorDisabled` variable would help with readability and perf here as the `indexOf` operation is needlessly performed twice

```js
onClick={ () => isColorDisabled || onSelectColor(color) }
```

src/components/color-picker/example/disabled.js
```diff
@@ -0,0 +1,35 @@
...
+
+  onSelectColor(color) {
+    this.setState({
+      currentColor: color,
+    });
+  }
+
+  render() {
+    const { currentColor } = this.state;
+    const disabledColors = ['rose', 'purple', 'teal', 'chartreuse', 'amber', 'brown'];
```
benji6: lift this up to the top of the module so it isn't created on every render cycle

src/components/color-picker/example/disabled.js
```diff
@@ -0,0 +1,35 @@
...
+
+    return (
+      <Example name="Color Picker (with disabled options)">
+        <Snippet>
+          <ColorPicker
+              disabledColors={ disabledColors }
+              onSelectColor={ this.onSelectColor.bind(this) }
```
benji6: I'm wondering whether we should start binding in the constructor rather than in the render method - would definitely be better for perf as `bind` has always been a costly operation (wrapping in a lambda is better from a performance perspective)

pmsorhaindo: Ahhh I set out on a  mission once to remove all instances of these and add a lint rule, but certain scenarios would have involved splitting up components to avoid binding in the render function. But yeah, in this case it's a no brainer, nice spot!

PR #222, opened 2017-06-01
feat(DataPicker): Adds clearable behaviour through `onClear` property

HHogg:
screen shot 2017-06-01 at 20 11 22

BREAKING CHANGE:

* DataPicker component now controls all of the props of the component as opposed to DataPickHeader having props.
* `headerText` has been split into `value` and `placeholder` properties
* `isInvalid` is no longer needed and is derived from `value`

This breaking change was needed to provide DataPickHeader with a state where a value had been selected in order to show the CLEAR action.

### DIFF
--- a/src/charts/data-picker/example/clearable.js
+++ b/src/charts/data-picker/example/clearable.js
@@ -26,10 +26,11 @@ export default class DataPickerExample extends Component {
     const selected = getData().find(({ id }) => id === selectedId);
 
     return (
-      <Example name="Data Picker with ColorPicker">
+      <Example name="Clearable Data Picker">
         <Snippet>
           <DataPicker
               color={ selectedColor || (selected && selected.color) }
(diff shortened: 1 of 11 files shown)


src/charts/data-picker/example/clearable.js
```diff
@@ -0,0 +1,61 @@
...
+    super();
+    this.state = {
+      selectedColor: undefined,
+      selectedId: undefined,
+    };
+  }
+
+  render() {
+    const { selectedColor, selectedId } = this.state;
+    const selected = getData().find(({ id }) => id === selectedId);
```
pmsorhaindo: I don't think we can use find in IE11.. 😩

HHogg: It's fine in the examples as the style guide has the polyfill 😉

pmsorhaindo: ah! in that case, everything LGTM!

PR #223, opened 2017-06-05
DotPlot mouseover behaviour 

HHogg:
dotplot

*I've removed the snapshot tests for the DotPlot because I'm not sure how much value they bring because of how large they are. I'm happy to add them back in if anybody objects.*


src/charts/chart-table/ChartTable.css
```diff
@@ -59,20 +74,11 @@
...
-  overflow: visible;
-}
-
 .ax-chart-table__grid-line {
-  stroke: var(--color-context-border-subtle);
-  vector-effect: non-scaling-stroke;
+  position: absolute;
+  top: 0;
+  bottom: 0;
+  border-right: 0.0625rem dotted var(--color-context-border-subtle);
```
HHogg: No I think the dot-plot-line-width is the distance line.... this is just a the grid line. I'll still move up into a variable thought 👍

src/charts/dot-plot/DotPlot.css
```diff
@@ -1,7 +1,73 @@
...
+  pointer-events: none;
+}
+
+.ax-dot-plot__line-container {
+  position: absolute;
+  width: 100%;
+  height: 100%;
+}
+
+.ax-dot-plot__line {
```
HHogg: Yeah it's still used inside the <DotPlotLine> component.

PR #224, opened 2017-06-06
fix(ColorPicker): align ColorPickerOptions disabled state to the styl…

pmsorhaindo:
…e guide

### DIFF
--- a/src/charts/data-picker/DataPickerHeader.css
+++ b/src/charts/data-picker/DataPickerHeader.css
@@ -4,6 +4,7 @@
 .ax-data-picker__header-title {
   display: inline-flex;
   align-items: center;
+  max-width: 100%;
 }
 
 .ax-data-picker__header-title-text {
--- a/src/components/color-picker/ColorPickerOption.css
+++ b/src/components/color-picker/ColorPickerOption.css
@@ -2,7 +2,12 @@
 @import '../../materials/colors';
 @import '../../materials/sizing';
 
+:root {
+  --cmp-color-picker-border-width: 0.1875rem;
+}
+
 .ax-color-picker__option {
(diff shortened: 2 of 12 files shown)


src/components/color-picker/ColorPickerOption.css
```diff
@@ -27,7 +32,8 @@
 }
 
 .ax-color-picker__option--disabled {
-  opacity: 0.1;
+  background-color: var(--color-ui-white);
```
HHogg: A white background colour won't work for the dark theme. Should this be transparent?

pmsorhaindo: Ahh ok! thanks!

src/charts/data-picker/DataPickerHeader.css
```diff
@@ -2,7 +2,7 @@
 @import '../../materials/sizing';
 
 .ax-data-picker__header-title {
-  display: inline-flex;
+  display: flex;
```
HHogg: I think inline-flex is used to prevent the link extended out further than the down chevron, so empty space doesn't appear clickable. What was the reason for changing this?

pmsorhaindo: inline-flex breaks the text-overflow. :/

HHogg: Ah ok, adding `max-width: 100%;` to this should contain it within the available space, and should allow you to keep display `inline-flex`

pmsorhaindo: Ahh I never think about max-width because I used to use it all the time in the past with actually pixel values. :/ max-width: 100% is awesome. Thanks!

PR #230, opened 2017-06-08
chore(StyleGuide): Fixed StyleGuide example rendering issue

HHogg:
Fixes the issue where when state in Examples changed, it re-renders the examples causing components to be unmounted and remounted.


style-guide/utils/example-filter.js
```diff
@@ -1,5 +1,4 @@
-import { createElement } from 'react';
-import { v4 } from 'uuid';
```
benji6: I think we can remove this from dependencies now?

PR #232, opened 2017-06-08
feat(DotPlotChart): Make key/label optional

larister:
Granted you generally want a key/label, but there are rare circumstances it's necessary to hide them.


src/charts/dot-plot/DotPlotChart.js
```diff
@@ -127,20 +149,7 @@ export default class DotPlotChart extends Component {
...
-                  <DataPoint color={ color } />
-                </DataPoints>
-              </ChartKeyItem>
-            ) }
-            <ChartKeyItem label={ chartKeyLineLabel }>
-              <DotPlotLine width="1rem" />
-            </ChartKeyItem>
-          </ChartKey>
-        </ChartTableKey>
+        { getChartTableKey() }
```
HHogg: Chart key is required to enforce the ordering of the stacked dots. This should probably be controlled with a flag like `showKey` (set to true in `defaultProps`). 

We follow the pattern of keeping templates all within the return function so it's easier to read without needing to looking around the component. To do this we just use a short circuit conditional statement like... 

```jsx
{ showKey && (
  <ChartTableKey>
    ....
  </ChartTableKey>
) }
```

benji6: Just to chime in on the temlate format - another reason to do it the way @HHogg suggests is to avoid creating (and possibly invoking) a new function every time `render` is invoked. Generally it's a very small perf cost, but I think it's always best to keep `render` as lean as possible. It's more of a consideration when you know there will be loads of instances though

larister: Thanks both; changing...

larister: Ok changed, thanks again both. Lemme know if I misunderstood

src/charts/dot-plot/DotPlotChart.js
```diff
@@ -51,6 +51,7 @@ export default class DotPlotChart extends Component {
     })).isRequired,
     expandButtonSuffix: PropTypes.string,
     labelColumnWidth: PropTypes.string.isRequired,
+    showKey: PropTypes.Boolean,
```
HHogg: Thanks for making those changes @larister. 

`PropTypes.Boolean`, isn't a valid prop type, it should be `.bool`. 

_Why the bool and func one is abbreviated and the others are not I don't know_

src/charts/dot-plot/example/no-key.js
```diff
@@ -0,0 +1,21 @@
...
+  render() {
+    return (
+      <Example name="Collapsible DotPlotChart without chart key or labels">
+        <Snippet>
+          <DotPlotChart
+              axisTitle=""
+              chartKey={ [] }
+              data={ data }
+              labelColumnWidth="11rem"
+              xAxisLabels={ [] } />
```
HHogg: It looks like you need to update the example here with the `showKey` changes.

HHogg:
A couple more changes need @larister. Also the release notes are generated from the commit messages so this commit needs updating a little bit to reflect the changes. Something like 

```
Adds `showKey` prop to `DotPlotChart` to allow hiding of the key
```

src/charts/dot-plot/DotPlotChart.js
larister: Wtf, sorry Harry I still need to tweak my vim autocomplete obviously :p

src/charts/dot-plot/example/no-key.js
larister: Dang, the example; thanks Harry

PR #237, opened 2017-06-12
feat(DotPlotChart): Adds  prop to  to allow hiding of the axis

larister:
Very similar to https://github.com/BrandwatchLtd/axiom/pull/232; I didn't spot that passing an empty array for the `xAxisLabels` property results in no gridlines. It was a bit of a hack anyway; this approach allows us to explicitly hide the axis.

### DIFF
--- a/src/charts/dot-plot/DotPlotChart.js
+++ b/src/charts/dot-plot/DotPlotChart.js
@@ -51,6 +51,7 @@ export default class DotPlotChart extends Component {
     })).isRequired,
     expandButtonSuffix: PropTypes.string,
     labelColumnWidth: PropTypes.string.isRequired,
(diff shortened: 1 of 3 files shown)


HHogg:
@larister I don't think this should be needed with the latest Chart API changes currently in #236. The grid has be moved out of the what was the `<ChartAxis>` component so now the new `<ChartAxisTitle>` is just the title, which is also been [made conditional here](https://github.com/BrandwatchLtd/axiom/pull/236/files#diff-7714b2d1bd7596d86141cecb97d44e1cR123). These changes have also be reflected in the new `<BarChart>` component also. 

If you agree, please feel free to close this one off.

larister:
Cool thanks @HHogg, closing...

PR #242, opened 2017-06-14
chore(SVG): Set SVG filters on componentWillMount

larister:
Previously, SVG _filter code was being executed only once, when the
Progress/Candytar components were first loaded. The assumption was that
the AxiomSvgDefs would always be there; however in some circumstances
(particularly embedding axiom components within another app) we may want
to remove this element.

This PR simply ensures that the _filter code is applied whenever the
calling component is mounted, rather than required.

@HHogg as discussed :)


src/components/avatar/Candytar.js
```diff
@@ -47,6 +47,8 @@ export default class Candytar extends Component {
     picker: defaultPickerFn,
   };
 
+  componentWillMount = filter
```
thatguynamedandy: The only thing I was wondering is whether it's worth aliasing 'filter' to something else, only because `filter` is a verb and it's not actually doing a filter. it's more just putting a filter on the DOM `renderFilter`, `attachFilter` (not sure)

larister:
Thanks @thatguynamedandy; have amended as suggested. I've left the files as `_filter`; lemme know if you want me to rename them as well :)

PR #243, opened 2017-06-14
chore(Structure): Moved DataPicker, Panel and Pagination out of Chart…

HHogg:
…s and Composites into Components

Some of the components were mistakenly inside Composites and Charts & Visualisations. 

```
<DataPicker>: Charts -> Components
<Panel>: Charts -> Components
<Pagination>: Composites -> Components
```

The way I see something being defined as a composite isn't necessarily something made up by one or more components but something that has a single purpose in context of an application. For example a `<SignUpForm>`, `<LoginPage>`, `<404Page>` or `<UserMenu>`.


thatguynamedandy:
Interesting point, I think I agree with you in that the current composites generally carry out a more specific task than the more generic components 

I think the initial thought that it is just a component made up of more than one component is no longer valid, as most components are now derived from others, begs the question 
> "should we rename composites to something else?"

PR #245, opened 2017-06-16
chore(StyleGuide): Configurable examples

HHogg:
screen shot 2017-06-16 at 16 45 38

At the moment examples are duplicated for every prop variation, which leaves the style guide pages cluttered and difficult to understand what an example is showing. 

This adds a feature to the style guide to configure a single example. It also gives our QA engineers the ability to try and break things :) 

[Deployed here](https://interactive-examples--hhogg-axiom.netlify.com/docs/charts/dot-plot)

_Note: It's just the composites and charts (except ColumnChart) that have been transferred over for the moment_


src/charts/bar-chart/example/example.js
```diff
@@ -66,23 +66,37 @@ class ContextDemoComponent extends Component {
...
+
+    const propTypes = {
+      BarChart: components.BarChart,
+    };
+
+    const initialProps = {
+      BarChart: {
+        axisTitle: '% of each something',
+        chartKey: chartKey,
+        ContextComponent: ContextDemoComponent,
```
pmsorhaindo: SNABD but shouldn't capitalized stuff come after?

HHogg: I think in the propTypes the linting wants them at the beginning 🤔 , I'll update this for consistency 👍

src/composites/pagination/example/example.js
```diff
@@ -0,0 +1,34 @@
+import PropTypes from 'prop-types';
```
pmsorhaindo: I just saw your return the example in the index.js of the example folder in the `Login` example folder above.. Shouldn't this be the case for all components where only one example file is needed or are you trying to keep things easily extendable for multi examples still? I think a single example is highly prefered we should return the `ExampleConfig` wrapped component from the index.

HHogg: Yeah I agree. I'll update so they are the same as the Login example. It will hopefully encourage single examples where possible.

PR #248, opened 2017-06-19
Bullet Chart

lpoulter:
Adds a new Bullet Chart Component.

Demo:
http://president-randy-42828.netlify.com/docs/charts/bullet-chart

Due to fairly specific requirements i've added a new component rather than adapt the existing ColumnChart.

- Takes two data series. 
- Normalises data.

screen shot 2017-06-23 at 12 08 48

Showing normalised data
screen shot 2017-06-22 at 13 22 20

### DIFF
--- a/src/charts/bars/Bar.js
+++ b/src/charts/bars/Bar.js
@@ -21,6 +21,8 @@ export default class Bar extends Component {
       'brown',
       'grey',
     ]).isRequired,
+    /** Overwriting label above the bar instead of default percentage */
+    label: PropTypes.string,
     /** Control for applying strong styling to the label */
     labelStrong: PropTypes.bool,
     /** Minimum thickness of the Bar */
@@ -49,7 +51,16 @@ export default class Bar extends Component {
 
   render() {
     const { direction } = this.context;
-    const { color, labelStrong, onClick, minSize, percent, showLabel, size, ...rest } = this.props;
+    const {
+      color,
+      label,
+      labelStrong,
+      onClick,
+      minSize,
+      percent,
+      showLabel,
+      size,
+      ...rest } = this.props;
     const isVertical = direction === 'up' || direction === 'down';
     const classes = classnames('ax-bars__bar', {
       'ax-bars__bar--center': size,
@@ -82,7 +93,7 @@ export default class Bar extends Component {
           style={ style }>
         <div className={ rectClasses } style={ rectStyle } />
         <div className={ labelClasses }>
-          <Small textStrong={ labelStrong }>{ percent }%</Small>
+          <Small textStrong={ labelStrong }>{ label || `${percent}%` }</Small>
         </div>
       </Base>
     );
--- a/src/charts/bullet-chart/BulletBars.js
+++ b/src/charts/bullet-chart/BulletBars.js
@@ -1,0 +1,57 @@
+import PropTypes from 'prop-types';
+import React, { Component } from 'react';
+import { Base, Bar, Bars } from 'bw-axiom';
+import { findBarGroupMax, arrayOfLength } from './utils';
+
+export default class BulletBars extends Component {
+  static propTypes = {
+    barLabel: PropTypes.number,
+    dataSetMax: PropTypes.number.isRequired,
+    label: PropTypes.string,
+    showBarLabel: PropTypes.bool,
+    values: arrayOfLength.bind(null, 2),
+  };
+
+  render() {
+    const {
+      dataSetMax,
+      label,
+      barLabel,
+      showBarLabel,
+      values,
+      ...rest } = this.props;
+
+    const barGroupMax = values.reduce(findBarGroupMax);
+
+    const maxDataPoint = values[barGroupMax.index];
+    const minDataPoint = values[barGroupMax.index === 1 ? 0 : 1];
+
+    const barGroupHeight = (maxDataPoint.value / dataSetMax) * 100;
+
+    const style = {
+      height: `${barGroupHeight}%`,
+    };
+
+    maxDataPoint.value = 100;
+    minDataPoint.value = ((minDataPoint.value / barGroupMax.value) * 100);
+
+    return (
+      <Base { ...rest } className="ax-bullet-chart__block">
+        <div className="ax-bullet-chart__bars" style={ style }>
+          { values.map(({ color, value }) =>
+            <div className="ax-bullet-chart__bars-bar" key={ `${color}-div` }>
+              <Bars direction="up" label={ label }>
+                <Bar
+                    color={ color }
+                    key={ color }
+                    label={ `${barLabel}%` }
+                    percent={ value }
+                    showLabel={ showBarLabel }/>
+              </Bars>
+            </div>
+                )}
+        </div>
+      </Base>
+    );
+  }
+}
--- a/src/charts/bullet-chart/BulletChart.js
+++ b/src/charts/bullet-chart/BulletChart.js
@@ -1,0 +1,89 @@
+import PropTypes from 'prop-types';
+import React, { Component } from 'react';
+import { Base } from 'bw-axiom';
+import BulletBars from './BulletBars';
+import ColumnChartBars from '../column-chart/ColumnChartBars';
+import ColumnChartVisual from '../column-chart/ColumnChartVisual';
+import ColumnChartRow from '../column-chart/ColumnChartRow';
+import ColumnChartXAxis from '../column-chart/ColumnChartXAxis';
+import ColumnChartXAxisLabel from '../column-chart/ColumnChartXAxisLabel';
+import './BulletChart.css';
+import { formatData, getHighestValue } from './utils';
+
+export default class BulletChart extends Component {
+  static propTypes = {
+    /** A key for the chart */
+    chartKey: PropTypes.arrayOf(PropTypes.shape({
+      color: PropTypes.string.isRequired,
+      label: PropTypes.string.isRequired,
+    })).isRequired,
+    /** The data to build the chart */
+    data: PropTypes.arrayOf(PropTypes.shape({
+      label: PropTypes.node.isRequired,
+      values: PropTypes.object.isRequired,
+    })).isRequired,
+    /** Controls the height of the containing element */
+    height: PropTypes.string.isRequired,
+    /** The offset value of the X-axis labels */
+    labelColumnWidth: PropTypes.string,
+    /** Controls which data set's bar label shows */
+    labelIndex: PropTypes.number,
+    /** Visibility control of the label above the each bar */
+    showBarLabel: PropTypes.bool,
+    /** Visibility control of the smaller label above the x-axis label*/
+    showSubLabel: PropTypes.bool,
+    /** Visibility control of the X-Axis label */
+    showXAxisLabels: PropTypes.bool,
+  };
+
+  static defaultProps = {
+    labelColumnWidth: '0rem',
+    labelIndex: 0,
+  };
+
+  render() {
+    const {
+      chartKey,
+      data,
+      height,
+      labelColumnWidth,
+      labelIndex,
+      showBarLabel,
+      showSubLabel,
+      showXAxisLabels,
+      ...rest
+    } = this.props;
+
+    const formattedData = formatData(chartKey, data);
+    const dataSetMax = getHighestValue(data);
+
+    return (
+      <Base { ...rest }
+          className="ax-bullet-chart"
+          style={ { height } }>
+        <ColumnChartRow>
+          <ColumnChartVisual>
+            { formattedData.map(({ values, label, subLabel }, index) =>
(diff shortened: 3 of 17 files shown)


src/charts/bullet-chart/BulletBar.css
```diff
@@ -0,0 +1,11 @@
+.ax-bullet-bar {
```
HHogg: So to keep the classes all under the same block all the classes should start with `ax-bullet-chart__`... so this would become `ax-bullet-chart__bar`

src/charts/bullet-chart/BulletBar.css
```diff
@@ -0,0 +1,11 @@
...
+  position: absolute;
+  bottom: 0;
+
+  &:first-child {
+    width: 100%;
+  }
+  &:nth-child(2)  {
+    width: 75%;
```
HHogg: To follow other pattern this should be a fixed padding amount set from the grid size.

src/charts/bullet-chart/BulletBar.js
```diff
@@ -0,0 +1,77 @@
+import PropTypes from 'prop-types';
+import React, { Component } from 'react';
+import { Base } from 'bw-axiom';
+import classnames from 'classnames';
+import './BulletBar.css';
```
HHogg: Following recent patterns all the CSS for a module can live in one file because they're all so connected.

src/charts/bullet-chart/BulletBar.js
```diff
@@ -0,0 +1,77 @@
...
+    /**
+     * Percentage value of the Bar, which determines the height and
+     * content for the label
+     */
+    percent: PropTypes.number.isRequired,
+    /** Visibility control of the label */
+    showLabel: PropTypes.bool,
+    /** Explicit thickness of the Bar */
+    size: PropTypes.string,
+    /** onClick handler that activates clickable styling */
```
HHogg: You only need to add these docs to the exposed components.

src/charts/bullet-chart/BulletBar.js
```diff
@@ -0,0 +1,77 @@
...
+    const {
+      color,
+      onClick,
+      minSize,
+      percent,
+      size,
+      ...rest } = this.props;
+    const classes = classnames('ax-bars__bar', 'ax-bullet-bar', {
+      'ax-bars__bar--center': size,
+      'ax-bars__bar--clickable': onClick,
```
HHogg: The bars classes should not be reused. It would be better to use the Bars components and use a wrapping container to handle the custom positioning requirements of the bullet chart.

src/charts/bullet-chart/BulletBar.js
```diff
@@ -0,0 +1,77 @@
...
+    const rectClasses = classnames('ax-bars__bar-rect',
+      `ax-bars__bar-rect--${color}`);
+
+    const style = {
+      height: `${percent}%`,
+    };
+
+    const rectStyle = {
+      width: size,
+      minWidth: !size && minSize,
```
HHogg: A lot of this duplication can be avoided by reusing the Bars components.

src/charts/bullet-chart/BulletBars.css
```diff
@@ -0,0 +1,32 @@
...
+  align-items: center;
+  width: 100%;
+  height: 100%;
+}
+
+.ax-bullet-bars__bar-label {
+  flex: 0 0 auto;
+  align-self: center;
+  margin-top: var(--cmp-chart-label-spacing);
+  padding-bottom: 0.5rem;
```
HHogg: Magic number ✨

src/charts/bullet-chart/BulletChart.js
```diff
@@ -0,0 +1,83 @@
...
+      label: PropTypes.node.isRequired,
+      values: PropTypes.object.isRequired,
+    })).isRequired,
+    height: PropTypes.string.isRequired,
+    label: PropTypes.node,
+    labelColumnWidth: PropTypes.string,
+    labelIndex: PropTypes.number,
+    showBarLabel: PropTypes.bool,
+    showSubLabel: PropTypes.bool,
+    showXAxisLabels: PropTypes.bool,
```
HHogg: This is the component that should have docs for the propTypes.

src/charts/bullet-chart/example/bullet.js
```diff
@@ -0,0 +1,46 @@
...
+  BulletChart,
+} from 'bw-axiom';
+import { data, dataFull, dataSmall, chartKey } from './data';
+
+export default class BulletExample extends Component {
+  render() {
+    return (
+      <Base>
+        <Example name="Bullet Chart">
+          <Snippet>
```
HHogg: As it's a new component it should use the new `<ExampleConfig>` component. You can see an example of it's use inside DotPlot example.

style-guide/routes.js
```diff
@@ -211,6 +211,11 @@ export default (
             examples={ require('../src/charts/bar-chart/example') }
             component={ Doc } />
         <Route
+            path="bullet-chart"
```
HHogg: This also needs adding to webpack.static.config.js file.

src/charts/bars/Bar.js
```diff
@@ -49,11 +50,21 @@ export default class Bar extends Component {
...
+      minSize,
+      percent,
+      showLabel,
+      size,
+      ...rest } = this.props;
     const isVertical = direction === 'up' || direction === 'down';
     const classes = classnames('ax-bars__bar', {
       'ax-bars__bar--center': size,
       'ax-bars__bar--clickable': onClick,
+      'ax-bullet-chart__bar': bulletBar,
```
HHogg: Sorry, this isn't really what I meant about reusing the Bar component... The class now doesn't represent the CSS inside the Bars module, they don't have any relation. The CSS you're applying with the class doesn't need to go directly on Bar component and can go on a element with the Bar component as a child, this will help keep separated the positioning requirements of BulletChart away from the generic directional Bar component.

src/charts/bullet-chart/BulletBars.js
```diff
@@ -0,0 +1,93 @@
...
+      labelStrong,
+      barLabel,
+      values,
+      ...rest } = this.props;
+
+    const labelClasses = classnames('ax-bullet-chart__bar-label', {
+      'ax-bars__bar-label--hidden': !showLabel,
+    });
+
+    const classes = classnames('ax-bars', 'ax-bars--up');
```
HHogg: These classes belong to the Bars component and should not be being used in here.

src/charts/bullet-chart/BulletBars.js
```diff
@@ -0,0 +1,93 @@
...
+    maxDataPoint.value = 100;
+    minDataPoint.value = ((minDataPoint.value / barGroupMax.value) * 100);
+
+    return (
+      <Base { ...rest } className={ classes }>
+        <div className="ax-bullet-chart__bars-container">
+          <div className="ax-bullet-chart__bars">
+            <div className="ax-bullet-chart__bars-wrapper" style={ style }>
+              { normalisedValues.map(({ color, value }) =>
+                <Bar
```
HHogg: <Bar> should always be wrapped in a <Bars> Component as it sets the direction styling.

src/charts/bullet-chart/BulletBars.js
```diff
@@ -0,0 +1,93 @@
...
+
+    const labelClasses = classnames('ax-bullet-chart__bar-label', {
+      'ax-bars__bar-label--hidden': !showLabel,
+    });
+
+    const classes = classnames('ax-bars', 'ax-bars--up');
+
+    const barGroupMax = values.reduce(findBarGroupMax);
+
+    const normalisedValues = [
```
HHogg: Is this needed? It doesn't look like you ever modify the values of any of the copied children children.

src/charts/bullet-chart/BulletChart.js
```diff
@@ -0,0 +1,86 @@
...
+import ColumnChartVisual from '../column-chart/ColumnChartVisual';
+import ColumnChartRow from '../column-chart/ColumnChartRow';
+import ColumnChartXAxis from '../column-chart/ColumnChartXAxis';
+import ColumnChartXAxisLabel from '../column-chart/ColumnChartXAxisLabel';
+import './BulletChart.css';
+import { formatData, getHighestValue } from './utils';
+
+export default class BulletChart extends Component {
+  static propTypes = {
+    chartKey: PropTypes.arrayOf(PropTypes.shape({
```
HHogg: Need a description for here.

src/charts/bullet-chart/BulletChart.js
```diff
@@ -0,0 +1,86 @@
...
+      color: PropTypes.string.isRequired,
+      label: PropTypes.string.isRequired,
+    })).isRequired,
+    data: PropTypes.arrayOf(PropTypes.shape({
```
HHogg: Need a description for here.

src/charts/bullet-chart/BulletChart.js
```diff
@@ -0,0 +1,86 @@
...
+      label: PropTypes.node.isRequired,
+      values: PropTypes.object.isRequired,
+    })).isRequired,
+    /** Controls the height of the containing element */
+    height: PropTypes.string.isRequired,
+    labelColumnWidth: PropTypes.string,
```
HHogg: Needs a description here.

src/charts/bullet-chart/BulletBars.js
lpoulter: The problem here is that i need to set the same style on all the parents of the bars.

```
display: flex;
    position: relative;
    flex-direction: column;
    align-items: center;
    width: 100%;
```

src/charts/bullet-chart/BulletBars.js
```diff
@@ -0,0 +1,67 @@
+import PropTypes from 'prop-types';
+import React, { Component } from 'react';
+import classnames from 'classnames';
+import Bar from '../bars/Bar';
+import Bars from '../bars/Bars';
+import { Base } from 'bw-axiom';
```
HHogg: Really minor, Bars and Bar are exported in Axiom so could include them here.

src/charts/bullet-chart/BulletBars.js
```diff
@@ -0,0 +1,67 @@
...
+  render() {
+    const {
+      dataSetMax,
+      label,
+      barLabel,
+      showBarLabel,
+      values,
+      ...rest } = this.props;
+
+    const classes = classnames('ax-bullet-bars');
```
HHogg: As there's no conditional classes, could inline this and then ditch classnames import :)

src/charts/bullet-chart/BulletBars.js
```diff
@@ -0,0 +1,67 @@
...
+
+    const barGroupMax = values.reduce(findBarGroupMax);
+
+    const normalisedValues = [
+      Object.assign({}, values[0]),
+      Object.assign({}, values[1]),
```
HHogg: These could be removed right? Probably should be any case of mutating the original given data ;)

src/charts/bullet-chart/BulletBars.js
```diff
@@ -0,0 +1,67 @@
...
+    const style = {
+      height: `${barGroupHeight}%`,
+    };
+
+    maxDataPoint.value = 100;
+    minDataPoint.value = ((minDataPoint.value / barGroupMax.value) * 100);
+
+    return (
+      <Base { ...rest } className={ classes }>
+        <div className="ax-bullet-chart__bars" style={ style }>
```
HHogg: 1. `.ax-bullet-bars` should be named to the `ax-bullet-chat__` block 
2. Looking at the CSS, it looks like these two elements could be combined?

src/charts/bullet-chart/BulletBars.js
```diff
@@ -0,0 +1,62 @@
...
+    const style = {
+      height: `${barGroupHeight}%`,
+    };
+
+    maxDataPoint.value = 100;
+    minDataPoint.value = ((minDataPoint.value / barGroupMax.value) * 100);
+
+    return (
+      <Base { ...rest } className="ax-bullet-bars">
+        <div className="ax-bullet-chart__bars" style={ style }>
```
HHogg: 1. .ax-bullet-bars should be named to the ax-bullet-chat__ block
2. Looking at the CSS, it looks like these two elements could be combined?

src/charts/bullet-chart/BulletBars.js
```diff
@@ -0,0 +1,62 @@
...
+          { normalisedValues.map(({ color, value }) =>
+            <div className="ax-bullet-chart__bars-bar" key={ `${color}-div` }>
+              <Bars direction="up" key={ `${color}-bars` } label={ label }>
```
HHogg: This doesn't need a `key` on it.

PR #251, opened 2017-06-21
feat(BarChart): Added benchmarking styling

HHogg:
screen shot 2017-06-21 at 17 46 11


src/charts/bar-chart/BarChartBenchmarkLine.js
```diff
@@ -0,0 +1,26 @@
...
+
+export default class BarChartBenchmarkLine extends Component {
+  static propTypes = {
+    height: PropTypes.string,
+    value: PropTypes.number,
+    width: PropTypes.string,
+  };
+
+  static defaultProps = {
+    value: 50,
```
thatguynamedandy: seems like a strange default, should it have a default at all?

HHogg: Similar to the DotPlotBenchmarkLine and DotPlotDifferenceLine they are used inside the chart and also inside the chat keys. The defaults are there to position them correctly inside the key.

thatguynamedandy: ah I see, makes sense, go for it 👍

PR #257, opened 2017-06-23
Fix date fns

benji6:
The date functions were printing hours instead of minutes 🙀 ! This is fixed now and there are rudimentary unit tests too :D Also I have replaced moment with date-fns which significantly reduces build size 🎉


HHogg:
Amazing 😄

PR #260, opened 2017-06-26
fix(bullet-chart): fix bar height for small values

lpoulter:
Fixes two issues with the bullet chart component.

1. Labels for both bars were showing this is now corrected to only show the label for the highest value in each group.
2. For smaller values the containing element was not large enough to hold the bars and labels. This has been fixed by setting a normalised height on each bar rather than the container. 

screen shot 2017-06-26 at 16 18 34

### DIFF
--- a/src/charts/bullet-chart/BulletBars.test.js
+++ b/src/charts/bullet-chart/BulletBars.test.js
@@ -5,7 +5,6 @@ import { formattedData } from './BulletBars.test.fixture';
 
 function getComponent(props = {}) {
   props.values = formattedData;
-  props.dataSetMax = props.dataSetMax || 101;
 
   return renderer.create(
     <BulletBars { ...props }>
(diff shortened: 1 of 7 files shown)


src/charts/bullet-chart/BulletBars.test.js
```diff
@@ -20,13 +19,13 @@ describe('BulletBars', () => {
...
-      });
-    });
+  describe('renders with the correct label', () => {
+    const props = {
+      barLabel: 'label for highest value',
+      showBarLabel: true,
+    };
+    const component = getComponent(props);
+    const tree = component.toJSON();
+    expect(tree).toMatchSnapshot();
```
HHogg: Sorry, should have picked this up in the original PR... but this needs to be inside an `it()` block for it to run in the tests.

jdelight:
Great work on this @lpoulter and thanks for your help @HHogg 🎉

PR #264, opened 2017-06-28
fix(MirroredBulletChart): Add control for which bar label to show.

lpoulter:
Quick fix to allow the labelIndex to be passed to the sub Bullet Chart components.

### DIFF
--- a/src/charts/bullet-chart/MirroredBulletChart.js
+++ b/src/charts/bullet-chart/MirroredBulletChart.js
@@ -8,22 +8,32 @@ import './MirroredBulletChart.css';
 
 export default class MirroredBulletChart extends Component {
   static propTypes = {
+    /** A key for the chart */
     chartKey: PropTypes.arrayOf(PropTypes.shape({
       color: PropTypes.string.isRequired,
       label: PropTypes.string.isRequired,
     })).isRequired,
+    /** The data to build the top half of the chart */
     data: PropTypes.arrayOf(PropTypes.shape({
       label: PropTypes.node.isRequired,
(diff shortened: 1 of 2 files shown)


src/charts/bullet-chart/MirroredBulletChart.js
```diff
@@ -8,22 +8,32 @@ import './MirroredBulletChart.css';
...
     chartKey: PropTypes.arrayOf(PropTypes.shape({
       color: PropTypes.string.isRequired,
       label: PropTypes.string.isRequired,
     })).isRequired,
+    /** The data to build the top half of the chart */
     data: PropTypes.arrayOf(PropTypes.shape({
       label: PropTypes.node.isRequired,
       values: PropTypes.object.isRequired,
     })).isRequired,
+    /** Controls the height of the containing element */
```
benji6: maybe just `Controls the height of the chart` ?

src/charts/bullet-chart/MirroredBulletChart.js
```diff
@@ -8,22 +8,32 @@ import './MirroredBulletChart.css';
...
     height: PropTypes.string.isRequired,
+    /** Chart Y axis label */
     label: PropTypes.node,
+    /** The offset value of the X-axis labels */
```
benji6: super nitpicky but above you write `Y axis` and here you write `X-axis`. I'm not sure if there is a precedent for this but I'd be tempted to just call them `x-axis` and `y-axis` (that's what this wiki section calls them at least https://en.wikipedia.org/wiki/Cartesian_coordinate_system#Two_dimensions)

PR #265, opened 2017-07-04
fix(DotPlot): Benchmark line is now correct height

benji6:
Was this:
screen shot 2017-07-04 at 14 35 37

Now this:
screen shot 2017-07-04 at 14 35 17

🎉


pmsorhaindo:
Twitter Average* ?

PR #270, opened 2017-07-06
Avatar and Candytar moved over to configurable examples. 

HHogg:
While moving the examples over, I also noticed an issue with the positioning of candytars and the image fallback component when the src changes.


src/components/candytar/Candytar.js
```diff
@@ -39,7 +40,14 @@ export default class Candytar extends Component {
       'brown',
       'grey',
     ]),
+    /**
+     * A function that can be passed to determine the colour of the candytar.
+     * By default if no colour is provided it will randomly picker a colour
+     * from the available options. Given function is called with the available
```
benji6: It's probably obvious, but just to be really specific maybe say "Given function is called with an array of the available colors"

Also mind the English "u"s sneaking into "color" here :D

PR #275, opened 2017-07-06
docs(Base): Moved over to new configurable examples

HHogg:


src/components/base/Base.js
```diff
@@ -5,19 +5,34 @@ import './Base.css';
...
+     * react registered element tag.
+     */
     Component: PropTypes.oneOfType([
       PropTypes.string,
       PropTypes.func,
     ]),
+    /** Class name to be appended to the element */
     className: PropTypes.string,
+    /**
+     * Control over when the element should be hidden until, passed that value
```
benji6: "passed" should be "past" here and for `visibleUntil`. I think it's a little confusing to read this. Maybe it's simpler just to drop that whole "past that value" clause and go with: "Control over when the element should be hidden until. Opposite of `visibleUntil`" or something like that?

src/components/base/Base.js
```diff
@@ -5,19 +5,34 @@ import './Base.css';
...
+    /** Text breaking styling */
     textBreak: PropTypes.oneOf(['none', 'all', 'word']),
+    /** Text casing styling */
     textCase: PropTypes.oneOf(['upper', 'capital', 'lower']),
-    textCenter: PropTypes.oneOfType([
-      PropTypes.bool,
-      PropTypes.oneOf(['small', 'medium', 'large']),
-    ]),
+    /**
+     * Text central alignment either all of the time, * with a value of `true`
```
benji6: I think the `*` in the middle of this line isn't supposed to be here? The comma before the asterisk can probably go too

src/components/base/Base.js
```diff
@@ -5,19 +5,34 @@ import './Base.css';
...
+     * otherwise at one of the breakpoints specified.
+     */
+    textCenter: PropTypes.oneOf([true, 'small', 'medium', 'large']),
+    /** Text colour styling */
```
benji6: American English spelling

HHogg:
I'm going to close this PR down as a lot of it is superseded with changes coming in for the Base component, but I'll make these changes inside that PR 👍

PR #277, opened 2017-07-11
Typography and Base design alignment 

HHogg:
Overall point of this is to align typography and the base component more with the design system... and have tagged on a change to the Toolbar. Please see commits for granular description.


src/components/base/Base.js
```diff
@@ -3,21 +3,43 @@ import React, { Component } from 'react';
...
+  display2: true,
+  headline: true,
+};
+
 export default class Base extends Component {
   static propTypes = {
+    /**
+     * The component that should be rendered with the Base component
+     * functionality. If a string is given then it must be a valid
+     * react registered element tag.
```
benji6: maybe capitalize "React"?

src/components/base/Base.js
```diff
@@ -3,21 +3,43 @@ import React, { Component } from 'react';
 import classnames from 'classnames';
 import './Base.css';
 
+const UNDERLINE_MAP = {
```
benji6: You could use a `Set` here - they're supported (most functionality) in IE11!

```js
const underlineTextSizes = new Set(['body', 'display1', 'display2', 'headline'])
```
then instead of `UNDERLINE_MAP[textSize || 'body']` you can go `underlineTextSizes.has(textSize || 'body')`

I think it makes more sense to use a `Set` here, but makes no difference to functionality so up to you

PR #279, opened 2017-07-13
feat(BarChart): Hover behavior now more consistent with DotPlotChart giving more emphasis to the hovered data

benji6:
https://bar-chart-hover-improvements--axiom-rules.netlify.com/docs/charts/bar-chart

screen shot 2017-07-13 at 14 51 39

### DIFF
--- a/src/charts/bar-chart/BarChart.css
+++ b/src/charts/bar-chart/BarChart.css
@@ -1,3 +1,4 @@
+@import '../../materials/animations';
 @import '../../materials/colors';
 @import '../../materials/sizing';
 @import '../vars';
@@ -17,12 +18,17 @@
   width: 100%;
   height: 100%;
   padding-right: var(--cmp-chart-overflow-space);
+  transition: opacity var(--transition-time-base) var(--transition-function);
   pointer-events: none;
 }
 
 .ax-bar-chart__benchmark-line {
   position: relative;
   height: 100%;
+}
+
+.ax-bar-chart__benchmark-line--faded {
+  opacity: var(--cmp-chart-opacity-faded);
 }
 
 .ax-bar-chart__benchmark-line-path {
--- a/src/charts/bar-chart/BarChartBars.js
+++ b/src/charts/bar-chart/BarChartBars.js
@@ -9,9 +9,11 @@ export default class BarChartBars extends Component {
     ContextComponent: PropTypes.func,
     benchmark: PropTypes.number,
     data: PropTypes.object.isRequired,
+    fadeBenchmarkLine: PropTypes.bool.isRequired,
+    hideBars: PropTypes.bool.isRequired,
     hoverColor: PropTypes.string,
+    isHovered: PropTypes.bool.isRequired,
     label: PropTypes.node.isRequired,
-    labelStrong: PropTypes.bool.isRequired,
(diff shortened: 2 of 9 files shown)


src/charts/bar-chart/BarChart.css
```diff
@@ -17,9 +17,14 @@
   width: 100%;
   height: 100%;
   padding-right: var(--cmp-chart-overflow-space);
+  transition: opacity var(--transition-time-base) var(--transition-function);
   pointer-events: none;
 }
 
+.ax-bar-chart__benchmark-line-container--faded {
+  opacity: 0.25;
```
HHogg: As this value is [also used in the DotPlot](https://github.com/BrandwatchLtd/axiom/blob/master/src/charts/dot-plot/DotPlot.css#L101), as is the go-to-faded-chart-opacity it might be nice to have this in the [charts/vars.css](https://github.com/BrandwatchLtd/axiom/blob/master/src/charts/vars.css) for a central point of tweaking in the future.

src/charts/bar-chart/BarChartBars.js
```diff
@@ -24,37 +27,49 @@ export default class BarChartBars extends Component {
...
-      labelStrong,
       showBarLabel,
       size,
       values,
       onMouseEnter,
       onMouseLeave,
     } = this.props;
 
+    const benchmarkLineContainerClasses = classnames('ax-bar-chart__benchmark-line-container', {
+      'ax-bar-chart__benchmark-line-container--faded': fadeBenchmarkLine,
```
HHogg: Is there any reason for doing this here rather than inside the `<BarChartBenchmarkLine>` component similar to [how `<DotPlotBenchmarkLine>` component does it](https://github.com/BrandwatchLtd/axiom/blob/master/src/charts/dot-plot/DotPlotBenchmarkLine.js#L23)?

src/charts/bar-chart/BarChartContext.js
```diff
@@ -21,11 +23,18 @@ export default class BarChartContext extends PureComponent {
     onMouseLeave: PropTypes.func,
   };
 
+  staticDefaultProps = {
```
HHogg: So `staticDefaultProps` isn't valid because it doesn't have the space... and also you don't need to defined default props for falsey values really. It's sometimes nice to have these default `false` values for the API docs, but it's not an exposed component so doesn't provide any benefit here.

src/charts/bars/Bar.css
```diff
@@ -5,12 +5,21 @@
...
   display: flex;
+  transition: opacity var(--transition-time-base) var(--transition-function);
 }
 
 .ax-bars__bar--center {
   align-items: center;
 }
 
+.ax-bars__bar--faded {
+  opacity: 0.25;
```
HHogg: Same comment about these all linking to the CSS variable.

src/charts/bar-chart/BarChartBars.js
benji6: no reason, I'll change it for consistency

src/charts/bar-chart/BarChartContext.js
benji6: Very true! Will delete this

PR #280, opened 2017-07-13
docs(Button): Moved over to configurable examples

HHogg:
Also added the ability to have varying child examples. 

screen shot 2017-07-13 at 21 51 50


src/components/button/ButtonIcon.js
```diff
@@ -6,8 +6,17 @@ import './ButtonIcon.css';
 
 export default class ButtonIcon extends Component {
   static propTypes = {
+    /**
+     * Whether the ButtonIcon is at the end of a button. If true,
+     * it will apply spacing from the text.
+     */
     isEnd: PropTypes.bool,
```
thatguynamedandy: Might be worth seeing if there is a way to hide certain props from the UI, there probably is a use case for hidden properties. In this scenario the child ButtonIcon has a prop which is set directly by its parent, so changing it in the UI doesn't do anything. Not an issue with the component library more the Axiom test bed, so happy to defer to a separate bug.

PR #281, opened 2017-07-14
docs(Setup): Added setup docs for postcss and create-react-app

HHogg:


docs/create-react-app.md
```diff
@@ -0,0 +1,92 @@
...
+Ready to go!
+
+```
+yarn start
+```
+
+
+#### Set up production build
+
+Inside `config/webpack.config.prod.js` replace the css rile for
```
jdelight: rile -> file minor thing

HHogg: This typo riles me up 😡

HHogg: I think I meant to write "rule"

docs/create-react-app.md
```diff
@@ -0,0 +1,92 @@
...
+yarn eject
+yarn add bw-axiom postcss-cssnext postcss-import
+```
+
+#### Set up postcss config
+
+Create `config/postcss.config.js`, doesn't have to be in config directory but that's where create-react-app keeps all of the config files.
+
+```js
+{
```
roburidge: Should we keep the `ident: 'postcss'` bugfix and `postcss-flexbugs-fixes` plugin here that CRA includes?

docs/create-react-app.md
```diff
@@ -0,0 +1,92 @@
...
+    require('postcss-import')(),
+    require('postcss-cssnext')({
+      browsers: [
+        'last 2 versions',
+        'not Explorer <= 10',
+        'not ExplorerMobile <= 10',
+      ],
+      features: {
+        customProperties: {
+          variables: Object.assign({},
```
roburidge: What's the reason for using Object.assign here?

docs/create-react-app.md
```diff
@@ -0,0 +1,92 @@
...
+  ],
+}
+```
+
+Inside `config/webpack.config.dev.js` replace the css rule for
+
+```js
+{
+  test: /\.css$/,
+  use: ['style-loader', 'css-loader', {
```
roburidge: Should we use the css-loader option `importLoaders: 1` here when using postcss-loader?
https://github.com/webpack-contrib/css-loader#importloaders

docs/create-react-app.md
HHogg: Very soon (like next week) Axiom is going to be dropping IE11 support (already it doesn't support IE10) out of the box, because of requirements on a project that it needs scoped CSS custom properties. 

Though there will be a doc about how to retrofit it for IE, with setting up preprocessing custom properties. So this is something we should definitely add to that 👍 ... but it wont be needed for the main setup because all of the bugs that plugin fixes are IE10/IE11.

docs/create-react-app.md
HHogg: It's not needed, I copied it from another project that adds additional variables. I'll remove, good shout! 👍

docs/create-react-app.md
HHogg: Yes we should! I did it for the production config one but not this, good shout! 👍

docs/create-react-app.md
roburidge: Edge only, awesome! 🎉 no need for the felxbug fixes then!

I think `indent: 'postcss'` might still be needed, although I don't fully understand the issue it's solving (something to do with css imports and the scripts included with CRA).
https://github.com/facebookincubator/create-react-app/pull/2430
https://github.com/facebookincubator/create-react-app/issues/2677#issuecomment-311924650

HHogg: As there's the possibility of CRAs config changing over time, I think I'll reword those sections to imply merging in the cssnext and imports configs rather than replace. That should hopefully account for this and also future custom stuff for CRA.

jdelight:
🎉 Thanks for this @HHogg (& @roburidge)

PR #282, opened 2017-07-14
fix(Charts): Zoom behaviour now includes benchmarking

HHogg:


src/charts/bar-chart/utils.js
```diff
@@ -14,6 +14,10 @@ export function getHighestValue(data) {
   let max = 0;
```
thatguynamedandy: I couldn't resist seeing what this looks like as a one liner, but it's a lot slower (about 88% according to JSPerf, @benji6 would be happy)

```
const max = Math.max(...data.map((i) => Math.max(...Object.values(i.values).concat(i.benchmark || 0))))
```

HHogg: I think that was the argument for @benji6 writing it like this to begin with 😄

PR #285, opened 2017-07-17
DockItem tweaks to allow custom content

antiBaconMachine:
* Allows `DockItems` to accept children
* Removes the anchor wrapper from `DockItem`s as these aren't nestable
* Adds a 'bottom' attribute to `DockItem` which aligns them to the bottom of the `Dock`

It works but the markup for `DockItem` is now in an ugly ternary to avoid making a breaking change. I wonder if it may be prudent to do something a bit more drastic and make a breaking change to `DockItem` to strip the icon functionality all together and put this in a new `DockIcon` component?

What do you think is best @HHogg?


src/components/platform/DockItem.css
```diff
@@ -27,7 +27,11 @@
   color: var(--color-ui-light--light);
 }
 
-.ax-platform__dock-item-link {
+.ax-platform__dock-item--bottom {
+  margin-top: auto;
```
HHogg: Using the `auto` margin trick probably won't allow for multiple items to be aligned to the bottom?

antiBaconMachine: Ah yeah, I guess it would start to evenly divide the space. That's rubbish.

antiBaconMachine:
Yeah I think having a header / footer is a good idea, bit clearer semantically. Ok I'll take a run at that, see what it looks like 👍

antiBaconMachine:
Have updated PR. Essentially this is what @HHogg suggested with two obvious deviations:

* `s/DockBody/DockHeader/` - Just preferred having a header/footer combo
* `DockIconLink` embeds it's own `DockItem`. I'm pretty sure this covers 100% of use cases today and not sure I see the utility in having a `DockIconLink` not in a `DockItem`. If we did want that it'd be at least a  semantic refactor anyway.

src/components/platform/DockItem.js
```diff
@@ -1,26 +1,22 @@
...
-import classnames from 'classnames';
-import { Icon } from 'bw-axiom';
 import './DockItem.css';
 
 export default class DockItem extends Component {
   static propTypes = {
     active: PropTypes.bool,
-    icon: PropTypes.string.isRequired,
+    children: PropTypes.node,
+    interactive: PropTypes.bool,
```
HHogg: Looks like it might have been left in by mistake as it's not used?

src/components/platform/DockFooter.js
```diff
@@ -0,0 +1,19 @@
+import PropTypes from 'prop-types';
+import React, { Component } from 'react';
+import './Dock.css';
```
HHogg: This should only needs importing once inside Dock.js

src/components/platform/DockFooter.js
```diff
@@ -0,0 +1,19 @@
...
+export default class DockFooter extends Component {
+  static propTypes = {
+    children: PropTypes.node,
+  };
+
+  render() {
+    const { children, ...rest } = this.props;
+
+    return (
+      <ul { ...rest } className="ax-platform__dock-list ax-platform__dock-list--footer">
```
HHogg: For predictable future maintenance, we don't use the same classname twice across components. You could avoid this by using both the classes in the CSS like 

```css
.ax-platform__dock-list-header,
.ax-platform__dock-list-footer {
  margin: 0;
  padding: 0;
}
```

src/components/platform/Dock.css
```diff
@@ -14,3 +14,12 @@
   list-style: none;
   z-index: var(--z-index-platform);
 }
+
+.ax-platform__dock-list {
+  margin: 0;
+  padding: 0;
+}
+
+.ax-platform__dock-list--footer {
```
HHogg: This is slightly incorrect BEM syntax `--` is used as a modifier like ... 

```css
.button--active
.button--red
```

`__` is used for elements like ... 

```css
.button__group
.button__icon
```

And we try not to go toooo crazy with BEM nesting like 

```css
.button__group__icon__svg__path--red
```

So  these classes could just be 

```css
.ax-platform__dock-list-footer
.ax-platform__dock-list-header
```

src/components/platform/DockHeader.js
```diff
@@ -0,0 +1,19 @@
+import PropTypes from 'prop-types';
+import React, { Component } from 'react';
+import './Dock.css';
```
HHogg: Same here with the import.

antiBaconMachine:
Cheers @HHogg, think I've addressed all those

HHogg:
As mentioned, but for the benefit of others. 

Commit messages are used for the release notes. The 2 commits here can be squashed with the message something like ...

```
feat(Platform): Added unclickable DockItems that can be aligned to the bottom of the Dock. 

BREAKING CHANGE: 
* Original `<DockItem>` has now become `<DockItemLink>`
* Original DockItems should now be wrapped within a `<DockHeader>` 
```

Where the `BREAKING CHANGE` is what will bump the major version, The rest will be added to the release notes like here https://github.com/BrandwatchLtd/axiom/releases/tag/v23.0.0.

PR #288, opened 2017-07-19
feat(Theming): Support combination of light and dark themes

HHogg:
BREAKING CHANGE:

Axiom has now changed to not preprocess css variables by default. This was needed to support nested theming and as a result IE11 is no longer supported out of the box, however using PostCSS and the provided CSS variables partial IE11 support can be added but will not have nested theme support. Please see the IE11 support documentation on how to set it up.

For non IE11 projects follow the following steps to migrate over.

1) Remove all css variables imports from your css files as these are no longer needed, for example...

```diff
- @import 'bw-axiom/lib/materials/animations';
- @import 'bw-axiom/lib/materials/colors';
- @import 'bw-axiom/lib/materials/layout';
- @import 'bw-axiom/lib/materials/sizing';
- @import 'bw-axiom/lib/materials/typography';
```

There is the exception if you are using the custom media queries, in which case you will need to include these into your css with...

```diff
+ @import 'bw-axiom/lib/materials/breakpoints.css';
```

2) Add the class `ax-theme--light` or `ax-theme--dark` to the *`<html>`* element.
3) `<TextInput>` when `theme` was used before, the prop should still remain on the component but the prop `style="overlay"` should be added


HHogg:
[Netlify Link](https://remove-processing-css-vars--hhogg-axiom.netlify.com/)

src/components/form/TextInput.js
```diff
@@ -12,7 +12,7 @@ export default class TextInput extends Component {
     invalid: PropTypes.bool,
     label: PropTypes.string,
     size: PropTypes.oneOf(['medium', 'large']),
-    theme: PropTypes.oneOf(['dark', 'light']),
+    style: PropTypes.oneOf(['overlay']),
```
benji6: I wonder if we should stay away from using `style` just because it has a different meaning in React components normally?

src/materials/breakpoints.css
```diff
@@ -0,0 +1,7 @@
+@custom-media --screen-less-than-small (max-width: 47.9375rem);
```
benji6: These seem oddly rounded. Is there a reason for that? I remember in my old company there was a philosophy of using only screen-more-than media queries - the idea being that you built for mobile first then you think about desktop. I quite liked that you only had one type at least so only had to ever think in one direction

src/materials/colors.css
```diff
@@ -1,22 +1,38 @@
 /* stylelint-disable color-no-hex */
 :root {
-  --color-ui-dark--dark: #343239;
-  --color-ui-dark: #434148;
-  --color-ui-dark--light: #48464D;
+  --rgb-ui-dark--dark: 48, 45, 53;
```
benji6: I did not realise you could do this!

src/components/form/TextInput.js
HHogg: Yeah I agree with this, I've seen on some projects that it also conflicts with the eslint rule [`style-prop-object`](https://github.com/yannickcr/eslint-plugin-react/blob/master/docs/rules/style-prop-object.md)... however it is also how design refer to these design variations... so I'm torn 😕. 

We're currently going through all of the components, props and values to align here with the sketch file. So maybe it's something we can do with that task?

src/materials/colors.css
HHogg: I would prefer to keep the `color()` function syntax though 😞

src/materials/breakpoints.css
HHogg: I'm not convinced with how Axiom is handling responsive, there hasn't been as much thought into it as I would like. Any improvements to it would be very much welcome! Maybe something we should sit down with design and talk about a better approach going forward?

PR #289, opened 2017-07-19
feat(Icon): Add a closed-folder icon

SeaOfTea:
folder-closed closed folder


HHogg:
The exported SVG is incorrect (and so is the bar chart one from #262) they are both using masks which will conflict with each, as shown in  the screen grab bellow from the style guide on this branch. 

screen shot 2017-07-19 at 16 14 11

There is a specific icon sketch file that icons should be exported from to prevent this from happening. This is my fault for there not being documentation around this. Adding now.

HHogg:
Awesome @SeaOfTea thanks for addressing so quickly :) one last minor gripe... release notes are auto-generated from the commit messages. The commit message refers to the icon name as `closed-folder`, but it's been added as `folder-closed`. I know it's minor, but would it be ok to update the commit message?

SeaOfTea:
Done.

PR #290, opened 2017-07-19
feat(IconIndicator): add primary and secondary IconIndicator styles	

pmsorhaindo:
https://foreground-icon-indicator-colors--librarian-chimpanzee-34483.netlify.com/

### DIFF
--- a/src/components/icon/IconIndicator.js
+++ b/src/components/icon/IconIndicator.js
@@ -9,18 +9,21 @@ export default class IconIndicator extends Component {
     color: PropTypes.oneOf(['success', 'warning', 'error', 'info', 'subtle']),
     name: PropTypes.string.isRequired,
     size: PropTypes.oneOf(['small', 'medium', 'large']),
+    style: PropTypes.oneOf(['primary', 'secondary']),
   };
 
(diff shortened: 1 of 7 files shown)


src/components/icon/IconIndicator.js
```diff
@@ -7,6 +7,7 @@ import './IconIndicator.css';
 export default class IconIndicator extends Component {
   static propTypes = {
     color: PropTypes.oneOf(['success', 'warning', 'error', 'info', 'subtle']),
+    foregroundColor: PropTypes.oneOf(['success', 'warning', 'error', 'info', 'subtle']),
```
HHogg: screen shot 2017-07-19 at 16 35 28

Looking at the design sketch file we should be following as close as possible the naming convention used by design. 

What do you think about keeping the existing `color` prop and using modifiers to switch between the styles like ... 
```js
static propTypes = {
  color: PropTypes.oneOf(['success', 'warning', 'error', 'info', 'subtle']),
  style: PropTypes.oneOf(['primary', 'secondary']),
};

static defaultPropTypes = {
  style: 'primary',
};
```js

PR #293, opened 2017-07-20
feat(Context): Multiselect menus with check boxes 

HHogg:
screen shot 2017-07-20 at 18 10 57

.... by adding `multiSelect` to `DropdownMenuItem/ContextMenuItem ` and the normal select will turn into a checkbox select. 

Example of it being used with the prop selection on color picker color array options. 

screen shot 2017-07-20 at 18 31 41


HHogg:
[Netlify Link](https://example-config-color-picker--hhogg-axiom.netlify.com/docs/components/context?state=%7B%22Context%22:%7B%22options%22:%7B%22children%22:%7B%22selection%22:%22ContextMenu%22%7D%7D%7D,%22ContextMenuItem%22:%7B%22props%22:%7B%22multiSelect%22:true,%22selected%22:true%7D%7D%7D)

style-guide/components/ExampleConfig/TypeArrayOfEnum.js
```diff
@@ -0,0 +1,70 @@
...
+  Context,
+  Dropdown,
+  DropdownContent,
+  DropdownMenu,
+  DropdownMenuItem,
+  DropdownTarget,
+} from 'bw-axiom';
+
+function isSelected(values, val) {
+  return Array.isArray(values) && !!values.find((v) => v == val);
```
benji6: I know it's just the styleguide but maybe we should use a `filter` `length` check here instead of `find` which isn't supported in IE11? I think the argument is that QA can't really check IE11 support if the styleguide itself doesn't support IE11

benji6: Although obviously there's a wider question over IE11 support atm I realise :D

style-guide/components/ExampleConfig/utils.js
```diff
@@ -7,12 +7,18 @@ const preparePropMap = {
...
       if (options.selection) {
-        return (options.options.find(({ name }) => name === options.selection) || {}).children
-          || propValue;
+        const selectedOptions = options.options.filter(({ name }) => name === options.selection)[0];
+        return selectedOptions ? selectedOptions.children : propValue;
       }
 
       if (options.count) {
-        return Array.from({ length: options.count }, () => propValue);
+        const children = [];
```
benji6: You could also have gone:

```js
Array.apply(null, {length: options.count}).map(() => propValue)
```

but that is some seriously black magic from the dark days of ES5! :D

HHogg: :D Nice

PR #297, opened 2017-07-28
Word cloud interactivity

thatguynamedandy:


thatguynamedandy:
https://word-cloud-interactivity--andyp-axiom.netlify.com/docs/charts/wordcloud

src/charts/wordcloud/Word.css
```diff
@@ -1,16 +1,21 @@
...
-  from  { opacity: 0; }
   to    { opacity: 1; }
 }
 
 .ax-word {
   opacity: 0;
   color: inherit;
+  fill-opacity: 1;
   fill: currentColor;
+  cursor: default;
```
HHogg: This is the initial property so shouldn't be needed here.

src/charts/wordcloud/Word.css
```diff
@@ -25,3 +30,53 @@
 .ax-word--orange { color: var(--color-brand-orange); }
 .ax-word--brown { color: var(--color-brand-brown); }
 .ax-word--grey { color: var(--color-brand-grey); }
+
+.ax-word--disabled {
+  fill-opacity: 0.3;
```
HHogg: Should the cursor be reset back to `default` when disabled? I would expect when a word is disabled, it shouldn't appear clickable with the pointer cursor, which it currently is when it has `onClick` given.

src/charts/wordcloud/Word.css
```diff
@@ -25,3 +30,53 @@
...
+  fill-opacity: 0;
+}
+
+.ax-word--clickable {
+  cursor: pointer;
+  &:active { color: var(--color-context-text--dark); }
+
+  &.ax-word--rose {
+    &:active { color: var(--color-brand-rose--dark); }
+  }
```
HHogg: As there is only one style added here (unlike the other situations where a `:hover/:disabled` style is given, there's not much needed for the extra nesting levels... 

```css
.ax-word--rose:active   { color: var(--color-brand-rose--dark); }
.ax-word--pink:active   { color: var(--color-brand-pink--dark); }
.ax-word--purple:active { color: var(--color-brand-pink--dark); }

/* ...etc */
```

... this is a little more readable.

src/charts/wordcloud/Word.css
```diff
@@ -25,3 +30,53 @@
 .ax-word--orange { color: var(--color-brand-orange); }
 .ax-word--brown { color: var(--color-brand-brown); }
 .ax-word--grey { color: var(--color-brand-grey); }
+
+.ax-word--disabled {
+  fill-opacity: 0.3;
+}
+
+.ax-word--hidden {
+  fill-opacity: 0;
```
HHogg: ❤️ This trickery to retain the opacity animation. Nice find 👍

src/charts/wordcloud/Word.js
```diff
@@ -22,10 +22,14 @@ export default class Word extends Component {
       'brown',
       'grey',
     ]),
+    /** Disabled state of the word **/
+    disabled: PropTypes.bool,
     /** Font family of the word */
     font: PropTypes.oneOf(['Roboto, Helvetica, Arial, sans-serif']).isRequired,
     /** Font size of the word */
     fontSize: PropTypes.string,
+    /** Hide the word **/
```
HHogg: NABD but it's a little inconsistent with the naming above... 

```
/** Disabled state of the word **/
/** Hidden state of the word **/
```

Maybe? ...

src/charts/wordcloud/Word.js
```diff
@@ -61,6 +70,9 @@ export default class Word extends Component {
 
     const classes = classnames('ax-word', {
       [`ax-word--${color}`]: color,
+      'ax-word--clickable': onClick,
```
HHogg: Relating to the `cursor` behaviour above, maybe?

```js
'ax-word--clickable': onClick && !disabled,
```

src/charts/wordcloud/Word.js
```diff
@@ -61,6 +70,9 @@ export default class Word extends Component {
...
+      'ax-word--disabled': disabled,
+      'ax-word--hidden': hidden,
```
HHogg: At the moment, the precedence of hidden and disabled styling is in the ordering of the CSS rules... I'm worried ordering of the CSS is something that can easily be changed and would lead to a styling bug when both of these properties are set.

What do you think about adding the logic here to make the precedence more clear and robust? so as to say ...

```js
'ax-word--disabled': !hidden && disabled,
'ax-word--hidden': hidden,
```

That way hidden styling will always take precedence.

HHogg: Also, this would benefit from some snapshot tests to test those precedences.

src/charts/wordcloud/Word.js
thatguynamedandy: Yes I think I agree, I initially had this but I wasn't sure if there was ever a use case for clicking a disabled word. This might muddy the separation of concerns slightly as currently it's quite a stupid component, but it would safeguard against implementations still passing the onClick

src/charts/wordcloud/Word.css
thatguynamedandy: Actually I remember now, the default is the text selection cursor, but with `user-select: none` it doesn't make as much sense, so the this reverts it to the arrow.

PR #303, opened 2017-08-01
refactor(Modal): Swapped over from using react-modal to Subtree component

HHogg:
Axiom has it's own subtree rendering for the Position component, so made sense to reuse this for the Dialog too, which allows us to remove react-modal (-1 dependency, ~17kb less on build size).

I've kept the API the same so not to introduce any breaking change. The opening and closing animations are no longer there so behaves the same as the Context components. These will be added back in when we go through and do an overhaul on mounting/unmounting animations.


pmsorhaindo:
Nuking a ticket from the road map! 🎉  nice!

style-guide/components/ExampleConfig/ExampleConfig.js
```diff
@@ -15,15 +16,19 @@ export default class ExampleConfig extends Component {
     initialPropOptions: PropTypes.object,
     initialProps: PropTypes.object,
     propTypes: PropTypes.object.isRequired,
+    visual: PropTypes.bool,
   };
 
   static defaultProps = {
     initialPropOptions: {},
+    visual: true,
```
pmsorhaindo: could possibly be named better?

src/components/subtree/Subtree.js
```diff
@@ -0,0 +1,89 @@
+import PropTypes from 'prop-types';
+import { Component } from 'react';
+import ReactDOM from 'react-dom';
+
+export default class Subtree extends Component {
+  static propTypes = {
+    children: PropTypes.node,
+    parentNode: PropTypes.object,
+    render: PropTypes.bool.isRequired,
```
pmsorhaindo: possibly rename this to `shouldRender` as render could be confused for a function.

PR #304, opened 2017-08-01
fix(BarChart): Remove mouseover interaction on 0% value bars

HHogg:
Removes any interactions and displaying the label when percentage is exactly 0%.

### DIFF
--- a/style-guide/components/ExampleConfig/ExampleConfig.js
+++ b/style-guide/components/ExampleConfig/ExampleConfig.js
@@ -59,7 +59,7 @@ export default class ExampleConfig extends Component {
   updateState(state) {
     window.history.pushState('', '',
       `${window.location.pathname}?state=${
-          window.encodeURI(JSON.stringify(state))}`);
+          window.encodeURIComponent(JSON.stringify(state))}`);
     this.setState(state);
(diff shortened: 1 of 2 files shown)


style-guide/components/ExampleConfig/ExampleConfig.js
```diff
@@ -59,7 +59,7 @@ export default class ExampleConfig extends Component {
   updateState(state) {
     window.history.pushState('', '',
       `${window.location.pathname}?state=${
-          window.encodeURI(JSON.stringify(state))}`);
+          window.encodeURIComponent(JSON.stringify(state))}`);
```
pmsorhaindo: Are you decoding this at any point? because now you'll need to use decodeURIComponent also.

pmsorhaindo: A quick search would indicate you're never decoding it.

HHogg: [`window.URLSearchParams` handles that](https://github.com/BrandwatchLtd/axiom/pull/304/files#diff-f791122d67169f9741c9b36414eee456R32)

PR #306, opened 2017-08-02
docs(Card): Swapped over to config example

HHogg:


src/components/card/Card.js
```diff
@@ -5,8 +5,11 @@ import { Base } from 'bw-axiom';
 
 export default class Card extends Component {
   static propTypes = {
+    /* Applies styling to indicate the Card has been activated */
```
benji6: missing a `*` here

PR #307, opened 2017-08-03
feat(TextInputIcon): Added optional align prop which can be set to left and defaults to right

benji6:
screen shot 2017-08-03 at 10 27 43
https://feat-textinputicon-align-left--axiom-rules.netlify.com/docs/components/form

### DIFF
--- a/src/components/form/TextInput.css
+++ b/src/components/form/TextInput.css
@@ -1,18 +1,3 @@
-.ax-input__container {
-  display: flex;
-  position: relative;
-  flex: 1 1 auto;
-  align-items: center;
-  border-width: var(--component-border-width);
-  border-style: solid;
-  border-radius: var(--component-border-radius);
-  border-color: var(--cmp-input-border-color);
-  background-color: var(--cmp-input-background-color);
-  transition-property: border-color;
-  transition-duration: var(--transition-time-base);
-  transition-timing-function: var(--transition-function);
-}
-
 .ax-input {
   flex: 1 1 auto;
   border: 0;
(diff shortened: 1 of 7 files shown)


src/components/form/TextInputIcon.js
```diff
@@ -1,19 +1,29 @@
+import classnames from 'classnames';
 import PropTypes from 'prop-types';
 import React, { Component } from 'react';
 import { Icon, Link } from 'bw-axiom';
 
 export default class TextInputIcon extends Component {
   static propTypes = {
+    align: PropTypes.oneOf(['left', 'right']).isRequired,
```
thatguynamedandy: You could probably drop `isRequired` as it has a default.

benji6: yeah I'm not sure if that even does anything if there are defaults! I'll remove it :D

src/components/form/TextInput.css
```diff
@@ -33,6 +33,10 @@
   }
 }
 
+.ax-input__icon--align-left {
+  order: -1;
```
HHogg: ❤️

src/components/form/TextInputIcon.js
```diff
@@ -1,19 +1,29 @@
...
+    const className = classnames('ax-input__icon', {
+      [`ax-input__icon--align-${align}`]: align,
+    });
 
     return (
-      <span className="ax-input__icon">
-        <Link { ...this.props } style="subtle">
-          <Icon name={ name }  size="1rem" />
+      <span className={ className }>
+        <Link { ...rest } style="subtle">
```
HHogg: I guess the Link part should now be conditional? Should the TextInputIcons on the left appear clickable?

HHogg: Yeah the `onClick` would also be consistent with other components like Bars and DataPoints.

src/components/form/TextInput.css
benji6: <3 flexbox!

src/components/form/TextInputIcon.js
HHogg: Yeah I'm not sure about warning but removing the `isRequired` will be correct for the API docs :)

PR #308, opened 2017-08-03
fix(ContextBox): Separate scrollable areas within a context popup

pmsorhaindo:
screen shot 2017-08-03 at 2 22 56 pm

https://598322b0424ef2766dc9944d--librarian-chimpanzee-34483.netlify.com/docs/components/context

### DIFF
--- a/src/components/context/ContextBox.css
+++ b/src/components/context/ContextBox.css
@@ -1,5 +1,6 @@
 .ax-context-box {
   position: relative;
+  overflow-y: auto;
 }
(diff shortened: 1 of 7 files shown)


src/components/context/ContextBox.css
```diff
@@ -1,5 +1,6 @@
 .ax-context-box {
   position: relative;
+  overflow-y: scroll;
```
benji6: I think this should be auto

PR #312, opened 2017-08-08
Update README.md for direct CSS setup

Binarytales:


README.md
```diff
@@ -38,6 +38,20 @@ The compiled CSS file is stored on the Brandwatch Google Cloud Platform CDN or t
 http://axiom.bwcom.io/axiom.{VERSION}.min.css
 ```
 
+You may also need to set some intial CSS for the `html` and `body` elements, especially if you want to support theme switching or want to use the dark theme:
```
HHogg: This is actually relevant for all projects, so it might bet better under the "Getting Started" section as opposed to "Using just the CSS". Maybe create a `css-setup.md` under `/docs`?

README.md
```diff
@@ -38,6 +38,20 @@ The compiled CSS file is stored on the Brandwatch Google Cloud Platform CDN or t
...
+
+```
+html {
+  height: 100%;
+}
+
+body {
+  background-color: var(--color-context-background);
+  color: var(--color-context-text);
+  min-height: 100%;
```
HHogg: These height properties on html and body, aren't really a requirement an app can use Axiom and it look correct without them.

Binarytales:
@HHogg Thanks for the review Harry. I just stuck this in via GitHub but I have the project checked out properly now so can tidy this up 🤓

HHogg:
Awesome @Binarytales also please fork the repo rather than pushing branches to the main one :)

PR #313, opened 2017-08-08
fix(Base): Sticky behaviour to support safari css prefixing

HHogg:
Safari currently only supports `position: sticky` with a `-webkit` prefix. So the position styling needs to be in the CSS for the autoprefixing to work.


thatguynamedandy:
QA'd on https://safari-sticky-fix--hhogg-axiom.netlify.com

PR #318, opened 2017-08-10
docs(Forms and Inputs): Updated to use ExampleConfig

HHogg:


src/components/form/TextInput.js
```diff
@@ -7,19 +7,38 @@ import './TextInput.css';
...
     size: PropTypes.oneOf(['medium', 'large']),
+    /** Visual style variations of the input field */
     style: PropTypes.oneOf(['overlay']),
+    /** Type of the input field */
+    type: PropTypes.oneOf(['password', 'text']),
+    /** Applies styling to indicate the users input was valid */
     valid: PropTypes.bool,
+    /** Value of the input field */
+    value: PropTypes.any,
+    /** Handler for when the input field is unfocused */
```
thatguynamedandy: really minor, but I wonder if you could just use `blurred` instead of `unfocused` the audience should understand it.

PR #319, opened 2017-08-10
docs(Grid): Moved over to ExampleConfig

HHogg:


src/components/grid/Grid.js
```diff
@@ -6,38 +6,44 @@ import './Grid.css';
...
-      PropTypes.bool,
-      PropTypes.oneOf(['small', 'medium', 'large']),
-    ]),
-    gutters: PropTypes.oneOfType([
-      PropTypes.bool,
-      PropTypes.oneOf(['tiny', 'small', 'medium', 'large']),
-    ]),
+    /**
+     * Applies fill styling for all GridCell children.
+     * See GridCell for fill explanation
```
thatguynamedandy: Is it possible to link in these docs? Nice issue ticket otherwise for someone to do.

HHogg: Yeah that would be nice... not sure there is any support in the docgenerator we are using but something to look into 👍

PR #320, opened 2017-08-10
docs(Icon): Moved over to use ExampleConfig

HHogg:
.... and `fix(IconIndicator): Primary subtle coloring`


src/components/icon/IconIndicator.css
```diff
@@ -41,7 +41,7 @@
   &.ax-icon-indicator--info { color: var(--color-ui-primary--light); }
 }
 
-.ax-icon-indication--primary,
```
thatguynamedandy: Good spot!

PR #321, opened 2017-08-10
docs(Image): Moved over to use ExampleConfig

HHogg:


src/components/image/ImageCircle.js
```diff
@@ -6,13 +6,21 @@ import './ImageCircle.css';
 
 export default class ImageCircle extends Component {
   static propTypes = {
+    /** Border applied around the image */
     border: PropTypes.oneOf(['small', 'large']),
+    /** Fallback content when the image fails to load */
     children: PropTypes.node,
+    /** Name of the icon to added to the hover overlay */
```
thatguynamedandy: Slight language malfunction here and the one below. `to be added`?

src/components/image/ImageCircle.js
```diff
@@ -6,13 +6,21 @@ import './ImageCircle.css';
...
     overlayIconName: PropTypes.string,
+    /** Size of the icon to added to the hover overlay */
     overlayIconSize: PropTypes.string,
+    /** Size (with unit) that will applied to the width and height of the image */
```
thatguynamedandy: Another missing `be` here too I think 🐝

src/components/image/ImageCircle.js
```diff
@@ -6,13 +6,21 @@ import './ImageCircle.css';
...
   static propTypes = {
+    /** Border applied around the image */
     border: PropTypes.oneOf(['small', 'large']),
+    /** Fallback content when the image fails to load */
     children: PropTypes.node,
+    /** Name of the icon to be added to the hover overlay */
     overlayIconName: PropTypes.string,
+    /** Size of the icon to be added to the hover overlay */
     overlayIconSize: PropTypes.string,
+    /** Size (with unit) that will be applied to the width and height of the image */
```
pmsorhaindo: Super picky but, when adding `(with unit)` it implies that the `overlayIconSize` above that doesn't have `(with unit)` in its explanation doesn't take a unit. I'm not 100% sure but I think it's the case that all string properties named size take a number with unit, so I think this parenthesis should be everywhere or nowhere and just implied.

pmsorhaindo:
Small super not a big deal comment but I approve.

src/components/image/ImageCircle.js
HHogg: Yeah good point... I agree it should just be implied as it's the general pattern in Axiom. I'll remove them 👍

PR #335, opened 2017-08-14
feat(Chart): Adds ability to set a minimum height on Charts and align…

HHogg:
… ChartBody content, similar to Panels

### DIFF
--- a/src/charts/chart/Chart.js
+++ b/src/charts/chart/Chart.js
@@ -6,13 +6,19 @@ import './Chart.css';
 export default class Chart extends Component {
   static propTypes = {
     children: PropTypes.node.isRequired,
+    /**
+     * Sets a minimum height on the Chart. Useful to center a Progress inside
+     * the ChartBody while a visual is loading or a no data message when the
+     * visual won't show any data.
+     */
+    minimumHeight: PropTypes.string,
   };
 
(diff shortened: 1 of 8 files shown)


src/charts/chart/Chart.test.js
```diff
@@ -0,0 +1,23 @@
+import React from 'react';
+import renderer from 'react-test-renderer';
+import { Chart } from 'bw-axiom';
+
+function getComponent(props = {}) {
+  return renderer.create(
+    <Chart { ...props }></Chart>
```
pmsorhaindo: Shouldn't we use a self closing tag here?

src/charts/chart/Chart.js
```diff
@@ -6,13 +6,19 @@ import './Chart.css';
 export default class Chart extends Component {
   static propTypes = {
     children: PropTypes.node.isRequired,
+    /**
+     * Sets a minimum height on the Chart. Useful to center a Progress inside
+     * the ChartBody while a visual is loading or an no data message when the
+     * visual won't show any data.
```
pmsorhaindo: ...`or a no data message when the visual`...

pmsorhaindo:
Super small changes.. but good to go.

PR #338, opened 2017-08-21
fix(InlineGroup): Removes margin from last child element

thatguynamedandy:


thatguynamedandy:
https://inline-group-last-child-margin-fix--andyp-axiom.netlify.com

PR #343, opened 2017-08-25
chore(stylelint-config): pull in stylelints lint rules from the stylelint-config-axiom package

pmsorhaindo:


src/charts/wordcloud/Word.css
```diff
@@ -16,6 +15,7 @@
   animation-fill-mode: forwards;
   animation-timing-function: var(--transition-function);
   user-select: none;
+  fill-opacity: 1;
```
HHogg: Rules at the end is kind of a catch all. It might be better to add both of these to the correct place over in the [config](https://github.com/BrandwatchLtd/stylelint-config-axiom/blob/master/index.js)...

For context, the order of the properties follows similar to [smacss](https://smacss.com/book/formatting) or [concentric css](http://rhodesmill.org/brandon/2011/concentric-css/) pattern. Where properties are ordered on the impact to that element or the elements around them.

thatguynamedandy: @pmsorhaindo poke.

PR #355, opened 2017-09-11
v26 - [+AlertDialog, +AlertIcon, +Animicons, +ChangePassword]

HHogg:
See commit messages.


HHogg:
[Netlify Deploy](https://alert-panels--hhogg-axiom.netlify.com/docs/components/alert-panel?state=%7B%22AlertPanel%22%3A%7B%22props%22%3A%7B%22isOpen%22%3Atrue%7D%7D%7D)

src/components/alert-icon/AlertIcon.js
```diff
@@ -0,0 +1,51 @@
...
+
+export const ALERT_ICON_NAME_MAP = {
+  error: 'warning',
+  info: 'information',
+  success: 'tick',
+  warning: 'warning',
+};
+
+const SIZE_MAP = {
+  'small': '1rem',
```
benji6: I guess the linter doesn't say anything here and this is the tiniest nitpick but maybe these shouldn't be strings like in the `ALERT_ICON_NAME_MAP`?

src/components/alert-panel/AlertPanelBody.js
```diff
@@ -0,0 +1,19 @@
+import PropTypes from 'prop-types';
+import React, { Component } from 'react';
+
+export default class AlertPanelBody extends Component {
+  static propTypes = {
+    /** The main for AlertPanel that occupies most of the space */
```
benji6: This confused me - maybe just `The main body for AlertPanel` or something like that?

src/components/alert-panel/AlertPanelHeader.js
```diff
@@ -0,0 +1,32 @@
+import PropTypes from 'prop-types';
+import React, { Component } from 'react';
+import { Alert, Strong } from 'bw-axiom';
+
+export default class AlertPanelHeader extends Component {
+  static propTypes = {
+    /** Header content inside the AlertPanel, a good place for a title */
+    children: PropTypes.node.isRequired,
+    /** Handler attached to the close button inside the header */
+    onRequestClose: PropTypes.func.isRequired,
```
benji6: Would it be better if you just use `onRequestClose` from the `AlertPanel` rather than having the user provide it to this component as well?

src/components/alert-panel/AlertPanel.js
```diff
@@ -0,0 +1,44 @@
...
+
+export default class AlertPanel extends Component {
+  static propTypes = {
+    children: PropTypes.node.isRequired,
+    /** Visibility toggle for the AlertPanel */
+    isOpen: PropTypes.bool.isRequired,
+    /** Alert type */
+    type: PropTypes.oneOf(['error', 'info', 'success', 'warning']).isRequired,
+    /** Callback for closing the AlertPanel by clicking on the overlay */
+    onRequestClose: PropTypes.func.isRequired,
```
benji6: I would have been tempted just to call this `onClose`. I suppose `onRequestClose` is more accurate in a way as the developer has to close the modal themselves - is that the reason for the name?

src/components/alert-panel/AlertPanelBody.js
HHogg: Ah yeah I think that's a typo, it should say 'content'

src/components/alert-panel/AlertPanel.js
HHogg: It was just left over from the removal of `react-modal`, so as not to introduce a breaking change. But yes I agree with you that it would be better called `onClose`.

src/components/alert/Alert.js
```diff
@@ -1,27 +1,46 @@
...
 import './Alert.css';
 
 export default class Alert extends Component {
   static propTypes = {
+    /** Content displayed next to the AlertIcon */
     children: PropTypes.node.isRequired,
-    color: PropTypes.oneOf(['success', 'warning', 'error', 'info']),
+    /** Size of the Alert */
+    size: PropTypes.oneOf(['small', 'medium']),
+    /** Type of Alert that affects the colouring and icon */
```
benji6: "colouring" should be "coloring"

PR #360, opened 2017-09-15
fix(Position): Positioning incorrect for start and end offsets

HHogg:
Popper offset to account for the margin that is applied around context when an arrow is used.

### DIFF
--- a/src/components/position/Position.js
+++ b/src/components/position/Position.js
@@ -81,7 +81,7 @@ export default class Position extends Component {
   }
 
   createPopper() {
-    const { flip } = this.props;
+    const { flip, showArrow } = this.props;
     const { placement } = this.state;
 
     return new popperJS(this._target, this._content, {
@@ -90,6 +90,7 @@ export default class Position extends Component {
       placement: placement,
       modifiers: {
         arrow: {
+          enabled: showArrow,
           element: this._arrow,
         },
         flip: {
--- a/src/components/context/Context.css
+++ b/src/components/context/Context.css
@@ -11,7 +11,7 @@
 }
 
 .ax-context {
-  margin: calc(var(--cmp-context-tip-edge-to-edge) * 0.5);
+  position: relative;
   border-radius: var(--component-border-radius);
   box-shadow: var(--drop-shadow-1), var(--drop-shadow-2);
 }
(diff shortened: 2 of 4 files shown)


src/components/position/Position.js
```diff
@@ -93,7 +98,11 @@ export default class Position extends Component {
           behavior: getPlacementFlipOrder(placement, flip),
         },
         inner: { enabled: false },
-        offset: { enabled: false },
+        offset: {
+          enabled: showArrow,
+          fn: offset,
+          offset: -12, /* Half the arrow size (1.5rem/2) */
```
pmsorhaindo: Just wanted to confirm, is it not possible to set the 1.5 rem size value for the arrow as a css variable and access it with `getCSSVar` in order to calculate it here? Just so in the extremely unlikely event that we change the arrow size in the future we don't have to update this value too.

HHogg: Unfortunately not, we still can't use `getCSSVar` at the moment due to IE11 support. Also popper only accepts `%`, `px` and `vw/vh` units.

FezVrasta: Wouldn't be easier to handle this from a CSS prospective?

HHogg: Hey @FezVrasta! 

We originally had it like the popper website where the margin is applied to the side the arrow appears. But this would cause a flicker when repositioning to a different axis. (shown in gif below). 

popper

I think this flicked is due to the margin changing the dimensions and then triggering another reposition, and it would then leave it repositioned incorrectly inset, ignoring the margin.

So we managed to get around this by applying the margin around every side no matter when the arrow is placed. This resolved the flicker but it then (rightly so) takes into account that margin when calculating the position... and would leave the arrow handing off the content within the margin. Like the screenshot below. 

screen shot 2017-09-15 at 11 32 45

I was thinking about logging an issue about the possibility of having a `padding` option around the `offset` modifier (similar to the `preventOverflow` modifier). 

What do you think?

HHogg: _(this flicker is more noticeable in one of our applications when it hits the bottom of the window during a scrolling)_

FezVrasta: I think you can add some margin on the sides of the arrow to get around the problem

HHogg: Thanks for the advice @FezVrasta, I was indeed able to solve it with just CSS happy days! Had to shift around a little bit where the margins were applied and where the arrow was positioned relative to, but the margins on the arrow element did the trick. Thanks again.

PR #361, opened 2017-09-15
Disable submit in the ChangePassword composite once the form has been submitted

pmsorhaindo:
This PR prevents the ChangePassword composite's submit button pressed multiple times when there is a in the form being re rendered with an error or being closed.

### DIFF
--- a/src/composites/change-password/ChangePassword.js
+++ b/src/composites/change-password/ChangePassword.js
@@ -86,7 +86,7 @@ export default class ChangePassword extends Component {
 
     return (
       <Dialog { ...rest } onRequestClose={ onRequestClose } size="medium">
-        <DialogHeader onRequestClose={ onRequestClose }>
+        <DialogHeader>
           <Heading textSize="headtitle">
             Change Password
           </Heading>
--- a/src/composites/change-password/ChangePassword.js
+++ b/src/composites/change-password/ChangePassword.js
@@ -44,45 +44,8 @@ export default class ChangePassword extends Component {
     ],
   };
 
-  constructor(props) {
-    super(props);
-
-    this.state = {
-      currentPassword: '',
-      newPassword: '',
-      confirmPassword: '',
-    };
-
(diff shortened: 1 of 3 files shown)


src/composites/change-password/ChangePassword.js
```diff
@@ -50,12 +50,17 @@ export default class ChangePassword extends Component {
...
       confirmPassword: '',
+      isSubmitting: false,
     };
 
     this.handleSubmit = this.handleSubmit.bind(this);
     this.handlePasswordChange = this.handlePasswordChange.bind(this);
   }
 
+  componentWillReceiveProps() {
+    this.setState({ isSubmitting: false });
```
HHogg: This is something that should probably be provided through props. Because this composite doesn't know when the async submission has finished. 

Aside from that, resetting it whenever props are received would likely lead to some weird behaviour... for example during a submission a `rule.label` changes.

src/composites/change-password/ChangePasswordControls.js
```diff
@@ -21,13 +21,16 @@ export default class ChangePasswordControls extends Component {
...
+            style="secondary"
+            type="button">
           Cancel
         </Button>
         <Button
             data-ax-at={ atIds.ChangePassword.submit }
             disabled={ submitDisabled }
-            onClick={ onSubmit }>
+            onClick={ onSubmit }
+            type="submit">
```
HHogg: Does this work as the buttons aren't directly inside the `<form>`?

package.json
```diff
@@ -66,7 +66,7 @@
     "glob": "^7.1.2",
     "html-webpack-plugin": "^2.24.1",
     "jest": "^19.0.2",
-    "js-beautify": "^1.6.2",
+    "js-beautify": "1.6.14",
```
HHogg: This change shouldn't be needed. The `yarn.lock` file protects us against the `leftpad.v2` issue.

pmsorhaindo: I'm sure I had a fixup for this but squashed it into the wrong commit.

PR #365, opened 2017-09-22
fix(Paths): Updated to use relative paths to better support tree-shaking

HHogg:
We think the absolute paths from within the module may be causing some circular dependence weirdness that is preventing tree shaking from working. This converts all those absolute internal paths to relative. 

I've left the aliases in for Style Guide because it includes all of the paths already so we might as well benefit from importing as if the library was like any other public library.


HHogg:
Closing this as it looks like tree shaking is not currently possible due to side effects caused by static properties on classes and extending classes. Currently looking into a flat structure option at build time to enable easy relative imports.

PR #366, opened 2017-09-25
fix(Dropdown): Retaining opened state when initially disabled.

HHogg:
When Dropdown was disabled (using the enabled prop on Position) it would still set the visible state when clicked, and could not be taken out of the visible state because no mask had been added because it was disabled.

### DIFF
--- a/src/components/dropdown/Dropdown.js
+++ b/src/components/dropdown/Dropdown.js
@@ -17,6 +17,10 @@ export default class Dropdown extends Component {
      */
     children: PropTypes.node.isRequired,
     /**
+     * Adds control to enable or disable showing the DropdownContent
+     */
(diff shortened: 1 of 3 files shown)


src/components/dropdown/Dropdown.js
```diff
@@ -17,6 +17,10 @@ export default class Dropdown extends Component {
      */
     children: PropTypes.node.isRequired,
     /**
+     * Adds control to conditional enable or disabled the positioning logic.
```
benji6: There's a couple typos here and below - should probably be "conditionally" and "disable". Also maybe it'd be clearer to say it enables or disables displaying the DropdownContent rather than the positioning logic?

PR #370, opened 2017-09-27
feat(locales): localise axiom composites

pmsorhaindo:


src/utils/locales/index.js
```diff
@@ -0,0 +1,16 @@
...
+const languageFileMap = {
+  de,
+  en,
+  es,
+  fr,
+};
+
+
+export const translate = (lng = 'en', textId) => {
+  return languageFileMap[lng][textId] || textId;
```
HHogg: I don't think outputting the `textId` is going to be super useful?

style-guide/components/Layout/LayoutSidebar.js
```diff
@@ -2,14 +2,16 @@ import PropTypes from 'prop-types';
 import React, { Component } from 'react';
 import GithubLink from '../GithubLink/GithubLink';
 import ThemeSwitcher from '../ThemeSwitcher/ThemeSwitcher';
+import LanguageSwitcher from '../LanguageSwitcher/LanguageSwitcher';
 
 export default class LayoutSidebar extends Component {
   static propTypes = {
     children: PropTypes.node.isRequired,
+    handleSwitchLanguage: PropTypes.func.isRequired,
```
HHogg: This should really be `onSwitchLanguage` as the handler is being given. The API exposed by this component registers the handlers. 

```js
static propTypes = {
  onClick: PropTypes.func,
  onSubmit: PropTypes.func,
  onSwitchLanguage: PropTypes.func,
};

const { onClick, onSubmit, onSwitchLanguage  } = this.props; // From parent 

<Component 
  onClick={ onClick }
  onSubmit={ onSubmit }
  onSwitchLanguage={ onSwitchLanguage }
  onMouseEnter={ this.handleMouseEnter } // Handle by this component 
  onMouseLeave={ this.handleMouseLeave } />
```

src/utils/locales/index.js
pmsorhaindo: I had a little console.error in here before, that'd probably be more useful.

PR #371, opened 2017-09-29
fix(locales): Fallback to language code

pmsorhaindo:


src/utils/locales/index.js
```diff
@@ -12,5 +12,13 @@ const languageFileMap = {
 
 
 export default (axiomLanguage = 'en', textId) => {
+  if (!languageFileMap[axiomLanguage]) {
+    axiomLanguage = axiomLanguage.split(/[^A-Za-z]/)[0];
```
HHogg: This could benefit from a test 😉

PR #373, opened 2017-09-29
Support for babel-plugin-axiom 

HHogg:
These are changes needed to support babel-plugin-axiom


HHogg:
@benji6 I've left it in there for the style-guide.

PR #374, opened 2017-09-29
Allow Console to Support Theme Switching

TomWija:


src/components/platform/Canvas.js
```diff
@@ -1,5 +1,6 @@
 import PropTypes from 'prop-types';
 import React, { Component } from 'react';
+import { Base } from 'bw-axiom';
```
HHogg: This is a recent change, but imports now need to be relative inside Axiom itself. So this should be 

```js
```

src/components/platform/Canvas.css
```diff
@@ -1,6 +1,7 @@
 .ax-platform__canvas {
   min-height: 100vh;
   margin-left: var(--cmp-platform-dock-size);
+  background-color: var(--color-context-canvas);
```
HHogg: I'm not sure it makes sense to have the background colour set on the canvas. For us we have a couple of use cases where the canvas colour is different on a single theme. We set this on the body and the canvas allows that underlying colour to come through.

PR #379, opened 2017-10-02
 feat(Status404): Added Status404 composite

HHogg:


style-guide/components/LanguageSwitcher/LanguageSwitcher.js
```diff
@@ -12,24 +12,27 @@ export default class ThemeSwitcher extends Component {
...
-            style="subtle"
-            title="French">fr</Link>
-      </Paragraph>
+      <Base space="x2" textCenter>
+        <Heading space="x0"><Strong>Language</Strong></Heading>
+        <Paragraph space="x0" textColor="subtle">
+          <Link
+              onClick={ () => onSwitchLanguage('en') }
+              style="subtle"
+              title="English">En </Link>/
```
pmsorhaindo: super not a big deal but I feel like these should be lower case or completely capitalised as they're codes. Abbreviations should follow with a `.`

src/composites/login/Login.js
```diff
@@ -18,21 +17,23 @@ import t from '../../utils/locales';
...
-  };
-
   static propTypes = {
     /** Name of the application that appears in the header */
     application: PropTypes.string.isRequired,
     /** Background image that is behind the login form */
     backgroundImage: PropTypes.string.isRequired,
     /** An error message that will be placed on the page */
     error: PropTypes.string,
+    /**  */
```
pmsorhaindo: empty comment :)

PR #383, opened 2017-10-04
React16 Upgrade

HHogg:
**tldr;** 
* React16, ReactRouter4 and ReactHotLoader upgrades done. 
* Swapped over the using `babel-env`
* New Portal component, no more Subtree component.
* Updated node version to 8.5.0

_____
This turned out to be a bit more than a quick `yarn upgrade` because. 

* ReactHotLoader upgrade has some outstanding issues *
  * https://github.com/gaearon/react-hot-loader/issues/304 : Which effected how Axiom does type comparisons in a few places. 
  * https://github.com/gaearon/react-hot-loader/issues/593 : Which required to transform es2015 classes down. 
* ReactRouter upgrade, which changed how we structure the routes, but by changing it we benefit from not having to specify the routes twice in the static webpack config.

**Notes**
* I've needed to switch React env to development for tests because of this issue https://github.com/facebook/react/issues/10938
* There's a warning on some tests that use requestAnimationFrame because of issue _best described_ here https://github.com/facebookincubator/create-react-app/issues/3199 (but is in React)
* I've removed the esm build I don't think it's benefiting us at the moment.


package.json
```diff
@@ -40,17 +39,20 @@
...
-    "babel-preset-react": "^6.11.1",
-    "babel-preset-react-hmre": "1.1.1",
-    "babel-preset-stage-2": "^6.24.1",
+    "babel-core": "^6.26.0",
+    "babel-loader": "^7.1.2",
+    "babel-plugin-transform-class-properties": "^6.24.1",
+    "babel-plugin-transform-es2015-classes": "^6.24.1",
+    "babel-plugin-transform-es2015-modules-commonjs": "^6.26.0",
+    "babel-plugin-transform-object-rest-spread": "^6.26.0",
+    "babel-plugin-transform-react-remove-prop-types": "^0.4.9",
```
HHogg: I need to remove this.

src/utils/isComponent.js
```diff
@@ -0,0 +1,6 @@
+import { isValidElement } from 'react';
+
+export default (component, Component) =>
```
pmsorhaindo: I feel weird about having the parameters being differentiated by only a capital letter. Could one be named componentInstance or something? 🤷‍♂️

HHogg: I'm kind of following typical convention of lowercase being the instance and Titlecase being the class. 

```js
class Car {}
const car = new Car();
car.type === Car; // true
```

But I'm happy to change it :D

package.json
HHogg: No I don't

.babelrc
```diff
@@ -1,28 +1,27 @@
...
+    "transform-object-rest-spread"
+  ],
+  "presets": [
+    "react",
+    ["env", {
+      "targets": {
+        "browsers": [
+          "last 2 Firefox versions",
+          "last 2 Chrome versions",
+          "last 2 Edge versions",
```
stefanbuck: This change kept me busy for awhile. In Analytics frontend we use uglifyjs which can't handle es6 syntax. Therefore my build was failing with `Unexpected token`. As a quick fix for me I added ie 11. Are there any plans to support ie 11?

HHogg: I don't think there is any harm adding IE11 here for now. I'll add it in 👍

PR #384, opened 2017-10-04
fix(StatusPage): adds AT hook for 404 page

thatguynamedandy:


thatguynamedandy:
Closing to do in consuming apps instead.

PR #386, opened 2017-10-05
fix(ChangePassword): perform client-side validation

pmsorhaindo:
https://librarian-chimpanzee-34483.netlify.com/docs/composites/change-password/

### DIFF
--- a/src/composites/change-password/ChangePassword.js
+++ b/src/composites/change-password/ChangePassword.js
@@ -1,5 +1,6 @@
 import PropTypes from 'prop-types';
 import React, { Component } from 'react';
+import omit from 'lodash.omit';
 import Alert from '../../components/alert/Alert';
 import Dialog from '../../components/dialog/Dialog';
 import DialogBody from '../../components/dialog/DialogBody';
@@ -18,6 +19,8 @@ export default class ChangePassword extends Component {
   static propTypes = {
     /** An error message that will be placed on the page */
     error: PropTypes.string,
+    /** Allows error styles to be applied to the current password text box manually */
+    isCurrentPasswordInvalid:  PropTypes.bool,
     /** Visibility toggle for the Dialog */
     isOpen: PropTypes.bool.isRequired,
     /** Toggles the disabled property on the submit button */
@@ -37,41 +40,80 @@ export default class ChangePassword extends Component {
   };
 
   static defaultProps = {
+    isCurrentPasswordInvalid: false,
     isSubmitting: false,
     rules: [
-      { label: '8-characters', pattern: /^.{8,}$/ },
-      { label: '1-numeric-character', pattern: /^.*[0-9].*$/ },
-      { label: '1-uppercase-character', pattern: /^.*[A-Z].*$/ },
-      { label: '1-non-alphanumeric-character', pattern: /^.*[^a-zA-Z\d:].*$/ },
+      {
+        errorMessage: '8-characters-error',
+        label: '8-characters',
+        pattern: /^.{8,}$/,
+      },
+      {
+        errorMessage: '1-numeric-character-error',
+        label: '1-numeric-character',
+        pattern: /^.*[0-9].*$/,
+      },
+      {
+        errorMessage: '1-uppercase-character-error',
+        label: '1-uppercase-character',
+        pattern: /^.*[A-Z].*$/,
+      },
+      {
+        errorMessage: '1-non-alphanumeric-character-error',
+        label: '1-non-alphanumeric-character',
+        pattern: /^.*[^a-zA-Z\d:].*$/,
+      },
     ],
   };
 
+  constructor(props) {
+    super(props);
+
+    this.state = {
+      internalError: false,
+    };
+    this.handleSubmit = this.handleSubmit.bind(this);
+    this.handleError = this.handleError.bind(this);
+  }
+
+  handleError(errorMessage) {
+    this.setState({ internalError: errorMessage });
+  }
+
+  handleSubmit(currentPassword, newPassword) {
+    this.setState({ internalError: false });
+    this.props.onSubmit({ currentPassword, newPassword });
+  }
+
   render() {
-    const { error, isSubmitting, onRequestClose, onSubmit, rules, ...rest } = this.props;
+    const { error, isCurrentPasswordInvalid, isSubmitting, onRequestClose, rules, ...rest } = this.props;
+    const { internalError } = this.state;
     const { axiomLanguage } = this.context;
 
     return (
-      <Dialog { ...rest } onRequestClose={ onRequestClose } size="medium">
+      <Dialog { ...omit(rest, ['onSubmit']) } onRequestClose={ onRequestClose } size="medium">
         <DialogHeader>
           <Heading textSize="headtitle">
             { t(axiomLanguage, 'change-password-title') }
           </Heading>
         </DialogHeader>
 
-        { error && (
-          <Alert space="x4" type="error" >
+        { (internalError || error) && (
+          <Alert type="error">
             <Paragraph data-ax-at={ atIds.ChangePassword.error }>
-              { error }
+              { internalError || error }
             </Paragraph>
           </Alert>
         ) }
 
         <DialogBody>
           <ChangePasswordForm
+              isCurrentPasswordInvalid={ isCurrentPasswordInvalid }
               isSubmitting={ isSubmitting }
               onCancel={ onRequestClose }
+              onError={ this.handleError }
               onRequestClose={ onRequestClose }
-              onSubmit={ onSubmit }
+              onSubmit={ this.handleSubmit }
               rules={ rules } />
         </DialogBody>
       </Dialog>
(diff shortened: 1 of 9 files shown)


src/composites/change-password/ChangePasswordForm.js
```diff
@@ -47,8 +47,34 @@ export default class ChangePasswordForm extends Component {
...
+      return onSubmit({
+        error: invalidRules.reduce((acc, { label }, index, arr) => {
+          if (arr.length === 1) return `${acc} ${label}`;
+          if (index === arr.length - 1) return `${acc.slice(0, -1)} and ${label}`;
+          return ` ${acc} ${label},`;
+        }, 'Sorry your new password must include'),
+      });
+    }
     onSubmit({ currentPassword, newPassword });
   }
```
Vitor-Carvalho: I think we are missing the current password validation, right?

pmsorhaindo: That happens server side, the frontend doesn't know the user's password. 💃

src/composites/change-password/ChangePassword.js
```diff
@@ -46,8 +46,27 @@ export default class ChangePassword extends Component {
...
+    this.handleSubmit = this.handleSubmit.bind(this);
+  }
+
+  handleSubmit({ currentPassword, newPassword, error }) {
+    if (error) {
+      return this.setState({ internalError: error });
+    }
+
+    this.setState({ internalError: false });
+    return this.props.onSubmit({ currentPassword, newPassword });
```
HHogg: Super minor, but is this `return` needed?

src/composites/change-password/ChangePassword.js
```diff
@@ -46,8 +46,27 @@ export default class ChangePassword extends Component {
...
+  }
+
   render() {
-    const { error, isSubmitting, onRequestClose, onSubmit, rules, ...rest } = this.props;
+    const { error, isSubmitting, onRequestClose, rules, ...rest } = this.props;
```
HHogg: The `onSubmit` will now be available in `rest` which is spread onto Dialog, this should probably be omitted?

src/composites/change-password/ChangePassword.js
```diff
@@ -46,8 +46,27 @@ export default class ChangePassword extends Component {
     ],
   };
 
+  constructor(props) {
+    super(props);
+
+    this.state = {
+      internalError: false,
```
benji6: I have a slight thing about calling things Errors when they're actually strings. I would probably name this `internalErrorMsg`. The problem then is that we're at odds with the API that expects an `error` prop so maybe don't worry about it

src/composites/change-password/ChangePassword.js
```diff
@@ -46,8 +46,27 @@ export default class ChangePassword extends Component {
...
+    };
+    this.handleSubmit = this.handleSubmit.bind(this);
+  }
+
+  handleSubmit({ currentPassword, newPassword, error }) {
+    if (error) {
+      return this.setState({ internalError: error });
```
HHogg: I'm wondering if this might cleared to separate out the error logic from the submit. Something like ...

```js
handleOnError() { ... }
handleOnSubmit() { ... }

<ChangePasswordForm
    onError={ this.handleOnError }
    onSubmit{ this.handleOnSubmit } />

if (!isFormFilledOut) {
  onError('...message');
}
```

src/composites/change-password/ChangePassword.js
pmsorhaindo: That doesn't trouble Dialog though right?

src/composites/change-password/ChangePassword.js
pmsorhaindo: Typescript would clarify things here 😂

src/composites/change-password/ChangePasswordForm.js
```diff
@@ -47,8 +47,34 @@ export default class ChangePasswordForm extends Component {
...
+
+    if (!isFormFilledOut) {
+      return  onSubmit({ error: 'Sorry, you need to complete all the password fields' });
+    }
+
+    if (!confirmPassword === newPassword) {
+      return  onSubmit({ error: 'Sorry, your password confirmation doesn\'t match' });
+    }
+
+    if (currentPassword === newPassword) {
```
HHogg: Even though we _can_ do this validation on the frontend, I don't think we should. It should be up to the backend to allow/disallow it.

pmsorhaindo: 💥  removed!

src/composites/change-password/ChangePassword.js
HHogg: `onSubmit` is actually quite a dangerous one to leak through because [as we found before](https://github.com/BrandwatchLtd/axiom/pull/59) it still get's fired when the events propagate up even when it's not on a form element. 

Unneeded Trickleling Props ™ shouldn't happen, as React quite rightly warns when non-standard props are put onto elements... and AFAIK all Axiom components handle this correctly.

src/composites/change-password/ChangePasswordForm.js
```diff
@@ -46,8 +47,30 @@ export default class ChangePasswordForm extends Component {
...
+      return onError({ internalError: t(axiomLanguage, 'complete-all-fields') });
+    }
+
+    if (confirmPassword !== newPassword) {
+      return  onError({ internalError: t(axiomLanguage, 'confirmation-doesnt-match') });
+    }
+
+    if (invalidRules.length > 0) {
+      return onError({
+        internalError: invalidRules.reduce((acc, { label }, index, arr) => {
```
benji6: why not just pass `internalError` as a string instead of an object with an `internalError` prop?

pmsorhaindo:
Hey guys unfortunately I have to dash off before getting the ChangePassword PR in. Here is a doc from design review https://docs.google.com/document/d/1N-vmCq2NTMrwTtfmNVCzLUCUbIRDRT9U6mvnOF2JTdY/edit#heading=h.6op1zumphfrq the issues raised have been partially addressed. with invalid states for text boxes being the main issue that I missed.

src/composites/change-password/ChangePassword.js
```diff
@@ -37,41 +40,80 @@ export default class ChangePassword extends Component {
...
   static defaultProps = {
+    isCurrentPasswordInvalid: false,
     isSubmitting: false,
     rules: [
-      { label: '8-characters', pattern: /^.{8,}$/ },
-      { label: '1-numeric-character', pattern: /^.*[0-9].*$/ },
-      { label: '1-uppercase-character', pattern: /^.*[A-Z].*$/ },
-      { label: '1-non-alphanumeric-character', pattern: /^.*[^a-zA-Z\d:].*$/ },
+      {
+        naturalLanguageLabel: '8-characters-error',
```
HHogg: This name doesn't really best describe what it is... looking at the translation keys, maybe it should just be called `error` or `errorMessage`?

src/composites/change-password/ChangePasswordForm.js
```diff
@@ -34,66 +36,104 @@ export default class ChangePasswordForm extends Component {
...
+        isConfirmPasswordInvalid: !confirmPassword.length,
+      });
+      return onError(t(axiomLanguage, 'complete-all-fields'));
+    }
+
+    if (invalidRules.length > 0) {
+      this.setState({
+        isNewPasswordInvalid: true,
+      });
+      return onError(invalidRules.reduce((acc, { naturalLanguageLabel }, index, arr) => {
```
HHogg: This method is a bit difficult to follow, it might benefit from moving this generic text joining function to utils... like 

```js
// listToSentence.js
export default (phrases) => phrases.reduce((acc, text ) => ...);

// ChangePasswordForm.js
import listToSentence from '../../utils/listToSentence';

if (invalidRules.length > 0) {
  return onError(listToSentence(invalidRules.map(({ error }) => t(axiomLanguage, error));
}
```

src/composites/change-password/ChangePasswordForm.js
```diff
@@ -34,66 +36,104 @@ export default class ChangePasswordForm extends Component {
...
+    const { axiomLanguage } = this.context;
+    const { confirmPassword, currentPassword, newPassword } = this.state;
+    const isFormFilledOut = confirmPassword.length && currentPassword.length && newPassword.length;
+    const invalidRules = this.props.rules.filter(rule => !rule.pattern.test(newPassword));
+
+    if (!isFormFilledOut) {
+      this.setState({
+        isCurrentPasswordInvalid: !currentPassword.length,
+        isNewPasswordInvalid: !newPassword.length,
+        isConfirmPasswordInvalid: !confirmPassword.length,
```
HHogg: It doesn't look like the invalid styles are being removed on submission, and also it's a bit confusing the same state being set in different places. It might better to always set state on all conditions in one place, this would also fix the style removal on submission, something like ... 

```js
  this.setState({
    isCurrentPasswordInvalid: !currentPassword,
    isNewPasswordInvalid: !newPassword || invalidRules.length > 0,
    isConfirmPasswordInvalid: !confirmPassword || confirmPassword !== newPassword,
  });

  if (!currentPassword ||  !newPassword || !confirmPassword) return onError(...);
  if (invalidRules.length > 0) return onError(...);
  if (confirmPassword !== newPassword) return onError(...);

  onSubmit(currentPassword, newPassword);
```

_Also there isn't any need for the `.length` on the strings. `!'' === true`_

PR #388, opened 2017-10-10
Dockless and positioned console

TomWija:


src/components/platform/Console.css
```diff
@@ -14,6 +14,21 @@
   transition-timing-function: var(--transition-function);
 }
 
+.ax-platform__console--right {
```
HHogg: Adding the two variations side by side (ie moving the defaults to an explicit `left` class) might be easier to understand for future editors. 

``
.ax-platform__console--left { ... }
.ax-platform__console--right { ... } 
```

File: src/components/platform/Canvas.js [hunk 2]
```diff
@@ -1,28 +1,34 @@
 import PropTypes from 'prop-types';
 import React, { Component } from 'react';
 import Base from '../base/Base';
+import classnames from 'classnames';
 import './Canvas.css';
 
 export default class Canvas extends Component {
   static propTypes = {
     children: PropTypes.node,
+    isDockless: PropTypes.bool,
```
HHogg: Maybe this is personal preference, but I tend to lean towards props being additive of behaviour. What do you think about reversing this prop to `hasDock` with a `defaultProp` of `true` (to avoid a breaking change)?

src/components/platform/Canvas.css
```diff
@@ -5,3 +5,7 @@
   transition-duration: var(--transition-time-slow);
   transition-timing-function: var(--transition-function);
 }
+
+.ax-platform__canvas--no-dock {
```
HHogg: Does the dock currently transition out as well? It's no big deal if it doesn't, but might be quite nice considering the canvas and console do transition between the states.

PR #392, opened 2017-10-18
feat(table): Totally overhauled API, style and behavior to match designs

benji6:
BREAKING CHANGE: Table API, style and behavior totally overhauled

Deployed here: https://table--axiom-rules.netlify.com/docs/components/table

screen shot 2017-10-18 at 11 28 10


src/components/table/Table.css
```diff
@@ -1,93 +1,93 @@
 :root {
-  --cmp-table-row-padding-small: calc(var(--spacing-grid-size) * 1);
-  --cmp-table-row-padding-medium: calc(var(--spacing-grid-size) * 2);
-  --cmp-table-row-padding-large: calc(var(--spacing-grid-size) * 4);
-
-  --cmp-table-column-padding-small: calc(var(--spacing-grid-size) * 1);
-  --cmp-table-column-padding-medium: calc(var(--spacing-grid-size) * 2);
-  --cmp-table-column-padding-large: calc(var(--spacing-grid-size) * 4);
+  --cmp-table-column-padding: var(--spacing-grid-size--x4);
```
pmsorhaindo: We're not allowing people to vary the padding size?

src/components/table/example/index.js
```diff
@@ -24,15 +26,19 @@ class TableExample extends Component {
...
     };
 
     const initialProps = {
       Table: {},
       TableBody: {},
       TableCell: {},
-      TableHead: {},
+      TableHeader: {},
+      TableHeaderLabel: {
+        onClick: () => console.log('TableHeaderLabel click'), // eslint-disable-line no-console
```
pmsorhaindo: I would argue consumers of the styleguide are unlikely to be looking at the console while interacting with examples. I believe alerts have been used to demonstrate callbacks in other places but I feel like in this circumstance no example is needed. SNABD, just my opinion.

pmsorhaindo:
Looks good to me just a couple NABD comments.

src/components/table/Table.css
```diff
@@ -1,93 +1,93 @@
...
-  width: 100%;
+.ax-table__header-label::after {
+  content: '';
+  position: absolute;
+  right: var(--cmp-table-column-padding);
+  bottom: 0;
+  left: var(--cmp-table-column-padding);
+  transform: scaleX(0);
+  border-bottom: var(--spacing-grid-size--x1) solid var(--color-ui-primary);
+  transition: transform var(--transition-time-base) var(--transition-function);
```
pmsorhaindo: Not really a fan of this animation.. It's not in the designs right? 🗡

benji6: This animation is sweet! We'll see what design say :D

src/components/table/example/index.js
benji6: Yeah I don't know what the answer is. I've seen alerts being used but I think they're fairly gross. To demonstrate the style change it needs to be set to a function, I thought a `console.log` would be better than an empty function :man_shrugging:

claudiodiliberto:
Hey @benji6 - great stuff with the tables! Looking forward to seeing how they will behave within the page. Please have a look at my feedback below.

**Selected Column**
The blue signifier of the selected col shouldn't stretch across the whole col. It should have the same width of the col header + the arrow (when the arrow it's visible). See pic below from Chris' Sketch file for reference.

**Width Animation on Selected Column**
I noticed that there's a transition on the blue signifier of the selected col (similar to the one used by Google on the input fields). We're not using any animation on selected tabs (e.g. in Audiences) so I was wondering if we should be consistent with that. Said that, we should also take into account that selecting a column on a table (or a tab) it's a pretty repetitive task while looking at data... Btw, something to discuss with Chris I guess...

**Position of the arrow**
The alignment of the arrow should be opposite to the alignment of the data in the column (see pic below for reference). Playing with the controls I couldn't get a left aligned arrow with right aligned data.

**Hover state on table headers**
I couldn't see any hover state on sortable table headers. 

Hope I've spotted everything...  as playing with all the property controls in the Axiom page it's not easy... It would be also cool if the preview at the top can have the sortable cols functionality enabled by default ;p

Hope this helps ;)
Claudio

table-from-chris-cketch-file

src/components/table/Table.css
```diff
@@ -1,93 +1,113 @@
...
+.ax-table__header-label--selected .ax-table__header-button {
+  color: var(--color-context-text);
+}
+
+.ax-table__header-label--selected .ax-table__header-button:hover {
+  color: var(--color-ui-primary--light);
+}
+
+.ax-table__header-label--shrink {
+  width: 1%;
```
HHogg: A shrunk cell might need to be forgiving on white spaces, do you think this class should have nowrap?

src/components/table/Table.css
```diff
@@ -1,93 +1,113 @@
...
+}
+
+.ax-table__header-label-underline::after {
+  content: '';
+  position: absolute;
+  right: 0;
+  bottom: 0;
+  left: 0;
+  transform: scaleX(0) translateY(var(--spacing-grid-size--x2));
```
HHogg: You could use `translateY(-100%)` here. Benefit of this is if in future the border width is changed this wouldn't need updating as well.

src/components/table/Table.css
```diff
@@ -1,93 +1,113 @@
...
+  border-bottom: var(--spacing-grid-size--x1) solid var(--color-ui-primary);
+  transition: transform var(--transition-time-base) var(--transition-function);
+}
+
+.ax-table__header-label-underline--selected::after {
+  transform: scaleX(1) translateY(var(--spacing-grid-size--x2));
```
HHogg: Same here.

HHogg: Also might need to check this animation over with Chris, just to get his input on it. (I like it, might be a bit nicer with an transform-origin on the same side as the text alignment though.)

src/components/table/TableHeaderLabel.js
```diff
@@ -0,0 +1,98 @@
...
+
+    let reactEl;
+
+    if (onClick) {
+      if (isSelected) {
+        reactEl = (
+          <button className="ax-table__header-button" onClick={ onClick }>
+            { [
+              children,
+              <TextIcon key="icon" name={ iconName } { ...textIconProps } />,
```
HHogg: Does this being removed and re-added on selection affect the column width? If it does it might cause the table column widths to be jumpy and reflow.

src/components/table/TableHeaderLabel.js
```diff
@@ -0,0 +1,98 @@
...
+    } else {
+      if (isSelected) {
+        reactEl = [
+          children,
+          <TextIcon key="icon" name={ iconName } { ...textIconProps } />,
+        ];
+      } else {
+        reactEl = children;
+      }
+    }
```
HHogg: This section is a bit difficult to see what ends up in the render. Could you achieve the same behaviour with a disabled button and a hidden icon? Something like ... 

```jsx
<button className="ax-table__header-button" disabled={ selected } onClick={ onClick }>
  { children }
  { sortDirection && (
    <TextIcon 
        name={ sortDirection === 'descending' ? 'triangle-down' : 'triangle-up' }
        spaceLeft={ textAlign === 'left' ? 'x2' : null }
        spaceLeft={ textAlign === 'right' ? 'x2' : null }
        style={ { opacity: isSelected ? 1 : 0 } } />
  ) }
</button>
```

* If no onClick is provided then it should do anything
* If there's no disabled styling then the onClick behaviour should be suppressed. 
* A hidden icon would also keep consistent column widths. 
* It avoids duplicating the `ax-table__header-button` class
* It's a little easier to see what should be in the render function because it's all inline.

src/components/table/TableHeaderLabel.js
```diff
@@ -0,0 +1,98 @@
...
+
+    return (
+      <Base { ...rest } Component="th" className={ className }>
+        <div className={ underlineClassName }>
```
HHogg: As `ax-table__header-label--selected` already exists you could avoid needing this className login here by just doing the below in the css.

```css
.ax-table__header-label--selected .ax-table__header-label-underline
```

src/components/table/Table.css
benji6: I have it on the `ax-table__header-label` class 👍

src/components/table/Table.css
benji6: That doesn't work :( My problem was getting that blue underline to appear on the table row line under the heading label, extending the full width of the text and maybe arrow too, but not extending the full width of the table cell. If I do no `translateY` the padding on the table cell means the line sits above the table-row border. If I set it to `100%` it just moves down by the line's height which is unrelated to the that padding. I can't think of a better way of getting this working except using that variable

src/components/table/Table.css
benji6: It's gone through design 👍  I experimented with the different origin, but thought it looked nicer this way. We should probably put it on the Menu too for consistency. As a side note I wanted the sort direction icon to fade in and out, but I couldn't think of a way to do that that didn't involve wrapping it in a `span` and transitioning the opacity that way - seemed way too complicated for a barely noticeable transition so I left it

src/components/table/TableHeaderLabel.js
benji6: Ah! Good catch - it probably would! But the table headers in the Axiom example are smaller than the data in the table so I never spotted it. What do you reckon the solution is? We could always render the Icon but just keep it invisible - but would that then look weird when the user hasn't clicked on it and it is not in a sorted state? Maybe it's better to have the width jump in that case 🤔

src/components/table/TableHeaderLabel.js
benji6: Interesting approach. My only concern is about whether this way is strictly speaking "semantic"... I mean when there is no `onClick` it's not really a button right? Especially since it will never be enabled. If we decide jumping column widths is worse than extra space in the table header labels then I'd be up for using this approach

src/components/table/TableHeaderLabel.js
benji6: Good shout, will update this 👍

src/components/table/Table.css
HHogg: Ahhh ok, I assumed it was just to move the bar down by it's height. Ok by me then 👍

PR #395, opened 2017-10-20
Validation pattern

HHogg:
Validation pattern as provided by UX works in the following way.

- No invalid indications are given until a form has been initially submitted. 
- Validation is then run in the following order:
  - All fields that are marked as required are indicated as invalid all together
    - Invalidations are then removed and re-added inline as they are changed. 
  - After all required fields have been satisfied, individual validations are run on inputs from top to bottom and only one at a time (and only on submit)
    - Any fields that are below an invalid field are shown as valid until the previous invalid field has been satisfied. 
    - Validations on the currently invalid field are removed and re-added inline (as they are satisfied)
    - The next invalid field doesn't get triggered until the form has been resubmitted.

In addition to the inline invalidation style, a single error hint is shown for the whole field and remains visible until the form has been resubmitted and that validation has been satisfied. 

#### How does this new Validation components accomplish this?

* Validation component provides the current error and a function to trigger the validation on submit.
* Validate component provides the inline validity state to apply invalid styling. 

*A stripped down example of using it with forms*
```jsx
<Form onSubmit={ () => {} }
  <TextInput error={ () => {} } patterns={ [/^.*[A-Z].*$/] } required value={ foo } />
</Form>
```

*A stripped down example of custom validation*

```jsx
const handleClick = (validate) => {
  const validationError = validate();

  this.setState({ validationError });

  if (!validationError) {
    console.log('Submit!');
  }
};

const { validationError } = this.state;

<Validation requiredError="All fields must be filled in!">
  { (validate) => (
    <Validate required value="somevalue">
      { (isValid) => isValid === true }
    </Validate>

    <Validate
        error={ () => 'Needs a capital letter!' }
        patterns={ [/^.*[A-Z].*$/] }
        required
        value="somevalue">
      { (isValid) => isValid === false }
    </Validate>

    { validationError }

    <Button onClick={ () => handleClick(validate) }>
      Validate
    </Button>
  ) }
</Validation>
```

> https://validation-pattern--hhogg-axiom.netlify.com/docs/composites/change-password


src/components/validate/Validate.js
```diff
@@ -0,0 +1,47 @@
...
+    this.setState({ index });
+  }
+
+  render() {
+    const { index } = this.state;
+    const { getValidity, hasMetPattern, hasMetRequired } = this.context;
+
+    return this.props.children(
+      getValidity(index),
+      () => hasMetRequired(index),
```
benji6: maybe this should just be a simple boolean

src/components/validate/Validate.js
```diff
@@ -0,0 +1,47 @@
...
+      (pattern) => hasMetPattern(index, pattern),
```
benji6: could pass a Map here

src/components/validate/Validator.js
```diff
@@ -0,0 +1,136 @@
+import PropTypes from 'prop-types';
+import { Component } from 'react';
+
+const getRequiredState = ({ required, value }) =>
+  !required || (value !== '' && value !== undefined && value !== null);
```
benji6: could rewrite as `!required || Boolean(value)`. Do you think it's worth guarding against whitespace too like `!required || Boolean(value && value.trim())`? But also - are we dealing with checkboxes here too? In which case can `value` be a boolean? It's set to `PropTypes.any` in the `Validate` component

src/components/validate/Validator.js
```diff
@@ -0,0 +1,136 @@
...
+  getOverallRequiredValidity() {
+    return this._validationGetters.every((_, index) => this.hasMetRequired(index));
+  }
+
+  getOverallValidity() {
+    return this.getOverallRequiredValidity() && this.getFirstInvalidIndex() === -1;
+  }
+
+  hasMetPattern(index, metPattern) {
+    return !this.getInvalidations(index).find((pattern) => pattern === metPattern);
```
benji6: `find` doesn't work in IE

src/components/validate/Validator.js
```diff
@@ -0,0 +1,136 @@
...
+
+  registerValidate(validationGetter) {
+    this._validationGetters.push(validationGetter);
+    this.setState({ invalidationIndex: this.getFirstInvalidIndex() });
+    return this._validationGetters.length - 1;
+  }
+
+  getFirstInvalidIndex() {
+    return this.getOverallRequiredValidity()
+      ? this.getAllInvalidations().findIndex((invalidation) => invalidation.length)
```
benji6: `findIndex` doesn't work in IE

src/components/validate/Validator.js
HHogg: I've tried to keep the components decoupled to form validation, so they can in theory be used for DataPickers, DatePickers, Dropdowns etc so yes they could be used with checkboxes. 

Using `Boolean(value)` would fail for checkboxes, so I think using `(value !== '' && value !== undefined && value !== null)` is correct and would work for checkbox validation too? 

I don't see any harm in trimming a string, and would save it being repeated on every required state. I'll change to ...

```
!required || (
  value !== undefined && 
  value !== null 
  (typeof value !== 'string' || string.trim() !== '')
);

File: src/composites/form-inputs/ConfirmPasswordInput.js [hunk 6]
```diff
@@ -0,0 +1,38 @@
...
+          value={ value }>
+        { (valid) =>
+          <TextInput { ...rest }
+              invalid={ !valid }
+              label={ t(axiomLanguage, 'confirm-new-password') }
+              space="x6"
+              type="password"
+              value={ value } />
+        }
+      </Validate>
```
thatguynamedandy: I'm not entirely sure how I feel about the API, in my head I imagined the consuming application could simply put the properties on the form field, and the containing form/fieldset could provide the context. But I hadn't thought about the actual implementation, which I guess is more complex. I don't feel too strongly about this, but this will become a big part of Axiom, so thought I would double check it.
```
<TextInput { ...rest }
  invalid={ !valid }
  label={ t(axiomLanguage, 'confirm-new-password') }
  space="x6"
  type="password"
  value={ value }
  error={ () => t(axiomLanguage, 'confirmation-doesnt-match') }
  patterns={ [(value) => value === newPasswordValue] }
  required />
```

src/components/validate/Validator.js
```diff
@@ -0,0 +1,136 @@
...
+  static childContextTypes = {
+    getValidity: PropTypes.func.isRequired,
+    hasMetPattern: PropTypes.func.isRequired,
+    hasMetRequired: PropTypes.func.isRequired,
+    registerValidate: PropTypes.func.isRequired,
+  };
+
+  constructor(props) {
+    super(props);
+    this._validationGetters = [];
```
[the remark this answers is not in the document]
benji6: TypeScript has true private members btw

src/components/validate/Validator.js
```diff
@@ -0,0 +1,136 @@
...
+
+export default class Validator extends Component {
+  static propTypes = {
+    children: PropTypes.func.isRequired,
+    requiredError: PropTypes.string,
+  };
+
+  static childContextTypes = {
+    getValidity: PropTypes.func.isRequired,
+    hasMetPattern: PropTypes.func.isRequired,
```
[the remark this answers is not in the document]
benji6: `checkPatternMet`?

src/components/validate/Validator.js
benji6: Good point about checkbox - that code looks good 👍

src/components/validate/Validator.js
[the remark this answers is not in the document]
HHogg: @thatguynamedandy I see what you mean... 😧

src/components/validate/Validator.js
```diff
@@ -0,0 +1,136 @@
...
+      hasMetPattern: this.hasMetPattern,
+      hasMetRequired: this.hasMetRequired,
+      registerValidate: this.registerValidate,
+    };
+  }
+
+  registerValidate(validationGetter) {
+    this._validationGetters.push(validationGetter);
+    this.setState({ invalidationIndex: this.getFirstInvalidIndex() });
+    return this._validationGetters.length - 1;
```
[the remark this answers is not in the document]
HHogg: The `this.getFirstInvalidIndex()` depends on validations being available in the array on the line before.

src/composites/form-inputs/ConfirmPasswordInput.js
HHogg: The main reason I've not baked it into the form components is because some designs we are working against require validation hints and styling which are outside or siblings of the input components. When this happens the validations can no longer happen inside the form components. 

It would also require form components and any custom validation component to replicate the setup that happens inside the Validate component.

HHogg:
@thatguynamedandy @benji6 I do agree with both your concerns that an API with the validation happening transparently inside form components would feel the more natural and friendly to the developer, it just doesn't allow us to accomplish some of the design we have been given where validation hints are presented as a sibling to the inputs. 

This approach I agree does have a slightly steeper learning curve, but it _will_ result in much less boilerplate code to set up custom validation components (with the use of context) and it's more flexible for the use outside of forms and doesn't impose any requirements on DOM structure (like having the form element high enough to wrap wherever the errors are being places).

PR #397, opened 2017-11-01
feat(Dropdown): Add onRequestOpen and onRequestClose optional props

benji6:


src/components/dropdown/Dropdown.js
```diff
@@ -73,7 +90,7 @@ export default class Dropdown extends Component {
       <Position
           { ...rest }
           isVisible={ isVisible }
-          onMaskClick={ () => this.close() }
+          onMaskClick={ this.close }
```
thatguynamedandy: Super minor, but I think the previous convention may have been deliberate, in the altered approach `event` is passed to close but not used in anyway.

benji6: yes - very good point - didn't realise `onMaskClick` was invoked with the event object, will change it back

benji6:
How do you mean @HHogg? Making the changes on `Position` instead won't allow us to hook into these events in the `Dropdown`.

benji6:
Oh right - I just realised it's all spread down... Then there's the question of the docs - I guess I should update the `PropTypes` for every component that uses `Position` then

benji6:
@HHogg doing this with the `Position` component is a lot more complicated when it comes to the `onRequestOpen` callback. It would involve looking for changes in the props. I feel like it would be simpler just to implement it like I have in this PR for any components that need it. Given that all other components will require `PropTypes`/docs writing I'm tempted to leave the implementation as it is.

PR #400, opened 2017-11-03
feat(Progress): Added ProgressButton component

HHogg:
progressbutton


src/components/progress/ProgressFinite.js
```diff
@@ -11,6 +11,8 @@ export default class ProgressFinite extends Component {
     percent: PropTypes.number.isRequired,
     /** Size of the indicator */
     size: PropTypes.oneOf(['small', 'medium', 'large']),
+    /** Size in REM units */
+    sizeRem: PropTypes.string,
   };
```
HHogg: With the global renaming task, this will eventually become the default size prop to use as we move towards our strict & flexible model.

PR #409, opened 2017-11-20
feat(Cloak): Added base property and alias components for consistent …

HHogg:
…cloaking ability

I've seen multiple places across every project using Axiom with the behaviour of hiding an element like a button inside a row and then hovering over the row reveals it (with transition) or when a menu is open on that button remaining visible. 

At the moment it's done by wrapping in an element, with a class on the element to be hidden and a class on the container to do the showing... but sometimes the transitions are being left off and not meeting the design requirements. 

This adds the behaviour to the Base class so it can be used like ... 

```jsx
<Card cloakContainer>
  <Grid>
    <GridCell>...</GridCell>
    <GridCell cloak={ !isOpen } shrink>
      <IconButton 
        name="ellipsis"
        onClick={ () => this.setState(({ isOpen }) => ({ isOpen: !isOpen })) } />
    </GridCell>
  </Grid>
</Card>
```

This will hide the button when `isOpen` is `false` and will show it when the `<Card>` is `:hover`ed and/or when `isOpen` is `true`.

I've also added some alias components `<Cloak invisible={ true }>` and `<CloakContainer>` where adding the base properties might interfere with other styles on the same element.

> [Netlify Link](https://cloak--hhogg-axiom.netlify.com/docs/components/cloak)

### DIFF
--- a/src/components/base/Base.css
+++ b/src/components/base/Base.css
@@ -128,3 +128,17 @@
   position: sticky;
   z-index: var(--z-index-sticky);
 }
+
+.ax-cloak {
+  opacity: 0;
+  transition-property: opacity;
+  transition-duration: var(--transition-time-base);
+  transition-timing-function: var(--transition-function);
+  pointer-events: none;
+
+  @nest .ax-cloak__container:hover &,
+  &.ax-cloak--visible {
+    opacity: 1;
+    pointer-events: all;
+  }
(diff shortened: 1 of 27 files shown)


thatguynamedandy:
Trying to think of something better than `cloakContainer` but failing.

PR #410, opened 2017-11-20
feat(HorizontalRule): add a styled horizontal rule component

pmsorhaindo:


src/components/horizontal-rule/HorizontalRule.js
```diff
@@ -0,0 +1,11 @@
+import React, { Component } from 'react';
+import './HorizontalRule.css';
+
+export default class HorizontalRule extends Component {
+  render() {
+    return (
+      <hr className="ax-horizontal-rule" />
```
HHogg: If you use `<Base Component="hr">` here instead, it will allow users to get the `space` property, which I can imagine will be really useful.

src/components/horizontal-rule/example/index.js
```diff
@@ -0,0 +1,15 @@
+import React, { Component } from 'react';
+import { ExampleConfig } from 'style-guide';
+import HorizontalRule from '../HorizontalRule';
+
+export class HorizontalRuleExample extends Component {
+  render() {
+    return (
+      <ExampleConfig propTypes={ {} }>
```
HHogg: Even though there's no props to configure, still set this up like the other just so in future if someone adds a property, the docs will automatically update.

```js
const propTypes = {
  HorizontalRule: components.HorizontalRule,
};

<ExampleConfig propTypes={ propTypes }>...</ExampleConfig>
```

src/components/horizontal-rule/HorizontalRule.css
```diff
@@ -0,0 +1,4 @@
+.ax-horizontal-rule {
+  width: 100%;
+  color: var(--color-context-border);
```
HHogg: I think by default just changing the colour will leave that double border effect. It's probably best to use a border for this. That was we can set the width as well. This should probably be something like ... 

```css
border-top: 0;
border-right: 0;
border-left: 0;
border-bottom: 0.0625rem solid var(--color-context-border);
```

PR #411, opened 2017-11-21
 fix(Table): add empty state table header border

pmsorhaindo:


src/components/table/Table.css
```diff
@@ -68,12 +68,17 @@
 
 .ax-table__header-label {
   padding: var(--spacing-grid-size--x2) var(--spacing-grid-size--x4);
+  border-top: var(--component-border-width-medium) solid var(--color-context-border);
```
HHogg: I don't think the top border should be something on the table. Looking at the designs ...
screen shot 2017-11-21 at 12 59 50
... It's valid to have a table without that top border. 

I think I know situation you are adding it for but that border there seems a result of dividing the content before it... which might be better suited to that HorizontalRule component. 

However the bottom border is legit 😉

PR #415, opened 2017-11-23
fix(Colors): Rename color variables

HHogg:


HHogg:
I forgot to add netlify link => https://rename-colors--hhogg-axiom.netlify.com/

PR #416, opened 2017-11-24
docs(PostCSS): Update create create app docs to correct css var file

lpoulter:
Docs pointed at old css vars file.


HHogg:
Thanks @lpoulter! Minor we probably don't want to publish a new release for this. Would you mind changing the scope from `fix(...)` to `docs(PostCSS): `?

PR #417, opened 2017-11-27
feat(TextInputButton): add a joined text input button

pmsorhaindo:
https://librarian-chimpanzee-34483.netlify.com/

### DIFF
--- a/src/components/form/TextInput.css
+++ b/src/components/form/TextInput.css
@@ -22,17 +22,17 @@
   order: 1;
 }
 
-.ax-input__container--small .ax-input {
+.ax-input__icon-container--small .ax-input {
   padding: var(--component-padding-vertical-small) var(--component-padding-horizontal-small);
   font-size: var(--font-size-small);
 }
 
-.ax-input__container--medium .ax-input {
+.ax-input__icon-container--medium .ax-input {
   padding: var(--component-padding-vertical-medium) var(--component-padding-horizontal-medium);
   font-size: var(--font-size-small);
 }
 
-.ax-input__container--large .ax-input {
+.ax-input__icon-container--large .ax-input {
   padding: var(--component-padding-vertical-large) var(--component-padding-horizontal-large);
   font-size: var(--font-size-large);
 }
@@ -45,7 +45,7 @@
   animation-fill-mode: forwards;
 }
 
-.ax-input__container {
+.ax-input__icon-container {
   display: flex;
   position: relative;
   flex: 1 1 auto;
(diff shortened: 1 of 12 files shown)


src/components/form/TextInputButton.js
```diff
@@ -0,0 +1,49 @@
...
+  };
+
+  getChildContext() {
+    return {
+      joined: true,
+    };
+  }
+
+  static defaultProps = {
+    size: 'medium',
```
HHogg: This size can just be passed down from the TextInput as a small TextInput will always have a small Button.

src/components/form/TextInputButton.js
```diff
@@ -0,0 +1,49 @@
...
+    style: 'primary',
```
HHogg: I don't think you need to specify a default here as the button already has a default.

src/components/form/TextInputButton.js
```diff
@@ -0,0 +1,49 @@
...
+
+export const TextInputButtonRef = 'TextInputButton';
+
+export default class TextInputButton extends Component {
+  static propTypes = {
+    children: PropTypes.node.isRequired,
+   /** Size of the input field */
+    size: PropTypes.oneOf(['small', 'medium', 'large']),
+   /** Style of the Button, that affects it's coloring and sizing */
+    style: PropTypes.oneOf([
```
HHogg: I don't think there's any need to duplicate the Buttons propTypes, if the incorrect propType is given it will trickle down and be caught by Button.

src/components/form/TextInput.js
```diff
@@ -2,14 +2,15 @@ import PropTypes from 'prop-types';
...
+import { TextInputButtonRef } from './TextInputButton';
 import TextGroup from './TextGroup';
 import Validate from '../validation/Validate';
 import findComponent from '../../utils/findComponent';
 import './TextInput.css';
 
 export default class TextInput extends Component {
   static propTypes = {
-    /** Optional TextInputIcon, all other children are ignored */
+    /** Optional TextInputIcon or TextInputButton all other children are ignored */
```
HHogg: It'll be safer to allow a combination of both TextInputIcon and TextInputButton. It shouldn't be too much extra work to do ([like it was before](https://github.com/BrandwatchLtd/axiom/blob/68892339e46f88569226fd9ba08f72589f4d4125/src/components/form/TextInput.js#L88))

src/components/form/TextInput.css
```diff
@@ -109,3 +113,19 @@
...
+  flex: 1 0 auto;
+
+  &:not(:first-child) {
+    border-top-right-radius: 0;
+    border-bottom-right-radius: 0;
+  }
+}
+
+.ax-input__button {
+  & .ax-button--joined {
```
HHogg: This class belongs to Buttons, redefining it here in TextInputs will lead to some tricky maintenance in the future.

src/components/form/TextInput.css
```diff
@@ -109,3 +113,19 @@
     color: transparent;
   }
 }
+
+.ax-input__container--with-button {
```
HHogg:
```css
.ax-input__container--with-button
.ax-input__group--with-button
```

I think it might be possible to put the button next to the input with a lot less CSS than is happening here. Take a look at the [template in the component](https://github.com/BrandwatchLtd/axiom/blob/68892339e46f88569226fd9ba08f72589f4d4125/src/components/form/TextInput.js#L68) before we removed this before, it was possible to do it with no CSS.

src/components/form/TextInput.js
pmsorhaindo: It's not an exclusive or, I can add an example and change this to 'and' to make this clearer.

PR #418, opened 2017-11-30
 feat(Cards): Added new Card components and updated style of Cards

HHogg:
[>>> Netlify Link <<<](https://rename-cards--hhogg-axiom.netlify.com/)


src/components/card/Card.js
```diff
@@ -2,26 +2,81 @@ import PropTypes from 'prop-types';
...
+import './Card.css';
+
+const cardListStyleProps = {
+  seamless: {
+    border: false,
+    color: 'default',
+    elevation: 'x0',
+    space: 'x0',
+  },
+  separate :{
```
thatguynamedandy: OCD `seamless: {` and `separate :{`

src/components/card/Card.js
```diff
@@ -2,26 +2,81 @@ import PropTypes from 'prop-types';
 import React, { Component } from 'react';
 import classnames from 'classnames';
 import Base from '../base/Base';
+import './Card.css';
+
+const cardListStyleProps = {
+  seamless: {
+    border: false,
+    color: 'default',
```
thatguynamedandy: Minor again but there maybe isn't a need to specify `color` here as it's always the same.

src/components/card/Card.js
```diff
@@ -2,26 +2,81 @@ import PropTypes from 'prop-types';
...
+  };
+
+  static defaultProps = {
+    color: 'default',
+    elevation: 'x0',
+    size: 'medium',
+  };
+
+  static contextTypes = {
+    cardListStyle: PropTypes.string,
```
thatguynamedandy: can this be `oneOf(['seamless', 'separate'])`?

src/components/card/Card.js
HHogg: This is to enforce the style so it can't be overridden when a Card is used within a list.

PR #425, opened 2017-12-12
fix(Platform): Allow DockItemLink to have component passed in to easil…

HHogg:
…ly work with ReactRouter Links


HHogg:
Good point @pmsorhaindo ... let's keep our fingers crossed 🤞

PR #426, opened 2017-12-12
docs(readme): Fix broken link to IE support docs and use casing that …

benji6:
…is consistent with the other docs links

Three guesses why I've been cribbing IE support! 😂


pmsorhaindo:
If it's why I think, I'm excited!

PR #436, opened 2017-12-20
fix(Design Alignment): Removed StatCards module as it's no longer a v…

HHogg:
…alid design pattern

This should have been included in v31, but it was missed.


HHogg:
@thatguynamedandy No, it's marked as a breaking change but should have been included within the design alignment branch.

PR #437, opened 2017-12-20
feat(Platform): Added new Platform components for console navigation and headers and changed API of Platform to better support multiple Consoles and a collapsible Console.

HHogg:
feat(Platform): Added new Platform components for console navigation and headers and changed API of Platform to better support multiple Consoles and a collapsible Console.

- CanvasHeader
- ConsoleHeader
- ConsoleMenuHeader
- ConsoleMenuHeaderTitle
- ConsoleMenu
- ConsoleMenuItem
- PlatformTitle

BREAKING CHANGE:

`<Platform>` component now takes these properties only
- `onConsoleClose` - A function that enables the collapsing of the Console and is called when the collapse handler is clicked.
- `openConsolePosition` - The position of the Console that is open (left or right)
- `openConsoleWidth` - The width of the Console that is open.

`<Console>` component now takes these properties only
- `position` - If the Console is positioned on the left or right.
- `width` - The width of the Console.

`<Canvas>` now takes no properties.


src/components/platform/ConsoleMenu.js
```diff
@@ -0,0 +1,13 @@
+import React, { Component } from 'react';
+import Base from '../base/Base';
+import './ConsoleMenu.css';
+
+export default class ConsoleMenu extends Component {
+  render() {
+    return (
+      <Base { ...this.props }
+          Component="ul"
+          className="ax-platform__console-menu" />
```
thatguynamedandy: Minor one but was just wondering whether in this scenario `ax-platform__console-menu` should actually be `ax-platform-console-menu`, because it declares it's own CSS file does that make it a Block level element?

src/components/platform/Platform.js
```diff
@@ -1,31 +1,49 @@
...
 import './Platform.css';
 
 export default class Platform extends Component {
   static propTypes = {
     children: PropTypes.node.isRequired,
-    consoleWidth: PropTypes.string.isRequired,
+    hasDock: PropTypes.bool,
+    onConsoleClose: PropTypes.func,
+    openConsolePosition: PropTypes.oneOf(['left', 'right']),
+    openConsoleWidth: PropTypes.string,
```
thatguynamedandy: Could this be a number?

HHogg:
Netlify -> https://platform-menu--hhogg-axiom.netlify.com/

src/components/platform/Canvas.js
```diff
@@ -1,39 +1,29 @@
...
-    consoleWidth: PropTypes.string.isRequired,
+    platformOpenConsolePosition: PropTypes.string,
+    platformOpenConsoleWidth: PropTypes.string,
   };
 
   render() {
-    const { children, hasDock, isShifted, shiftDirection, ...rest } = this.props;
-    const { consoleWidth } = this.context;
+    const { children, ...rest } = this.props;
+    const { platformOpenConsolePosition, platformOpenConsoleWidth } = this.context;
```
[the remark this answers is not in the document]
HHogg: Because it's context and likely to the top level component, I thought prefixing it with `platform` would be safer for the child components but maybe that doesn't actually matter as the level at which it's needed is quite shallow... yeah I'm happy to change it to something shorter 👍

src/components/platform/Platform.js
HHogg: Not really as it needs the unit to go with it, and feels a bit opinionated if we dictate the unit.

HHogg:
Thanks @thatguynamedandy, I agree about the toolbar, once it's gone through design review I'll make it a bit more subtle. We need to get round to putting the rendered examples into an iframe.

PR #442, opened 2017-12-22
feat(CardList): Added new 'divided' list style

HHogg:


src/components/card/Card.css
```diff
@@ -137,9 +137,24 @@
 .ax-card--clickable {
   cursor: pointer;
 
+  &:hover,
+  &:active,
+  &.ax-card--hover {
+    transform: translateZ(0); /* New stacking context for box-shadow */
```
thatguynamedandy: Let's keep an eye out for the half pixel blur issue.

PR #449, opened 2018-01-16
fix(Image): adds minimum and maximum width properties to Image

pmsorhaindo:
https://5a5dce62a1147706299bf091--librarian-chimpanzee-34483.netlify.com/docs/components/image


pmsorhaindo:
It's a my.brandwatch thing. We were passing down style on images and making use of max-width for products. Now Axiom's upgraded there the style is no longer passed down to the base image.

PR #450, opened 2018-01-16
fix(HorizontalRule): adds dotted horizontal rule style

pmsorhaindo:
https://5a5f29c9df9953478cb1ba35--librarian-chimpanzee-34483.netlify.com/docs/components/horizontal-rule?state=%7B%22HorizontalRule%22%3A%7B%22props%22%3A%7B%22lineStyle%22%3A%22dotted%22%7D%7D%7D

### DIFF
--- a/src/components/horizontal-rule/HorizontalRule.js
+++ b/src/components/horizontal-rule/HorizontalRule.js
@@ -1,13 +1,28 @@
+import PropTypes from 'prop-types';
+import classnames from 'classnames';
 import React, { Component } from 'react';
-import './HorizontalRule.css';
 import Base from '../base/Base';
+import './HorizontalRule.css';
 
 export default class HorizontalRule extends Component {
+  static propTypes = {
+    borderStyle: PropTypes.oneOf(['solid', 'dotted']),
+  };
+
+  static defaultProps = {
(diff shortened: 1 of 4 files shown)


src/components/horizontal-rule/HorizontalRule.js
```diff
@@ -1,13 +1,27 @@
...
+
   render() {
-    const { ...rest } = this.props;
+    const { style, ...rest } = this.props;
 
     return (
-      <Base space="x0" { ...rest } Component="hr" className="ax-horizontal-rule" />
+      <Base
+          space="x0"
+          style={ { borderStyle: style } }
```
HHogg: I'm not overly worries about blocking passed in inline styles, but it's not something that has to be added inline and could be done through a class. 

Inline styles is generally best used for dynamic values, like widths and heights.

thatguynamedandy:
I think the above would kind of block inline styles because the prop is called `style`.

HHogg:
@thatguynamedandy True, @pmsorhaindo what do you think about going with `borderStyle`?

pmsorhaindo:
I went with lineStyle, but will change to borderStyle.

PR #452, opened 2018-01-17
fix(TableHeaderLabel): adds ability to declare a column as unsortable…

pmsorhaindo:
… in order to remove cloaked sorting icon

https://5a5f74c0a6188f16f35500b2--librarian-chimpanzee-34483.netlify.com/docs/components/table?state=%7B%22TableHeaderLabel%22%3A%7B%22props%22%3A%7B%22sortable%22%3Afalse%7D%7D%7D

### DIFF
--- a/src/components/table/TableHeaderLabel.js
+++ b/src/components/table/TableHeaderLabel.js
@@ -16,12 +16,16 @@ export default class TableHeaderLabel extends Component {
     shrink: PropTypes.bool,
     /** Selects the column and sets the sort direction */
     sortDirection: PropTypes.oneOf(['ascending', 'descending']),
+    /** Indicates the column is sortable if false it will remove the cloaked icon
+    indicating sort direction */
+    sortable: PropTypes.bool,
     /** Set text-align */
     textAlign: PropTypes.oneOf(['left', 'right']),
   };
 
   static defaultProps = {
     textAlign: 'left',
+    sortable: true,
   }
 
   render() {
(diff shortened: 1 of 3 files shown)


src/components/table/TableHeaderLabel.test.js
```diff
@@ -36,4 +36,10 @@ describe('TableHeaderLabel', () => {
     const tree = component.toJSON();
     expect(tree).toMatchSnapshot();
   });
+
+  it('renders with sortable', () => {
+    const component = getComponent({ unsortable: false });
```
HHogg: Should this say `sortable`?

pmsorhaindo:
@HHogg I went with sortable to avoid `!unsortable && ...`. as I think it reads nicer. If you disagree with sortable being `true` by default I'm happy to change it.

src/components/table/TableHeaderLabel.js
```diff
@@ -16,12 +16,16 @@ export default class TableHeaderLabel extends Component {
...
+    /** Indicates the column is sortable if false it will remove the cloaked icon
+    indicating sort direction */
+    sortable: PropTypes.bool,
     /** Set text-align */
     textAlign: PropTypes.oneOf(['left', 'right']),
   };
 
   static defaultProps = {
     textAlign: 'left',
+    sortable: true,
```
HHogg: Even though the previous behaviour defaulted to adding the Icon whether sorting was on or not (which was incorrect), do you think it makes more sense if columns weren't sortable by default?

pmsorhaindo: I was thinking that most tabular data should be sortable, a column of checkboxes is rarer than a column of prices, names or dates. Happy to change it though.

PR #455, opened 2018-01-21
Select component

thatguynamedandy:


thatguynamedandy:
https://5a65af39a1147776deed6caf--andyp-axiom.netlify.com/docs/components/select

src/components/select/SelectInput.js
```diff
@@ -0,0 +1,50 @@
...
+    case 'Enter':
+      closeDropdown();
+      break;
+    case 'Tab':
+      closeDropdown();
+      break;
+    case 'ArrowDown':
+      openDropdown();
+      break;
+    case 'ArrowUp':
```
HHogg: I wonder if you need any `openDropdown` cases here if the default is to open dropdown on every key?

src/components/select/SelectOption.js
```diff
@@ -0,0 +1,49 @@
...
+    this.context.handleSelectOption(value);
+    if (onClick) onClick(event);
+  }
+
+  render() {
+    const { children, value, ...props } = this.props;
+    const { selectedOptionValue } = this.context;
+
+    const selected =
+      (Array.isArray(selectedOptionValue) && selectedOptionValue.includes(value))
```
HHogg: We can't use `includes` for the moment, so will have to go with `indexOf`

src/components/dropdown/DropdownContent.js
```diff
@@ -1,18 +1,138 @@
...
+
+  static contextTypes = {
+    closeDropdown: PropTypes.func,
+  };
+
+  static childContextTypes = {
+    addMenuItem: PropTypes.func.isRequired,
+    moveMenuItem: PropTypes.func.isRequired,
+    removeMenuItem: PropTypes.func.isRequired,
+    focusedIndex: PropTypes.any,
```
pmsorhaindo: It looks like this should be a number `-1` are their any cases when some other type is provided? I didn't see any.

src/components/dropdown/DropdownContent.js
```diff
@@ -1,18 +1,138 @@
...
+
+  addMenuItem(index) {
+    this.menuItems = [...this.menuItems, index];
+    this.setState(({ focusedIndex }, { focusOnOpen }) => ({
+      focusedIndex: (focusOnOpen && !focusedIndex) ? index : focusedIndex,
+    }));
+  }
+
+  removeMenuItem(index, focused) {
+    this.menuItems = this.menuItems.filter((_, p, arr) => p !== arr.indexOf(index));
```
HHogg: As it's not in state, you could probably avoid the `filter` and multiple `indexOfs` with a `splice`.

src/components/dropdown/DropdownContent.js
```diff
@@ -1,18 +1,138 @@
...
+  }
+
+  addMenuItem(index) {
+    this.menuItems = [...this.menuItems, index];
+    this.setState(({ focusedIndex }, { focusOnOpen }) => ({
+      focusedIndex: (focusOnOpen && !focusedIndex) ? index : focusedIndex,
+    }));
+  }
+
+  removeMenuItem(index, focused) {
```
HHogg: Could you avoid the `focused` argument, it looks like where it's being passed in, it's doing a check against the context, and that context comes from this component, so this component already has the information to perform that check.

src/components/select/example/index.js
```diff
@@ -0,0 +1,132 @@
+import PropTypes from 'prop-types';
+import React, { Component } from 'react';
+import flatten from 'lodash/flatten';
```
HHogg: `lodash` isn't a dependency and probably only works because it's a dependency of something else. To be on the safe side install the `.` package version... `yarn add lodash.flatten --dev`.

src/components/select/example/index.js
```diff
@@ -0,0 +1,132 @@
...
+      value: '',
+      selectedValue: null,
+    });
+  }
+
+  handleChange({ target }) {
+    this.setState(({ options }) => ({
+      value: target.value,
+      selectedValue:
+        (flatten(options).find(
```
HHogg: Similar to `inlcudes` we can't use `find` for now. (same again below)

src/components/select/example/index.js
```diff
@@ -0,0 +1,132 @@
...
+        onClear: value.length ? this.handleClear : undefined,
+        onSelect: this.handleSelect,
+        selectedValue,
+        value,
+      },
+      SelectOption: {},
+      SelectOptionGroup: {},
+    };
+
+    const optionFilter = ({ label }) => label.toLowerCase().startsWith(value.toLowerCase());
```
HHogg: Same again `startsWith` isn't available in IE.

src/components/select/example/index.js
```diff
@@ -0,0 +1,132 @@
...
+
+    return (
+      <ExampleConfig
+          initialProps={ initialProps }
+          propTypes={ propTypes }>
+
+        <Form onSubmit={ this.handleSubmit }>
+
+          Selected value: { selectedValue }
```
HHogg: Purely for example purposed, it might be nice to pass this into the `label` property on `Select` (it will just be passed down to the `TextInput` and be displayed correctly.

pmsorhaindo:
I think this breaks the current multi DropdownMenu functionality. It is selecting the first index of all DropdownMenu's inside a Dropdown when hovering over the first item.

src/components/context/Context.js
```diff
@@ -37,7 +56,15 @@ export default class Context extends Component {
...
 
     return (
-      <Base theme="day" { ...rest } className={ classes } style={ { width } }>
+      <Base
+          theme="day"
+          { ...rest }
+          className={ classes }
+          ref={ element => {
+            const node = ReactDOM.findDOMNode(element);
+            this.element = (node) ? node : this.element;
```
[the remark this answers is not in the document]
thatguynamedandy: Yes and I couldn't quite pinpoint the issue, it seemed like it was something to do with `Position`

src/components/dropdown/DropdownContent.js
thatguynamedandy: Good spot. 👍

src/components/dropdown/DropdownContent.js
thatguynamedandy: Good spot, although a string not a number 👍

src/components/context/Context.js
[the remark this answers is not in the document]
thatguynamedandy: The first time it renders fine, the second time it gets called with `null`

src/components/context/Context.js
```diff
@@ -24,6 +27,22 @@ export default class Context extends Component {
...
+  getChildContext() {
+    return {
+      isContextItemVisible: (target) => this.isVisible(target),
+    };
+  }
+
+  isVisible(target) {
+    const { top, height } = this.element.getBoundingClientRect();
+    const { top: targetTop, height: targetHeight } = target.getBoundingClientRect();
+    return targetTop + targetHeight < top + height && targetTop > top;
```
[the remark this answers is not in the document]
thatguynamedandy: I don't think this works, in the case of multiple `ContextMenu` the offset parent will be that menu and not the Context popup itself (which is what controls the scrolling)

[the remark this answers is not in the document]
HHogg: Ah good point 👍

src/components/context/ContextMenuItem.js
```diff
@@ -32,6 +51,8 @@ export default class ContextMenuItem extends Component {
       <Base
           Component="li"
           className="ax-context-menu__list-item"
+          ref={ element => this.element = (this.context.isContextItemVisible)
+              ? ReactDOM.findDOMNode(element) : null }
```
HHogg: Ah my previous comment, I was suggesting skipping the ref function altogether like ... 

```jsx
ref={ this.context.isContextItemVisible && (el) => this.element = ReactDOM.findDOMNode(el) }
```

thatguynamedandy:
No idea why Travis has stalled on this, I might rebase it all in the hope that it kicks it off.

PR #459, opened 2018-01-31
fix(TextInput): Use Base under the hood to allow base props to be applied. 

roburidge:
##### This PR updates the <TextInput> component to make it easier to customise.

By allowing use of the props from <Base> we get a bunch of really useful axiom utilities (like textCenter) to let us customise the component in a consistent way.

### DIFF
--- a/src/components/form/TextInput.js
+++ b/src/components/form/TextInput.js
@@ -1,6 +1,7 @@
 import PropTypes from 'prop-types';
 import React, { Component } from 'react';
 import classnames from 'classnames';
+import Base from '../base/Base';
 import TextInputIcon, { TextInputIconRef } from './TextInputIcon';
 import { TextInputButtonRef } from './TextInputButton';
 import TextGroup from './TextGroup';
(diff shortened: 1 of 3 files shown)


src/components/form/TextInput.js
```diff
@@ -111,6 +114,7 @@ export default class TextInput extends Component {
       'ax-input__icon-container--valid': valid,
       'ax-input__icon-container--invalid': invalid || isValid === false,
     });
+    const mergePropClasses = (className) => classnames(className, 'ax-input');
```
HHogg: Hey 👋 , so we probably don't want to allow custom classes to be passed in. Going by the conversation we were having the other day for aligning the text, just by using the Base component (like you've done below) will allow this like... 

```jsx
<TextInput textCenter />
```

roburidge: Yep, I can get what I need from this PR by just using the Base props which is great. The class name stuff was born out of a frustration I've been having with using Axiom that it locks the developer out of using the platforms api. I'd love to be able to use an axiom component but still have access to all the default web api attributes that I would usually have on a native <input> element.

This probably isn't the right place to get into the design principles of Axiom, but I'd love to get your thoughts on this some time (obv I 'm new to this project and I'm not aware of the context that design implementation decisions like this have been made in).

PR #462, opened 2018-02-14
Add boundariesElement prop to Select / Popper.js

jdelight:
Adds support for `boundariesElement` to be set on the Select and Dropdown components. Eventually applies to Position so anything using Position should support `boundariesElement` too.

Comments and feedback welcome :)

### DIFF
--- a/src/components/position/Position.js
+++ b/src/components/position/Position.js
@@ -17,6 +17,11 @@ export default class Position extends Component {
      * Children inside Position this should contain all of and
      * only PositionContent and PositionTarget!
      */
+    /** Element to use as the boundaries of the overlay */
+    boundariesElement: PropTypes.oneOfType([
+      PropTypes.string,
+      PropTypes.element,
+    ]),
     children: PropTypes.array.isRequired,
     /**
      * Adds control to conditionally enable or disable the positioning logic.
@@ -56,6 +61,7 @@ export default class Position extends Component {
     offset: 'middle',
     position: 'top',
     showArrow:  true,
+    boundariesElement: 'scrollParent',
   };
 
   constructor(props) {
@@ -93,7 +99,8 @@ export default class Position extends Component {
   }
 
   createPopper() {
-    const { flip, showArrow } = this.props;
+    const { boundariesElement, flip, showArrow } = this.props;
+
     const { placement } = this.state;
 
     return new popperJS(this._target, this._content, {
@@ -110,6 +117,9 @@ export default class Position extends Component {
         },
         inner: { enabled: false },
         offset: { enabled: false },
+        preventOverflow: {
+          boundariesElement,
+        },
       },
     });
   }
@@ -161,6 +171,7 @@ export default class Position extends Component {
       'offset',
       'position',
       'onPositionChange',
+      'boundariesElement',
     ]);
 
     return [
--- a/src/components/select/Select.js
+++ b/src/components/select/Select.js
@@ -13,6 +13,11 @@ export default class Select extends Component {
      * Children inside Select should contain all of and
      * only SelectOption and SelectOptionGroup!
      */
+    /** Element to use as the boundaries of the select overlay */
+    boundariesElement: PropTypes.oneOfType([
+      PropTypes.string,
+      PropTypes.element,
+    ]),
     children: PropTypes.node,
     /** Event that is fired when the input field is cleared */
     onClear: PropTypes.func,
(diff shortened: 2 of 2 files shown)


src/components/select/Select.js
```diff
@@ -13,6 +13,11 @@ export default class Select extends Component {
      * Children inside Select should contain all of and
      * only SelectOption and SelectOptionGroup!
      */
+    /** Boundaries used by the modifier */
+    boundariesElement: PropTypes.oneOfType([
```
thatguynamedandy: I think the prop types should actually go onto the `Position` component, and `Select` can remain unchanged, the spreading of the props will pass it down. Currently select is requesting that prop, but not directly using it. (the fact that it won't be documented is a separate problem). 

Anything other components that use `Position` can also benefit from the `boundariesElement`, but we wouldn't want to have to add that prop type to everything that uses `Position` (including future elements).

thatguynamedandy: (Adding propTypes to `Position` will also fix the build issue)

jdelight: Makes a lot of sense thanks @thatguynamedandy - I was wondering about exactly this 👍

jdelight:
@thatguynamedandy I've put the `boundariesElement` prop back in as it passes the build and is needed explicitly in the Select component.

Would be great to get your thoughts on it. 
Also Cc. @pmsorhaindo 
Thanks!

src/components/position/Position.js
```diff
@@ -17,6 +17,11 @@ export default class Position extends Component {
      * Children inside Position this should contain all of and
      * only PositionContent and PositionTarget!
      */
+    /** Boundaries used by the modifier */
```
thatguynamedandy: I wonder whether the prop types would ever need the `string` version, if at the moment we only need the element version I wonder about just using that instead, If we can restrict how much of Popper we expose I think  
I would be more comfortable. Popper is great, but keeping Axiom relatively free from dependencies would be a bonus.

On a smaller note, I think the docs could be slightly more descriptive.

```
/** Element to use as the boundaries of the select overlay
```

jdelight: Good call on the docs change 👍 

Yes we'll need to have the ability to use both the string and element. I've set the default to be the same as popper which is the string 'scrollParent' but actually might be worth not setting the boundariesElement at all and just use the popper default.

I think with something like `boundariesElement` which is essentially designed to ensure the UI is usable we want to be allowing devs to use popper.js options. For stylistic things like arrow direction I agree it makes sense to limit access to popper.

src/components/select/Select.js
```diff
@@ -13,6 +13,11 @@ export default class Select extends Component {
      * Children inside Select should contain all of and
      * only SelectOption and SelectOptionGroup!
      */
+    /** Element to use as the boundaries of the select overlay */
+    boundariesElement: PropTypes.oneOfType([
```
thatguynamedandy: I thought I wrote a comment about this, but I seemingly neglected to submit it, I wonder whether `boundariesElement` should just be restricted to just `element` as that is all that is required at the moment. My concern is that the more of popper we directly expose the more we become tightly coupled.

```

Comment (thatguynamedandy) [reply src/components/select/Select.js]: @jdelight I re-worded the above, sorry it wasn't a very well written comment.

Comment (jdelight) [reply src/components/select/Select.js]: Ah ok thanks for clarifying @thatguynamedandy - there are a few "special" strings that are supported by popper.js 'scrollParent' (default), 'viewport' and 'window' - so those are the string options (as well as a reference to a DOM node).

For example, there may well be times where the boundaries of the DOM element do not fully fit within the current viewport in which case it would be useful to use the "viewport" special string rather than the DOM node. There may also be times when components are full screen (using the full screen API) where we might always want the boundary to be the window. 

But yes popper.js is certainly complicated - lots of options. :)

Comment (thatguynamedandy) [reply src/components/select/Select.js]: hmm, I see the use case, my concern was around propagating popper options all the way up to the Axiom api, thus tightly coupling the two. If there is a solid use case then my one suggestion is that we at least use `oneOf` for the string prop type.

```js
boundariesElement: PropTypes.oneOfType([
  PropTypes.oneOf(['scrollParent', 'viewport', window']),
  PropTypes.element,
])
```

For anyone new to the component/Axiom it wouldn't be clear what string was expected from the docs/propTypes alone, and we don't directly state that it is popper handling the positioning.

jdelight:
Thanks for your help with this @pmsorhaindo @HHogg @pmsorhaindo 👍 really interesting discussion. We'll know more about our integration with Popper in future and will have a better idea of what we might need from Axiom. But can close this for now :)

src/components/select/Select.js
jdelight: I didn't know you could do that :) nice idea 👍

PR #465, opened 2018-03-01
feat(BarChart) Add BarChartLabel component prop

stefanbuck:
Wouldn't it be nice to be able to pass down a component for a BarChart label rather than just a boring string? This PR adds this functionally with a new prop `BarChartLabelComponent`.

screen shot 2018-03-01 at 10 44 35

### DIFF
--- a/src/charts/bar-chart/BarChart.js
+++ b/src/charts/bar-chart/BarChart.js
@@ -19,6 +19,8 @@ import './BarChart.css';
 
 export default class BarChart extends Component {
   static propTypes = {
+    /** The label that appears next to the bar */
+    BarChartLabelComponent: PropTypes.func,
     /**
      * Contextual component that appears when click on a dot.
      * It is provided with the `colors`, `label`, and `value` that has
@@ -93,6 +95,7 @@ export default class BarChart extends Component {
       chartKeyBenchmarkLabel,
       collapsedVisibleRowCount,
       ContextComponent,
+      BarChartLabelComponent,
       data,
       expandButtonSuffix,
       labelColumnWidth,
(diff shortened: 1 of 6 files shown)


src/charts/bar-chart/BarChart.js
```diff
@@ -24,6 +24,7 @@ export default class BarChart extends Component {
      * It is provided with the `colors`, `label`, and `value` that has
      * been clicked on, along with the `data` that was given for that row.
      */
+    BarChartLabelComponent: PropTypes.func,
```
stefanbuck: FYI: Will update inline docs soon

src/charts/bar-chart/BarChart.js
```diff
@@ -93,6 +95,7 @@ export default class BarChart extends Component {
       chartKeyBenchmarkLabel,
       collapsedVisibleRowCount,
       ContextComponent,
+      BarChartLabelComponent,
```
HHogg: ContextComponent is a ReactComponent because it renders quite a complex structure inside of a Dropdown. Would it be enough for this to be a function with the value given to it? Like a...

```jsx
barLabelFormatter={ (value) => `${value}%` }
```

stefanbuck: If I understood the concept of a formatter, this should return just a string, right? In our case we need to insert an arrow up or down based on the data. 

screen shot 2018-03-05 at 09 04 13

HHogg: Oooooo nice 👍

stefanbuck: Does this mean I should continue with a Component rather than a fromatter?

HHogg: Sorry, yeah Component looks good to me.

stefanbuck:
Closing this in favour of #473

PR #466, opened 2018-03-01
Lerna packages

HHogg:
The file moves and import changes cause this PR to look insanely big... notable changes. 

- The packages themselves ([README](https://github.com/HHogg/axiom/blob/lerna/README.md))
- [The migration guide](https://github.com/HHogg/axiom/blob/lerna/MIGRATION.md)
- [The build and deployment scripts](https://github.com/HHogg/axiom/tree/lerna/scripts)

_Commits are atomic and explain everything that has been done._

_Semantic releasing still happens but without the help of `semantic-release` it now uses `lerna --publish --conventional-comits`_


scripts/publish_site.sh
```diff
@@ -0,0 +1,37 @@
+#!/usr/bin/env bash
+set -e
+
+# We only want to deploy to gcp when on master
```
thatguynamedandy: You may need to throw a few extra commands into here to get it authorised to the correct project, but we could tweak this after.

thatguynamedandy:
Some minor UX improvements, could go in now or later..
1. The 'Axiom' header text would be nice if it was a link (just noticed the package header link does that too, a nice improvement would be a dropdown up there)
2. On the home page, instead of 'Awaiting Docs', maybe just have a 'Docs' link for the things that do have docs, it might read a bit more positively, at the moment it looks docs sparse, but things like the lint config don't require too much. (I also didn't pick up on the link at first)
3. The arrow on the table header next to `Package` bugs me as it's no clickable

(Looks awesome though, these are minor things).

src/utils/locales/index.test.js
```diff
@@ -1,52 +0,0 @@
...
-      Object.keys(defaultLocale).forEach((key) => {
-        it(`translates ${key}`, () => {
-          expect(!!locale.translations[key]).toBe(true);
-        });
-      });
-    });
-  });
-});
-
-describe('default - translate', () => {
```
thatguynamedandy: I wondered if a more succinct version of these tests would still be useful, just to prevent accidentally breaking changes for anyone tinkering with it, not a huge issue though.

thatguynamedandy:
This pattern still bugs me a little, as you have no way of knowing what is selected in that dropdown, made a bit more confusing by the `default` alongside. Future improvement?

screen shot 2018-03-01 at 14 48 45

thatguynamedandy:
Nice word cloud! I think the component container could do with a bit more padding to give the component a bit more space.
screen shot 2018-03-01 at 14 52 25

scripts/publish_site.sh
HHogg: Travis has some new ENVs that should keep it working the way it did before 🤞

src/utils/locales/index.test.js
HHogg: I've added these tests back in 👍 :)

HHogg:
@thatguynamedandy Good shout on the selected prop value UX, has bugged me for ages too... swapped it over to use your trusty Select component 😉

PR #467, opened 2018-03-05
chore(README): Add notice about nodejs version

stefanbuck:


.nvmrc
```diff
@@ -1 +0,0 @@
-v8.5.0
```
HHogg: I'm not opposed to removing this, just wondering why? 😄Looks like you can enforce the the latest version with with an [.nvmrc containing just "node"](https://github.com/creationix/nvm#nvmrc)

stefanbuck: Ohh nice find, I didn't know that. I'm happy to raise a new PR to fix this. Thanks for the hint

PR #470, opened 2018-03-05
 feat(SparkLine): Added new SparkLine component

HHogg:


packages/axiom-charts/src/Line/Line.js
```diff
@@ -0,0 +1,129 @@
+import PropTypes from 'prop-types';
+import React, { Component, cloneElement } from 'react';
+import classnames from 'classnames';
+import './Line.css';
+
+/** Browsers that don't support SVG verctor-effect ... IE11 */
```
thatguynamedandy: Slight typo in your IE11 SVG fun, `verctor`

packages/axiom-charts/src/Line/Line.js
```diff
@@ -0,0 +1,129 @@
...
+    /** Dash length */
+    dasharray: PropTypes.string,
+    /** Dash array */
+    dasharrayWithoutScalingStroke: PropTypes.string,
+    /**
+     * Data that determines the position of the line, if no lower
+     * or upper bounds are given then the lower and upper values
+     * will be taken from this data set.
+     */
+    data: PropTypes.array.isRequired,
```
thatguynamedandy: Is it possible to provide the shape of the data array for this?

packages/axiom-charts/src/Line/Line.js
```diff
@@ -0,0 +1,129 @@
...
+  static defaultProps = {
+    dasharray: '',
+    dasharrayWithoutScalingStroke: '',
+    width: '0.125rem',
+    widthWithoutScalingStroke: '0.015625rem',
+  };
+
+  render() {
+    const dataLower = Math.min(...this.props.data);
+    const dataUpper = Math.max(...this.props.data);
```
thatguynamedandy: Minor, but I wonder if you could save some computation by only doing the `Math.max` when it's needed, i.e. when `lower' and `upper` isn't set, it might also read a bit nicer.

I'm surprised you can get away with setting `lower = dataLower` again seeing as the initial assignment is a `const`.

packages/axiom-charts/src/Line/Line.js
```diff
@@ -0,0 +1,129 @@
...
+      lower = dataLower,
+      upper = dataUpper,
+      width,
+      widthWithoutScalingStroke,
+      ...rest
+    } = this.props;
+
+    const strokeDasharray = hasVectorEffectSupport ? dasharray : dasharrayWithoutScalingStroke;
+    const strokeWidth = hasVectorEffectSupport ? width : widthWithoutScalingStroke;
+    const finalLower = Math.min(lower, dataLower);
```
thatguynamedandy: If lower is set, are these the same things? `Math.min(lower, dataLower)`
Reads a little funny unless I missed a trick.

packages/axiom-charts/src/Line/Line.js
```diff
@@ -0,0 +1,129 @@
...
+      'moon-lagoon',
+      'terra-form',
+      'primeval-soup',
+      'new-horizon',
+      'blast-off',
+      'ground-control',
+      'luna-dust',
+    ]).isRequired,
+    /** Dash length */
+    dasharray: PropTypes.string,
```
thatguynamedandy: Would we want the dash length to be configurable? Or is it worth setting it in here (maybe even a default), and then let people just turn dashes on?

packages/axiom-charts/src/Line/LinePoint.js
```diff
@@ -0,0 +1,112 @@
...
+      onDropdownOpen,
+      size,
+      style,
+      value,
+      x,
+      y,
+      ...rest
+    } = this.props;
+
+    if (x === undefined || y === undefined) {
```
thatguynamedandy: Just wondering why `x` and `y` aren't required, is it to help it render in the demo? Would we ever use this outside of `Line`? not a big deal if that is the case.

packages/axiom-charts/src/ChartContext/ChartContext.js
```diff
@@ -0,0 +1,79 @@
...
+import {
+  Dropdown,
+  DropdownSource,
+  DropdownTarget,
+  Tooltip,
+  TooltipSource,
+  TooltipTarget,
+} from '@brandwatch/axiom-components';
+
+export default class ChartContext extends Component {
```
thatguynamedandy: I wasn't too sure about this component, it doesn't seem to follow existing patterns, it almost feels like 3 components wrapped into one. I'll reserve judgement until I have a moment to spin up the branch and see it working.

packages/axiom-charts/src/Line/Line.js
HHogg: The final `upper` and `lower` values have to take into account the data *and* if explicit upper and lower values are given. I wanted to do...

```js
const finalLower = Math.min(lower, ...this.props.data); // NaN when lower is undefined
const finalUpper = Math.max(upper, ...this.props.data);  // NaN when upper is undefined
```

Can you think of a nicer way of doing this? 

_`lower = dataLower` doesn't reassign, it's set to `dataLower` if `lower` is `undefined` 😉_

packages/axiom-charts/src/Line/Line.js
HHogg: Yeah same comment above that it has to take into account both the data and explicit lower. This check is only for when the two values are not the same.

packages/axiom-charts/src/Line/Line.js
HHogg: I did originally have it so it was a `style: PropTypes.oneOf(['solid', 'dashed'])` but when talking through with design, there is the possibility of the SparkLines benchmark line becoming a 1px dash. 

As this is the lowest level component, I thought it would be safer for future changes to allow it to be configurable. The dashes are enforced in the LineChart with a simple `dashed` property though.

packages/axiom-charts/src/Line/LinePoint.js
HHogg: It wasn't for the demo, but it's the Line component that passes the position down because it knows the who data set and container width... but I think this check isn't actually needed. If it's not used inside the Line or is not given an x and y then I guess it's just not being used correctly. 

Will remove 👍

packages/axiom-charts/src/ChartContext/ChartContext.js
HHogg: Yeah, between the BarChart, DotPlotChart and LineChart, each have the flexibility to show context on hover and/or click. All 3 had/will have the same logic ... 

```jsx
if (DropdownContext && TooltipContext) return wrappedComponent;
if (DropdownContext) return wrappedComponent;
if (TooltipContext) return wrappedComponent;
```

... except this `wrappedComponent` would be the same hefty template, when all it needed to do was just wrap children. This is a just a routing/wrapping component.

packages/axiom-charts/src/Line/Line.js
HHogg: Changed to `PropTypes.arrayOf(PropTypes.number).isRequired`

packages/axiom-charts/src/Line/Line.js
thatguynamedandy: good point, it's been so long since I wrote any code that I was getting `lower = dataLower` and `lower: dataLower` muddled up

PR #471, opened 2018-03-05
Fixes Charts retaining hover states

HHogg:
Fixes the issue with BarChart and DotPlotChart where hover states were not retained when the Dropdown was opened. 

_Required #470 to be merged first_


packages/axiom-charts/src/BarChart/BarChart.js
```diff
@@ -122,9 +139,11 @@ export default class BarChart extends Component {
             xAxisLabels={ xAxisLabels }
             zoomTo={ zoomValue }>
           { formattedData.map(({ values, label, benchmark }, index) =>
-            <ChartTableRow key={ isValidElement(label) ? index : label }>
+            <ChartTableRow
+                hover={ index === hoverIndex || index === clickedIndex }
```
thatguynamedandy: Do you need `hoverIndex` and `clickedIndex`, I was just wondering if you could get away with a `selectedIndex` instead.

If you can it might help with the readability of this..
`isHovered={ index === hoverIndex || index === clickedIndex }`

packages/axiom-charts/src/BarChart/BarChartBars.js
```diff
@@ -52,20 +57,26 @@ export default class BarChartBars extends Component {
...
-                  value={ value } />
+                  onDropdownClose={ onDropdownClose }
+                  onDropdownOpen={ () => onDropdownOpen(color) }
+                  value={ value }>
+                <Bar
+                    color={ color }
+                    isFaded={ isFaded }
+                    isHidden={ hideBars && isFaded }
+                    labelStrong={ isHovered }
+                    onMouseEnter={ onMouseEnter && (() => onMouseEnter(color)) }
```
thatguynamedandy: onMouseEnter is required, so can you simplify this?
`onMouseEnter={ () => onMouseEnter(color) }`

packages/axiom-charts/src/DotPlotChart/DotPlot.js
```diff
@@ -68,17 +73,23 @@ export default class DotPlot extends Component {
...
               key={ value }
               label={ label }
-              onMouseEnter={ () => onDotMouseEnter(colors) }
-              onMouseLeave={ onDotMouseLeave }
-              value={ value } />
+              onDropdownClose={ onDropdownClose }
+              onDropdownOpen={ () => onDropdownOpen(getDotColors(mouseOverRowIndex, mouseOverColors, rowIndex, colors) ) }
+              value={ value }>
+            <DotPlotDots
+                colors={ getDotColors(mouseOverRowIndex, mouseOverColors, rowIndex, colors) }
```
thatguynamedandy: There is a part of me that wanted to put these args into an array.
`const dropdownArgs = [mouseOverRowIndex, mouseOverColors, rowIndex, colors]`
and do this instead..
`getDotColors(...dropdownArgs)`

No biggie, and I know this is older code.

packages/axiom-charts/src/DotPlotChart/DotPlotChart.js
```diff
@@ -74,11 +74,27 @@ export default class DotPlotChart extends Component {
...
     super(props);
     this.state = {
+      clickedColors: [],
+      clickedRowIndex: -1,
       mouseOverColors: [],
       mouseOverRowIndex: -1,
     };
   }
 
+  handleDropdownOpen(rowIndex, colors) {
```
thatguynamedandy: Elsewhere we would maybe write it like..
```js
handleDropdownOpen(clickedColors, clickedRowIndex) {
  this.setState({
    clickedColors,
    clickedRowIndex
  });
}
```

packages/axiom-charts/src/BarChart/BarChart.js
HHogg: Yeah I'm not a big fan of this either, it was around the mouseleave removing the selected index when the dropdown was opened. I'll have a go to pin it down to a single property 👍

PR #473, opened 2018-03-05
feat(BarChart):  Flexible labels and absolute data support

stefanbuck:
This PR makes the BarChart component even more appealing with a few changes.

screen shot 2018-03-05 at 16 26 39

### Custom range

Before the BarChart was limited to values between 0% and 100%. With the new properties `lower` and `upper` you can specify the range like you want. If you don't any of these values, the component will automatically take the lowest and highest value from the passed data.

### Tweak bar label

Sometimes you want to show more than just a boring number in your BarChart. With the new property `barLabel` you can easily append units like a percentage sign or return a rich component and go wild.

## Breaking changes

I'm sorry, this PR contains a few breaking changes:

1. To retain the previous behaviour of the `BarChart` and `DotPlotChart` with values going from 0 up to 100, you have to pass `lower=0` and `upper=100` to not let the new behaviour kick-in. Otherwise your Chart will start from the lowest value up to the highest value.

2. In favour of the custom range, the `BarChart` and `DotPlotChart` does not add a percentage sign anymore. Use the `barLabel` or `dotPlotLabel` property to suffix your values: `barLabelFormatter={ value => `${value}%` }`

3. Also in the context of the custom range, we had to mark `xAxisLabels` as required.

4. The new custom bar labels required to change the positing of the labels. Before Flexbox was used to position the labels at the end of the bar chart. This was conflicting with the new behaviour. 

@HHogg Since this is my first feature PR to Axiom, please don't hesitate to spam my inbox 😉

### DIFF
--- a/packages/axiom-charts/src/Bar/Bar.css
+++ b/packages/axiom-charts/src/Bar/Bar.css
@@ -1,5 +1,6 @@
 .ax-bars__bar {
   display: flex;
+  position: relative;
   transition: opacity var(--transition-time-base) var(--transition-function);
 }
 
@@ -121,7 +122,23 @@
 .ax-bars--right,
 .ax-bars--left {
   & .ax-bars__bar-label {
-    flex: 0 0 var(--cmp-chart-label-width);
+    position: absolute;
+  }
+}
+
+.ax-bars--right {
+  & .ax-bars__bar-label {
+    left: 100%;
+    margin-right: calc(var(--cmp-chart-label-width) * -1);
+    padding-left: var(--cmp-chart-label-margin);
+  }
+}
+
+.ax-bars--left {
+  & .ax-bars__bar-label {
+    right: 100%;
+    margin-left: calc(var(--cmp-chart-label-width) * -1);
+    padding-right: var(--cmp-chart-label-margin);
   }
 }
 
@@ -143,19 +160,5 @@
   & .ax-bars__bar-label {
     margin-bottom: calc(var(--cmp-chart-label-height) * -1);
     padding-top: var(--cmp-chart-label-margin);
-  }
-}
-
-.ax-bars--right {
-  & .ax-bars__bar-label {
-    margin-right: calc(var(--cmp-chart-label-width) * -1);
-    padding-left: var(--cmp-chart-label-margin);
-  }
-}
-
-.ax-bars--left {
-  & .ax-bars__bar-label {
-    margin-left: calc(var(--cmp-chart-label-width) * -1);
-    padding-right: var(--cmp-chart-label-margin);
   }
 }
--- a/packages/axiom-charts/src/BarChart/BarChart.js
+++ b/packages/axiom-charts/src/BarChart/BarChart.js
@@ -26,6 +26,8 @@ export default class BarChart extends Component {
     DropdownContext: PropTypes.func,
     /** The title that appears along the xAxis */
     axisTitle: PropTypes.string,
+    /** Control the appearance of the bar label */
+    barLabel: PropTypes.func,
     /**
      * The key that is shown along the bottom of the axis. It is also used
      * to determine the order of stacked dots.
@@ -96,6 +98,7 @@ export default class BarChart extends Component {
 
     const {
       axisTitle,
+      barLabel,
       chartKey,
       chartKeyBenchmarkLabel,
       collapsedVisibleRowCount,
@@ -145,6 +148,7 @@ export default class BarChart extends Component {
               <ChartTableVisual>
                 <BarChartBars
                     DropdownContext={ DropdownContext }
+                    barLabel={ barLabel }
                     benchmark={ benchmark }
                     benchmarkHeight={ rowSpace }
                     data={ data[index] }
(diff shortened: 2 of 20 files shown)


packages/axiom-charts/src/Bar/Bar.css
```diff
@@ -121,7 +122,19 @@
...
+
+.ax-bars--right {
+  & .ax-bars__bar-label {
+    left: 100%;
+  }
+}
+
+.ax-bars--left {
+  & .ax-bars__bar-label {
+    right: 100%;
```
stefanbuck: I'm slightly worried that this change will break either another component or an app which is using Axiom. Do you have any best practice to follow?

packages/axiom-charts/src/BarChart/BarChart.js
```diff
@@ -88,29 +96,37 @@ export default class BarChart extends Component {
...
+
     const formattedData = formatData(chartKey, data);
-    const highestValue = getHighestValue(data);
-    const zoomValue = zoom
-      ? Math.max(zoomMax, Math.min(100, Math.ceil(highestValue / 10) * 10))
-      : 100;
+    const flatValues = flattenValues(formattedData);
+
+    const highestValue = Math.max(...flatValues);
+    const dLower = lower === undefined ? Math.min(...flatValues) : lower;
```
HHogg: One issue I found with normalising the LineChart bits is if lower/upper is given and the data contains a lower/higher value. So ideally the lower and upper value should also be included in the minimum and maximum calculation. 

[Reference](https://github.com/BrandwatchLtd/axiom/pull/472/files#diff-67bf5b6a7f0ec3face40c49f93bdf6abR92)

packages/axiom-charts/src/Bar/Bar.css
HHogg: I think it should be ok, but I'll double check some apps. But yes anything that is going to require changes we add `BREAKING CHANGE` into the commit message body. 

For example

```
feat(Component): Changed something to the API

BREAKING CHANGE: The component prop `some prop` has been changed and you need to do something.
```

The `BREAKING CHANGE` will trigger a `MAJOR.minor.patch` change when publishing... but I don't think here is a case for it. I'll double check though.

HHogg: Oh I see you already added BREAKING CHANGE to your commits so you know about it 😄 ignore me.

HHogg:
There's a little bit of spill over into the DotPlotCharts here also as they use the same layout and components (ChartTable). Do you know it if would be much effort to apply the changes to DotPlot and ChartTable too? 

As there's going to be a breaking change it might be nice to do it once for both charts, but if it's a major bit of work then not to worry ;)

packages/axiom-charts/src/BarChart/BarChart.js
stefanbuck: I address your comment in the following fixup commit dd96ffa0

stefanbuck:
Thanks for your review so far. I'll try to adapt the changes to `DotPlotCharts` and `ChartTable` as well.

packages/axiom-charts/src/BarChart/BarChart.js
```diff
@@ -86,31 +92,41 @@ export default class BarChart extends Component {
...
     const formattedData = formatData(chartKey, data);
-    const highestValue = getHighestValue(data);
-    const zoomValue = zoom
-      ? Math.max(zoomMax, Math.min(100, Math.ceil(highestValue / 10) * 10))
-      : 100;
+
+    const finalLower = Math.min(lower, dataLower);
+    const finalUpper = Math.max(upper, dataUpper);
+
+    const zoomValue = zoom ? Math.max(zoomMax, dataUpper) : undefined;
```
HHogg: Should this be using `finalUpper`?

packages/axiom-charts/src/BarChart/utils.test.js
```diff
@@ -106,7 +106,7 @@ describe('BarChart (utils)', () => {
     }]);
   });
 
-  it('it gets highest value', () => {
-    expect(getHighestValue(data)).toBe(100);
+  it('it gets a flatt list of all values', () => {
+    expect(flattenValues(data)).toMatchSnapshot();
```
HHogg: Not saying this is wrong but I've only seen snapshot tests for templates before, is this standard over a `toEqual`? 

```js
expect(flattenValues(data)).toEqual([100, 0, 90, 50, ...]);
```

packages/axiom-charts/src/DotPlotChart/DotPlotChart.js
```diff
@@ -113,10 +127,11 @@ export default class DotPlotChart extends Component {
...
     const formattedData = formatData(chartKey, data);
-    const highestValue = getHighestValue(data);
-    const zoomValue = zoom
-      ? Math.max(zoomMax, Math.min(100, Math.ceil(highestValue / 10) * 10))
-      : 100;
+
+    const finalLower = Math.min(lower, dataLower);
+    const finalUpper = Math.max(upper, dataUpper);
+
+    const zoomValue = zoom ? Math.max(zoomMax, dataUpper) : undefined;
```
HHogg: Same comment here about the `finalUpper`?

packages/axiom-charts/src/BarChart/BarChart.js
stefanbuck: Indeed, good catch 👍

packages/axiom-charts/src/BarChart/utils.test.js
stefanbuck: Yes, it is equivalent. I like to use snapshots for more complex data like this (although in that case it's just long)

HHogg: 👍

PR #476, opened 2018-03-07
fix(Dependecy): Added missing @brandwatch/axiom-automation-testing de…

HHogg:
…pendency

### DIFF
--- a/packages/axiom-components/package.json
+++ b/packages/axiom-components/package.json
@@ -8,6 +8,7 @@
     "access": "public"
   },
   "dependencies": {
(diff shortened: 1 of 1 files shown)


packages/axiom-components/package.json
```diff
@@ -8,6 +8,7 @@
     "access": "public"
   },
   "dependencies": {
+    "@brandwatch/axiom-automation-testing": "^0.0.3",
```
stefanbuck: Shouldn't this be a devDependencies?

HHogg: It's adding automations test ids/attributes which are also needed for live automation tests.

PR #478, opened 2018-03-09
fix(Card): Added CardImage dotted separator

HHogg:
screen shot 2018-03-09 at 09 26 20


pmsorhaindo:
Will we always have the dots? it looks odd to me.
But otherwise the PR is 👍

HHogg:
@pmsorhaindo Yeah it's always supposed to be there so there is a very subtle separator between the content and any image with a white background.

PR #479, opened 2018-03-12
feat(Package): Added new babel-plugin-axiom-imports (modified version…

HHogg:
… of old babel-plugin-axiom)

Thanks a lot to @stefanbuck for help porting and tweaking this over from the old plugin.

### DIFF
--- a/packages/babel-plugin-axiom-imports/src/babel-plugin-axiom-imports.js
+++ b/packages/babel-plugin-axiom-imports/src/babel-plugin-axiom-imports.js
@@ -1,0 +1,104 @@
+const resolveImport = require('./resolve-import');
+
+module.exports = ({ types }) => {
+  let axioms;
+  let specified;
+  let selectedAxioms;
+
+  const importAxiom = ({ pkg, name }, file) => {
+    if (!selectedAxioms[name]) {
+      selectedAxioms[name] = file.addImport(resolveImport(pkg, name), 'default');
+    }
+
+    return types.clone(selectedAxioms[name]);
+  };
+
+  const isSpecialTypes = (node) =>
+    types.isMemberExpression(node) ||
+    types.isProperty(node);
+
+  const hasBindingOfType = (scope, name, type) => scope.hasBinding(name) &&
+    scope.getBinding(name).path.type === type;
+
+  const matchesAxiom = (path, name) => axioms[name] && (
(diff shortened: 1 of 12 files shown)


packages/babel-plugin-axiom-imports/src/resolve-import.js
```diff
@@ -0,0 +1,45 @@
...
+
+const exportsList = supportedAxiomPackages.reduce((memo, packageName) => {
+  const exportList = fs.readFileSync(path.join(path.dirname(
+    Module._resolveFilename(packageName, Object.assign({}, new Module, {
+      'paths': Module._nodeModulePaths(process.cwd()),
+    }))
+  ), '../src/index.js'), 'utf8');
+
+  memo[packageName] = parse(exportList, { sourceType: 'module' }).program.body
+    .reduce((exprts, exprt) => Object.assign({}, exprts,
```
siggysamson: Did you call it `exprts` because `exports` is a reserved keyword? 😄

packages/babel-plugin-axiom-imports/src/babel-plugin-axiom-imports.js
```diff
@@ -0,0 +1,103 @@
+const resolveImport = require('./resolve-import');
+
+const isSpecialTypes = (types, node) => ['isMemberExpression', 'isProperty']
+  .filter((type) => types[type](node)).length > 0;
```
t-sauer: Nothing big, but that could also be done with `Array.prototype.some`

packages/babel-plugin-axiom-imports/src/resolve-import.js
HHogg: I don't think `exports` is actually a reserved word, but it did mess with my syntax highlighting :D I though tit would be friendlier to mine and possible others IDEs to change it, but I'm happy to add the `o` if you prefer :)

siggysamson: Nah, I don't mind 🙂

packages/babel-plugin-axiom-imports/src/babel-plugin-axiom-imports.js
HHogg: Yeah, I think we could drop the Array all together tbh. Will update, thanks!

PR #485, opened 2018-03-20
feat: adds an axis utils

tomru:
Now that the BarChart can display absolute values instead of just percent, I added this axis helper providing: 

* `axiUtilsgetAxisUpper([25, 41, ...])`
  Calculates an upper based on a array of values to be charted. This is still pretty dumb, as it'll always choose the next power of 10 based on the maximum value used.
* `axisUtils.getEquallyDistributedAxisLabels(upper, lower, tickCount = 5, labelFormatter)`
  Generates labels for equally distributed ticks between upper and lower.

### DIFF
--- a/packages/axiom-charts/src/index.js
+++ b/packages/axiom-charts/src/index.js
@@ -20,3 +20,4 @@ export { default as MirroredColumnChart } from './ColumnChart/MirroredColumnChar
 export { default as SparkLine } from './SparkLine/SparkLine';
 export { default as Word } from './WordCloud/Word';
 export { default as WordCloud } from './WordCloud/WordCloud';
+export { default as axisUtils } from './utils/axisUtils';
--- a/packages/axiom-charts/src/utils/__snapshots__/axisUtils.test.js.snap
+++ b/packages/axiom-charts/src/utils/__snapshots__/axisUtils.test.js.snap
@@ -1,0 +1,97 @@
+// Jest Snapshot v1, https://goo.gl/fbAQLP
+
+exports[`getEquallyDistributedAxisLabels creates equal labels when lower -1, upper 1, tickCount 7 1`] = `
+Array [
+  -1,
+  -0.6666666666666667,
+  -0.33333333333333337,
+  0,
+  0.33333333333333326,
+  0.6666666666666665,
+  1,
+]
+`;
+
+exports[`getEquallyDistributedAxisLabels creates equal labels when lower -10k, upper 10k, tickCount 11 1`] = `
+Array [
+  -10000,
+  -8000,
+  -6000,
+  -4000,
+  -2000,
+  0,
+  2000,
+  4000,
+  6000,
+  8000,
+  10000,
+]
+`;
+
+exports[`getEquallyDistributedAxisLabels creates equal labels when lower 0, upper 1, tickCount 11 1`] = `
+Array [
+  0,
+  0.1,
+  0.2,
+  0.30000000000000004,
+  0.4,
+  0.5,
+  0.6000000000000001,
+  0.7000000000000001,
+  0.8,
+  0.9,
+  1,
+]
+`;
+
+exports[`getEquallyDistributedAxisLabels creates equal labels when lower 0, upper 10k, tickCount 6 1`] = `
+Array [
+  0,
+  2000,
+  4000,
+  6000,
+  8000,
+  10000,
+]
+`;
+
+exports[`getEquallyDistributedAxisLabels creates equal labels when lower 1k, upper 10k, tickCount 6 1`] = `
+Array [
+  1000,
+  2800,
+  4600,
+  6400,
+  8200,
+  10000,
+]
+`;
+
+exports[`getEquallyDistributedAxisLabels creates equal labels when lower 70k, upper 1m, tickCount 6 1`] = `
+Array [
+  70000,
+  256000,
+  442000,
+  628000,
+  814000,
+  1000000,
+]
+`;
+
+exports[`getEquallyDistributedAxisLabels formats label when "labelFormatter" is provided 1`] = `
+Array [
+  "formatted 1000 unit",
+  "formatted 2800 unit",
+  "formatted 4600 unit",
+  "formatted 6400 unit",
+  "formatted 8200 unit",
+  "formatted 10000 unit",
+]
+`;
+
+exports[`getEquallyDistributedAxisLabels provides a minimum of 3 ticks if less is specified 1`] = `
+Array [
+  0,
+  5000,
+  10000,
+]
+`;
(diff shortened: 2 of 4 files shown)


packages/axiom-charts/src/utils/axisHelper.js
```diff
@@ -0,0 +1,38 @@
...
+const MIN_TICK_COUNT = 3;
+
+export function getEquallyDistributedAxisLabels({
+    lower,
+    upper,
+    tickCount = DEFAULT_TICK_COUNT,
+}) {
+  const finalTickCount = Math.max(MIN_TICK_COUNT, tickCount);
+  const tickSize = (upper - lower) / (finalTickCount - 1);
+  const ticks = Array.from(Array(finalTickCount).keys()).map((value, index) =>
```
HHogg: We've avoided any ES2015 functions because we've tried to keep Axiom supporting IE11 without requiring any polyfills.

packages/axiom-charts/src/utils/axisHelper.js
```diff
@@ -0,0 +1,38 @@
...
+      Math.floor(index * tickSize + lower)
+  );
+
+  return ticks.map(value => tinyNumber(value));
```
HHogg: I think `tinyNumber` assumes the formatting, but might not make sense for percentages? Maybe this could passed in as a formatting function or the raw values are returned?

packages/axiom-charts/src/index.js
```diff
@@ -20,3 +20,4 @@ export { default as MirroredColumnChart } from './ColumnChart/MirroredColumnChar
 export { default as SparkLine } from './SparkLine/SparkLine';
 export { default as Word } from './WordCloud/Word';
 export { default as WordCloud } from './WordCloud/WordCloud';
+export { default as axisHelper } from './utils/axisHelper';
```
HHogg: I can't see a `default` export inside `axisHelper`?

packages/axiom-charts/src/utils/axisHelper.js
```diff
@@ -0,0 +1,38 @@
...
+  }
+
+  return Math.floor(Math.log10(value));
+}
+
+export function getAxisUpper(values) {
+  const max = Math.max.apply(null, values);
+
+  if (max < 10) {
+    return 10;
```
HHogg: I don't think I understand this check? What if the data is ranged `0...1` (for example [0.012, 0.123, 0.354])

HHogg: I don't think I understand this check? What if the data is ranged `0...1` (for example [0.012, 0.123, 0.354])

tomru: was probably looking to much at our use case where there are always just integers. will change it to support digits as well

packages/axiom-charts/src/index.js
tomru: fixed now

packages/axiom-charts/src/utils/__snapshots__/axisHelper.test.js.snap
```diff
@@ -0,0 +1,97 @@
...
+  10000,
+]
+`;
+
+exports[`getEquallyDistributedAxisLabels creates equal labels when lower 0, upper 1, tickCount 11 1`] = `
+Array [
+  0,
+  0.1,
+  0.2,
+  0.30000000000000004,
```
HHogg: 😂

packages/axiom-charts/src/utils/axisHelper.js
```diff
@@ -0,0 +1,35 @@
...
+
+export function getEquallyDistributedAxisLabels({
+    lower,
+    upper,
+    tickCount = DEFAULT_TICK_COUNT,
+    labelFormatter = function(v) { return v; },
+}) {
+  const finalTickCount = Math.max(MIN_TICK_COUNT, tickCount);
+  const tickSize = (upper - lower) / (finalTickCount - 1);
+  const ticks = Array.from(Array(finalTickCount).keys()).map(function(value, index) {
```
HHogg: It's the `.from` and `.keys()` which unfortunately needs rewriting pre2-ES015 😞

packages/axiom-charts/src/index.js
HHogg: I have a little issue with the naming, only because the default export now includes multiple utility functions, where as this naming (helper) implies a singular. Maybe it would be better named `axisUtils`?

packages/axiom-charts/src/utils/__snapshots__/axisHelper.test.js.snap
tomru: Yeah, was thinking about adding `labelFormatter: v => Number.parseFloat(v).toPrecision(1)` to the test, but thought that might make it less obvious that you need to take care of that :/

packages/axiom-charts/src/utils/axisHelper.test.js
```diff
@@ -0,0 +1,110 @@
...
+describe('getAxisUpper', () => {
+  it('returns 1 when max is below 1', () => {
+    expect(getAxisUpper([0, 0.5])).toEqual(1);
+  });
+
+  it('returns 10 when max is below 10', () => {
+    expect(getAxisUpper([1, 0, 9])).toEqual(10);
+  });
+
+  it('returns 100 when max is below 100', () => {
```
HHogg: I think there might be an issue with the function, when I run the data set `[0, 10, 20]` it returns `10` is that correct?

packages/axiom-charts/src/utils/axisHelper.js
tomru: oh, stupid me - sry

packages/axiom-charts/src/utils/__snapshots__/axisHelper.test.js.snap
HHogg: Yeah sorry, I was just having a language chuckle to myself :)

packages/axiom-charts/src/utils/axisHelper.test.js
tomru: Good catch -  [here](https://github.com/tomru/axiom/blob/950eaeeee7f06b496619a70a0e848f2a4ba7a55f/packages/axiom-charts/src/utils/axisHelper.js#L24) needs to go a `Math.ceil` instead of `Math.round` - will cover that with a couple more tests.

tomru: fixed in 91ac5c2

packages/axiom-charts/src/index.js
tomru: right, I choose `utils` as directory as well, so would be inline 👍

packages/axiom-charts/src/utils/axisHelper.js
tomru: addressed in df05ff8

tomru:
Thanks @HHogg - rebased now

PR #488, opened 2018-03-22
(feat): AlertDropdown

siggysamson:
This Pull Request adds a new `AlertDropdown` component.

I've modified `Context` to accept a `type` property from the `context` but wasn't sure whether this is actually correct or if we should explicitly pass it.
Initially I planned to use the existing `DropdownSource` and `DropdownTarget` components but then went for a separate implementation because they indeed are different.

screen shot 2018-03-22 at 15 19 37


packages/axiom-components/src/Context/Context.js
```diff
@@ -30,6 +30,10 @@ export default class Context extends Component {
     scrollIntoView: PropTypes.func.isRequired,
   };
 
+  static contextTypes = {
+    type: PropTypes.oneOf(['success', 'warning', 'error', 'info']),
```
HHogg: Good shout context definitely needs this. I would perhaps rename it to `color` though, as there is a task coming up to allow Tooltips to be any one of the product colours. So having `type="new-horizon"` might not make much sense, but `color="success"` still works and follows the `textColor="success"` pattern.

packages/axiom-components/src/Context/Context.js
```diff
@@ -59,12 +63,17 @@ export default class Context extends Component {
       ...rest
     } = this.props;
 
+    const {
+      type,
+    } = this.context;
+
     const classes = classnames('ax-context', `ax-context--${position}`);
+    const arrowClasses = classnames('ax-context__arrow', type ? `ax-context__arrow--${type}` : '');
```
HHogg: You can use the object format for this type of conditional class...  :)

```js
classnames('ax-context__arrow', {
  [`ax-context__arrow--${color}`]: color,
});
```

HHogg: Also I would suggest hoisting this up to the `ax-context` element because of the tooltip colouring behaviour coming which will also colour the background of any context content. (But don't worry about colouring that element for now)

packages/axiom-components/src/Context/Context.js
siggysamson: `color` or `colour`? 😜

packages/axiom-components/src/Context/Context.js
siggysamson: Makes sense, changed that 👍

siggysamson:
I've changed `Context` to accept a `color` context instead of a `type` context and also applied it to `ax-context` rather than the arrow itself.

One change I had to make though was to add a `color` and a `type` context to `AlertDropdown` because currently the `AlertDropdownHeader` expects a `type` context. This context is used for the `AlertBar`. For this one it makes sense to keep a `type` context because the `AlertBar` has a limited set of options. (Also looking at #487 it doesn't look like the `AlertBar` is affected by the color changes - which makes sense)

Do you think it is okay to pass the `type` property to `AlertDropdown` and then set it on the context as `type` and `color`? It feels a bit weird but I wasn't sure what a better solution would be at the moment.

packages/axiom-components/src/AlertDropdown/AlertDropdown.js
```diff
@@ -0,0 +1,99 @@
...
+
+    this.state = { isVisible: false };
+  }
+
+  getChildContext() {
+    return {
+      closeDropdown: () => this.close(),
+      openDropdown: () => this.open(),
+      type: this.props.type,
+      color: this.props.type,
```
HHogg: I think you could avoid having this on context and having Context component read off of `context.color` by putting the logic inside `AlertDropdownContext` for example.

```jsx
render() {
  return (
    <Context { ...this.props } color={ this.context.type } />
  );
}
```

But you're absolutely right, keeping it as `type` inside `AlertDropdown` makes sense as they are different patterns. As you said, Alerts are limited to the 4 alerting colours and Context is open to all product and alerting colours 👍

siggysamson:
Just pushed some more changes and also rebased so the fixup commits are gone 🙂

siggysamson:
Just pushed some more changes and also rebased so the fixup commits are gone 🙂

PR #494, opened 2018-03-26
fix(BarChart) do not break labels with (thin)spaces

tomru:
When using `tinyNumber` formatter on `xAxisLabels` on the bar chart, the last label is breaks on the thinspace for large numbers.

I'd argue labels on this chart should never break.

Before:
image

After:
image


tomru:
thanks @thatguynamedandy - opsi, my editor even told me 😅  -  fixed now

PR #496, opened 2018-03-27
fix(BarChart): hide other bars only when BarChart displays multiple bars

tomru:
When using the BarChart with multiple bars per label and hovering over a single bar on a label, other bars on that label are dimmed and on other labels the bars of different colour are hidden.

#### hover on BarChart with multiple bars per label

multiple-as-before

#### hover on BarChart with one bar per label

This does not make sense if you use just one bar per label, as there is no need to focus the users attention:
old

With this PR the BarChart will not hide if there is just **one** bar per label. If any label has more than one bar, the behaviour is the same as before.
new

### DIFF
--- a/packages/axiom-charts/src/BarChart/utils.js
+++ b/packages/axiom-charts/src/BarChart/utils.js
@@ -20,3 +20,6 @@ export const flattenValues = (data) => {
     return memo;
   }, []);
 };
+
+export const hasMultipleValues = (data) =>
+  data.some(({ values }) => Object.keys(values).length > 1);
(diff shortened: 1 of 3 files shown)


packages/axiom-charts/src/BarChart/utils.js
```diff
@@ -20,3 +20,8 @@ export const flattenValues = (data) => {
     return memo;
   }, []);
 };
+
+export const hasMultipleValues = (data) => {
+  const valuesCount = data.map(({ values }) => Object.keys(values).length);
+  return Math.max(...valuesCount) > 1;
```
HHogg: I think you could wrap this up in an exit early check with `some` like ... 

```js
export const hasMultipleValues = (data) => 
  data.some(({ values }) => Object.keys(values).length > 1));
```

So as soon as it hits a data set with multiple values it returns true.

HHogg:
This also might be a `fix()` 🤔  as there is no additional props or values. It'll just work out of the box :)

packages/axiom-charts/src/BarChart/utils.js
tomru: true, even better!

PR #497, opened 2018-03-28
Validate commit message to ensure correct format

stefanbuck:
Since we rely on the commit message for auto publishing new versions, we should validate them to ensure a new `npm publish` gets triggered after a PR is merged. 

This build proves that validating is working https://travis-ci.org/BrandwatchLtd/axiom/builds/359252903

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -10,6 +10,7 @@ branches:
 before_script:
 - yarn ci:setup
 script:
+- yarn lint:commits
 - yarn lint
 - yarn ci:test
 after_success:
--- a/package.json
+++ b/package.json
@@ -8,6 +8,7 @@
     "lint": "yarn run lint:js && yarn run lint:css",
     "lint:css": "stylelint 'packages/**/*.css' 'site/**/*.css'",
     "lint:js": "eslint --config .eslintrc packages site",
+    "lint:commits": "./scripts/validate_commit_messages.sh",
     "publish:packages": "bash ./scripts/publish_packages.sh",
     "publish:site": "bash ./scripts/publish_site.sh",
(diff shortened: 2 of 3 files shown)


stefanbuck:
It fails unexpected with ... need to investigate further. 

```
Error: ENOENT: no such file or directory, open '/home/travis/build/BrandwatchLtd/axiom/.git/COMMIT_EDITMSG'`
```

stefanbuck:
`COMMIT_EDITMSG` is essentially just a temporary file. Git creates it when you git commit. Therefore this file is not available on a CI.

stefanbuck:
The message is referring to a merge commit 😢

stefanbuck:
`git log -n 1 --no-merges --pretty=format:'%s' HEAD)` does the trick

stefanbuck:
Actually just the last commit, wich I realise is not enough ... but I have an idea how to solve this, stay tuned!

stefanbuck:
Now it checks all commits. Special thanks to @tomru for his bash skills

thatguynamedandy:
@stefanbuck / @tomru looks awesome, could I be fussy and just request one thing. Is it possible to move the bash into the `/scripts` directory, maybe call it `validate_commit_messages.sh` (or similar). Then in `package.json` add it to the scripts..

```json
"scripts": {
  "lint:commits": "./scripts/validate_commit_messages.sh"
},
```

This should hopefully give us the option to...
a) Run it locally if we want (to avoid surprises)
b) It would be easier to use in other places, if for example we moved to Concourse or similar

stefanbuck:
No worries, I like your suggestion and it make sense. I added the change to the previous commit to keep the history clean

PR #499, opened 2018-03-29
STS-81 allow sentiment colors in bar chart

tomru:
### Allows sentiment colors in the BarChart

- fixed the typo ` --color-sentiment-neagtive` and marked it as a fix, as I didn't see it anywhere used on github
- Added the color `sentiment-neutral` as specified by @lilybrandwatch
- addes support for  sentiment colors for `Bar` and `DataPoint` components

image

### DIFF
--- a/packages/axiom-charts/src/Bar/Bar.css
+++ b/packages/axiom-charts/src/Bar/Bar.css
@@ -49,6 +49,9 @@
   &.ax-bars__bar-rect--space-oddity { background-color: var(--color-product-space-oddity); }
   &.ax-bars__bar-rect--deep-thought { background-color: var(--color-product-deep-thought); }
   &.ax-bars__bar-rect--luna-dust { background-color: var(--color-product-luna-dust); }
+  &.ax-bars__bar-rect--sentiment-positive { background-color: var(--color-sentiment-positive); }
+  &.ax-bars__bar-rect--sentiment-negative { background-color: var(--color-sentiment-negative); }
+  &.ax-bars__bar-rect--sentiment-neutral { background-color: var(--color-sentiment-neutral); }
 }
 
 .ax-bars__bar--clickable {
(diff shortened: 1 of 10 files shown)


packages/axiom-charts/src/Bar/Bar.css
```diff
@@ -49,6 +49,9 @@
   &.ax-bars__bar-rect--space-oddity { background-color: var(--color-product-space-oddity); }
   &.ax-bars__bar-rect--deep-thought { background-color: var(--color-product-deep-thought); }
   &.ax-bars__bar-rect--luna-dust { background-color: var(--color-product-luna-dust); }
+  &.ax-bars__bar-rect--sentiment-positive { background-color: var(--color-sentiment-positive); }
```
HHogg: Maybe an oversight on my part, but I didn't prefix the others with `product`... do you think we should for the sentiment ones, or just go with `positive`, `neutral` and `negative`?

packages/axiom-materials/src/colors.css
```diff
@@ -147,14 +147,18 @@
   --color-product-luna-dust: rgb(120, 144, 156);
   --color-product-luna-dust--hover: rgb(138, 158, 168);
 
-  --color-sentiment-neagtive--active: rgb(201, 58, 58);
```
HHogg: 🤦‍♂️ ....   thank you!

packages/axiom-materials/src/colors.js
```diff
@@ -51,6 +51,10 @@ export const sentimentNegativeActive = { r: 201, g: 58, b: 58 };
 export const sentimentNegative = { r: 215, g: 74, b: 74 };
 export const sentimentNegativeHover = { r: 227, g: 89, b: 89 };
 
+export const sentimentNeutralActive = { r: 167, g: 162, b: 179 };
+export const sentimentNeutral = { r: 187, g: 185, b: 192 };
+export const sentimentNeutralHover = { r: 204, g: 204, b: 204 };
```
HHogg: As this is kind of an alias, could this just the `productDeepThought[hover/active]` constants in this file too like you've done with the CSS variables?

packages/axiom-charts/src/Bar/Bar.css
tomru: Right, I'm a bit torn apart on this one. 

For the product colors we just use the color names, and there is not really a benefit of knowing that this is a product color.

But for things like sentiment (and the soon coming gender) I'd thought keeping the prefix to make it clear what they relate to.

packages/axiom-materials/src/colors.js
tomru: But need to move it below the definition of `productDeepThought` to be able to reference it

packages/axiom-charts/src/Bar/Bar.css
HHogg: Yeah that makes sense 👍

tomru:
@HHogg adopted now `Line` and `LinePlot` so all charts do now support sentiment colors

PR #500, opened 2018-03-29
chore(CI): Detect modified yarn.lock on CI

t-sauer:
After running `lerna bootstrap` I immediately had a modified `yarn.lock` because the lockfile wasn't updated after the `package.json` was changed. With this change, changes to the `package.json` without updating the lockfile are detected and will fail CI.
Ideally I would do this with the yarn option `--frozen-lockfile` but I have no idea how to do that with lerna. The only option I found was to add it to `npmClientArgs` in `lerna.json` but then it would also freeze the lockfile in development which is quite inconvenient.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -9,6 +9,7 @@ branches:
     - master
 before_script:
 - yarn ci:setup
+- git diff --exit-code # make sure that yarn.lock didn't change
 script:
 - yarn lint:commits
 - yarn lint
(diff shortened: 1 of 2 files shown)


.travis.yml
```diff
@@ -9,6 +9,7 @@ branches:
     - master
 before_script:
 - yarn ci:setup
+- git diff --exit-code # make sure that yarn.lock didn't change
```
stefanbuck: https://git-scm.com/docs/git-diff#git-diff---exit-code

PR #501, opened 2018-03-29
feat(Base): Add container property to Base component for relative positioning

t-sauer:
This adds a new property called `container` to the Base component. When set to true it will set `position: relative;` on the element.

### DIFF
--- a/packages/axiom-components/src/Base/Base.test.js
+++ b/packages/axiom-components/src/Base/Base.test.js
@@ -95,6 +95,12 @@ describe('Base', () => {
     expect(tree).toMatchSnapshot();
   });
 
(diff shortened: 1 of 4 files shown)


packages/axiom-components/src/Base/Base.test.js
```diff
@@ -95,6 +95,12 @@ describe('Base', () => {
     expect(tree).toMatchSnapshot();
   });
 
+  it('renders sticky with container', () => {
```
HHogg: Maybe a slight incorrect message here? (sticky).

t-sauer: Oops, it's fixed now 😅

PR #506, opened 2018-04-04
fix(axisUtils): when upper and lower are equal return no labels

tomru:
This guards against getting duplicate axis labels, of e.g. [0, 0, 0] for e.g. 

```js
axisUtils.getEquallyDistributedAxisLabels({upper=0, lower=0})
```

In case upper and lower are equal, just return no labels at all.

This currently causes React to loose track of the labels, as it sets a `key` of the label [here](https://github.com/BrandwatchLtd/axiom/blob/ec6eb66314f2014337318476b8308f09a1a9e99d/packages/axiom-charts/src/ChartTable/ChartTableRows.js#L66). This should probably be fixed separately as well.

### DIFF
--- a/packages/axiom-charts/src/utils/axisUtils.js
+++ b/packages/axiom-charts/src/utils/axisUtils.js
@@ -7,6 +7,11 @@ export function getEquallyDistributedAxisLabels({
     tickCount = DEFAULT_TICK_COUNT,
     labelFormatter = v => v,
 }) {
+
+  if (lower === upper) {
+    return [];
+  }
+
   const finalTickCount = Math.max(MIN_TICK_COUNT, tickCount);
   const tickSize = (upper - lower) / (finalTickCount - 1);
   const ticks = [];
--- a/packages/axiom-charts/src/utils/axisUtils.test.js
+++ b/packages/axiom-charts/src/utils/axisUtils.test.js
@@ -18,6 +18,23 @@ describe('getEquallyDistributedAxisLabels', () => {
     expect(xAxisLabels).toMatchSnapshot();
   });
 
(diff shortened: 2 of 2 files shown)


packages/axiom-charts/src/utils/axisUtils.js
```diff
@@ -8,7 +8,8 @@ export function getEquallyDistributedAxisLabels({
     labelFormatter = v => v,
 }) {
   const finalTickCount = Math.max(MIN_TICK_COUNT, tickCount);
-  const tickSize = (upper - lower) / (finalTickCount - 1);
+  const finalUpper = lower === upper ? lower + 1 : upper;
```
HHogg: Just curious as to why the lower is incremented and not the upper is decremented? Or why both aren't incremented and decremented so the label is centered? 

_(I don't know the use case so this comment could be reeeally irrelevant)_

tomru: Oh, increasing the lower is makes the less sense, I meant to increase the upper thb 😄 

But thinking again, maybe we should simply return an empty array if the `upper` and `lower` is the same instead of picking some kind of range.

The main use-case is that when using the axisUtil you'll never end up multiple labels with the same number.

What do you think?

HHogg: I'm not really sure, to be honest. I imagine that if there was a single same value (basically a straight line) then it would probably appear vertically centered to the chart with a single label (also centered)... so would it make sense to return an array with a single value... like ... 

```js
if (lower === upper) return [lower];
```

tomru: As discussed, return no label if upper a lower are equal.

PR #508, opened 2018-04-05
feat(DropdownMenuItem): Adds keepOpen property

thatguynamedandy:
Enables the menu to stay open after the item is clicked


thatguynamedandy:
hi @stefanbuck agreed tests are a bit thin on the ground.

I've added some functional and snapshot tests. The approach is slightly new for Axiom, the `react-test-renderer` isn't great for dealing with context and some slightly more complex scenarios. So I have switched to `Enzyme` which is great for functional testing, and it then uses `enzyme-to-json` for the snapshot testing.

PR #513, opened 2018-04-06
fix(Candytar): Only the user defined product colour set are supported…

HHogg:
… colours


HHogg:
@thatguynamedandy They won't change, but this fixes the issue with completely black candytars after the new product colours were added.

PR #517, opened 2018-04-09
Dropdown - Fixes issues with multi selecting from different menus and...

HHogg:
... also improves performance for larger lists. 

[Netlify](https://fix-dropdown--hhogg-axiom.netlify.com/docs/packages/axiom-components/select/)


thatguynamedandy:
I'll try and have a proper play with this, one quick thing I noticed was that if you keyboard cycle down it will skip back to the beginning, but cycle up and it won't overlap. (my preference is probably for it just to stop, that's the default browser behavior)

HHogg:
@thatguynamedandy ah yes good catch that shouldn't happen. I'll update

PR #522, opened 2018-04-12
chore(Transition): Add new Transition component

stefanbuck:
This PR adds a new `<Transition>` component which (surprise) allows you to animate components. I started with a very simple implementation which was rendering all children next to each other. By setting the parent containers  `width`  to `${children.length} * 00%` I was able to use `left: -${currentIndex}00%` for the transition.  This approach has one serious disadvantage: I had to render all children upfront:

### First implementation
old

### Latest implementation

We don't need more than two children at the same time to make the transition from one state to antoher. I started with a new implementation with the goal of rendering less as possible children. As you can see on the gif below, as soon as you navigate to the next item, this item will be rendered outside of the visible area (either on the left or right). After the transition the previous child get removed to ensure the same state after each transition.

new

👉  [Live demo](https://brave-keller-55c6a8.netlify.com/docs/packages/axiom-components/transition) 👈 

#### Known issues

- ✅ If you click really fast on `next` the animation will look awkward. 
   A: ~~This needs more thinking how to solve this in a proper way)~~ 
        Solved by using `react-transition-group`
- ⚠️ It doesn't prevent you from going beyond `children.length` or below zero. 
  A: I wanted to create a pagerHelper anyway which should solve this issue

### Todo
- [ ] Add tests
- [ ] Add helper to calculate next and previous index.


jdelight:
Nice work @stefanbuck.

Could be a good use case for https://developer.mozilla.org/en-US/docs/Web/API/IntersectionObserver (your second diagram particularly)
and possibly [react-transition-group](https://github.com/reactjs/react-transition-group)?

packages/axiom-components/src/Transition/Transition.js
```diff
@@ -0,0 +1,94 @@
...
+      'ax-transition-move-to-right': directionPrevious,
+    });
+
+    const style = {
+      animationDuration: `${animations.transitionTimeSlow}ms`,
+    };
+
+    return (
+      <div className="ax-transition__wrapper" style={ { height, width } }>
+        <div className={ classesIn } ref="animationRef" style={ style }>{ childIn }</div>
```
SeaOfTea: Just a note on string refs. https://reactjs.org/docs/refs-and-the-dom.html#legacy-api-string-refs

packages/axiom-components/src/Transition/Transition.js
```diff
@@ -0,0 +1,94 @@
...
+import { animations } from '@brandwatch/axiom-materials';
+import './Transition.css';
+
+export default class Transition extends Component {
+  static propTypes = {
+    activeIndex: PropTypes.number,
+    children: (props, propName, componentName) => {
+      const children = props[propName];
+
+      const hasNotSupportedChildren = Children.toArray(children).some((child) => !isComponent(child, TransitionItemRef));
```
HHogg: Something [we found recently](#517), is that we cannot guarantee that legitimate renderable structures will always be in the tree where we expect them to be. For example this would fail this propType check... 

```jsx
<Transition> 
  <TransitionItem>...</TransitionItem>
  { someCondition && (
    <Fragment>
      <TransitionItem>...</TransitionItem>
      <TransitionItem>...</TransitionItem>
    </Fragment>
  ) }
</Transition>
```

This would also fail this propType ... 

```jsx
<Transition>
  <TransitionItem>...</TransitionItem>
  <MyAppSpecificResusableTransitionItem>...</MyAppSpecificResusableTransitionItem>
</Transition>
```

_(Where MyAppSpecificResusableTransitionItem uses TransitionStep under the hood)_. 

Note: Axiom still has this pattern in a few places, but we should probably avoid there being any more.

HHogg: A `children: PropType.node,` should be fine here.

packages/axiom-components/src/Transition/Transition.js
```diff
@@ -0,0 +1,94 @@
...
+    elm.addEventListener('animationend', this.animationEnd);
+  }
+
+  componentWillUnmount() {
+    const elm = this.refs.animationRef;
+    elm.removeEventListener('animationend', this.animationEnd);
+  }
+
+  animationEnd() {
+    this.forceUpdate();
```
HHogg: I agree with @jdelight around looking into `react-transition-group`. We've experimented with this to handle the animations of showing and hiding the context components, and handles the delayed unmounts quite well. 

It would remove a lot of the logic around tracking what is mounted, what is in a certain transitioning state. It also has a CSSTransition component so the CSS that you already have would still be applicable.

stefanbuck:
I added a bunch of fixup commits to share my progress in small portions 😉

- 2c266321bb488c5065c34a00cf48ea0ef557ab7b is adding `react-transition-group` with a first working implemention
- a376a6cc66023bd599b5665f4708e05313a3f9c3 is fixing an issue when navigation back and forth. 
- 91090c7f9dbe2f65c3ee54253536ae43f7916c2b removes the custom propType validation function as requested by @HHogg 
-  c5f357154f50e8cc0e243f65efb65931c04acc55 removes the no longer needed `<TransitionItem>` component

packages/axiom-components/src/Transition/Transition.js
stefanbuck: @jdelight @HHogg I've added `react-transition-group` and it removes indeed a lot of logic. Thank you for the suggestion.

packages/axiom-components/src/Transition/Transition.js
stefanbuck: Thanks for explaining the issue in detail 👍I removed the custom validation in favour of `PropTypes.node`

packages/axiom-components/src/Transition/Transition.js
stefanbuck: Good call, while I was implementing `<Transition>` I read about the new `create Ref API` and wanted to change ... but I forgot. Anyway, after switching to `react-transition-group` the listener for `animationEnd` is not needed anymore.

packages/axiom-components/src/Transition/Transition.js
```diff
@@ -0,0 +1,62 @@
...
+    const childFactoryCreator = (classNames) => (
+      (child) => (
+        React.cloneElement(child, {
+          classNames,
+        })
+      )
+    );
+
+    return (
+      <TransitionGroup childFactory={ childFactoryCreator(className) } className="ax-transition__wrapper" style={ { height, width } }>
```
HHogg: I think you could avoid the childFactory with `component={ Base }` as the merging of `classNames` already happens inside of it. If you combine this with a `...rest` spread of the props as well, it will have the benefits of the Base props.

```
<Paragraph>...</Paragraph>
<Transition space="x2" theme="night">...</Transition>
<Paragraph>...</Paragraph>
```

packages/axiom-components/src/Transition/Transition.js
```diff
@@ -0,0 +1,62 @@
+import PropTypes from 'prop-types';
+import React, { Component } from 'react';
+import { CSSTransition, TransitionGroup } from 'react-transition-group';
+import { animations } from '@brandwatch/axiom-materials';
+import './Transition.css';
+
+export default class Transitio2n extends Component {
```
HHogg: "Rouge Two - A Star Wars Story".

packages/axiom-components/src/Transition/Transition.css
```diff
@@ -0,0 +1,54 @@
+.ax-transition__wrapper  {
+  position: relative;
+  overflow-x: hidden;
+}
+
+.ax-transition__item--current {
+  z-index: 2;
```
HHogg: Is this needed anymore?

packages/axiom-components/src/Transition/Transition.css
```diff
@@ -0,0 +1,54 @@
...
+}
+
+.ax-transition__item {
+  position: absolute;
+  animation-fill-mode: forwards;
+  width: 100%;
+}
+
+.animate-next-enter-active {
```
HHogg: You can configure the entering and exiting classnames to have the `ax` prefix.

packages/axiom-components/src/Transition/Transition.js
stefanbuck: I wonder how that sneaked in :midnight-coding:

packages/axiom-components/src/Transition/Transition.css
stefanbuck: Nope

packages/axiom-components/src/Transition/Transition.css
stefanbuck: Good point! Yes, that's possible.

packages/axiom-components/src/Transition/Transition.js
stefanbuck: Cool, will try that out. Thanks

stefanbuck: Seems to not work. @HHogg can you check 669619750bd87f627b8ffc62a2c613eed4fa9317 please. Maybe I implemented it wrong.

stefanbuck: The `component` prop defines the outer container not the children
 http://reactcommunity.org/react-transition-group/transition-group#TransitionGroup-prop-component

stefanbuck: Maybe still a good idea to use `Base` as the childs container instead of a `div`.

PR #524, opened 2018-04-13
feat(navigationUtils): Add navigation utils

stefanbuck:
This is needed for #522


packages/axiom-components/src/utils/navigationUtils.js
```diff
@@ -0,0 +1,30 @@
...
+  }
+
+  return 0;
+}
+
+export default {
+  hasNext,
+  hasPrev,
+  getNext,
+  getPrev,
```
stefanbuck: Do people prefer `getPrevious` over `getPrev`? If so, I'm happy to change it

HHogg:
Haha I like the enjoyable sentiment 😄 Yeah I think excluding this would be best for now, axiom already exports quite a lot and we should keep it focused on the design and UX aspect, but also happy to get other opinions on it.

thatguynamedandy:
I'm a bit unsure on letting too many utils in, they can be pretty useful, but they do have a tendency to get orphaned / out of date if not widely used, and we would have to be careful to semver them correctly.

Looking at #522 I can see the use case though, however I think if we were for example wanting to make an `ImageCarousel` we would probably bake that logic into the component itself.

stefanbuck:
Thanks for your thoughts.

PR #525, opened 2018-04-13
feat(LineChart): Added onPointClick action to the LineChart

t-sauer:


packages/axiom-charts/src/Line/LinePoint.js
```diff
@@ -115,7 +118,7 @@ export default class LinePoint extends Component {
             onDropdownOpen={ onDropdownOpen }
             style={ style }
             value={ value }>
-          <DataPoints { ...rest } size={ size }>
+          <DataPoints { ...rest } onClick={ () => onClick({ color, index, label, value }) } size={ size }>
```
HHogg: Should probably only try to call `onClick` if an `onClick` is given otherwise it will throw and error? 

```jsx
onClick={ onClick && () => onClick(...) }
```

packages/axiom-components/src/Tooltip/TooltipTarget.js
```diff
@@ -40,6 +40,7 @@ export default class TooltipTarget extends Component {
       ...rest,
       onMouseMove: this.handleMouseMove.bind(this),
       onMouseLeave: this.handleMouseLeave.bind(this),
+      onClick: children.props.onClick,
```
HHogg: I think you need to merge this is with the `onClick` given to TooltipTarget from Tooltip (similar to how DropdownTarget handles it). This is to support the nested Tooltip/Dropdown behaviour. 

_There is probably a better way to structure these components, so you don't need to do this but that would be a quicker fix for now_

t-sauer:
@HHogg Updated the PR. Hope I got it right.

PR #528, opened 2018-04-17
More specific cache control directives

ahume:
- Only cache /assets with far-future directives.
- Set a max-age on html to stop Expires header being set to
  far-future.


ahume:
@HHogg Think this is ready to look at again.

PR #529, opened 2018-04-18
fix(Transition): Slide into correct direction on initial change

t-sauer:


packages/axiom-components/src/Transition/Transition.js
```diff
@@ -13,8 +13,13 @@ export default class Transition extends Component {
     width: PropTypes.string.isRequired,
   };
 
-  componentDidUpdate(prevProps) {
-    this.previousIndex = prevProps.activeIndex;
+  constructor(props) {
+    super(props);
+    this.previousIndex = props.activeIndex || 0;
```
HHogg: Should this be set with `defaultProps`? 🤔

t-sauer:
Have to go over this fix again. It's not actually fixing everything.

packages/axiom-components/src/Transition/Transition.js
t-sauer: `activeIndex` is a required prop, so the `|| 0` isn't needed.

PR #530, opened 2018-04-18
Sticky Dropdown Header and Footer example. 

HHogg:
This was technically already kind of possible, except for:

* a bug with the placement of the separator on a scrollable list (6a95a67)
* compact content area with smaller vertical padding (055f3d8)

I also added an alias component for `DropdownHeader` and `DropdownFooter` (e517994), although these aren't really needed so happy to drop off if anybody thinks they are just noise. 

I also moved over the propType docs to Dropdown (5f514a4) , because the lower level Context components do not appear in the Style Guide.

### DIFF
--- a/packages/axiom-components/src/Dropdown/DropdownMenuItem.js
+++ b/packages/axiom-components/src/Dropdown/DropdownMenuItem.js
@@ -5,8 +5,11 @@ import ContextMenuItem from '../Context/ContextMenuItem';
 
 export default class DropdownMenuItem extends Component {
   static propTypes = {
+    /** SKIP */
     children: PropTypes.node,
-    /**
+    /** Disabled interactions and applies styling */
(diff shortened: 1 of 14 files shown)


packages/axiom-components/src/Dropdown/DropdownMenuItem.js
```diff
@@ -5,8 +5,11 @@ import ContextMenuItem from '../Context/ContextMenuItem';
 
 export default class DropdownMenuItem extends Component {
   static propTypes = {
+    /** SKIP */
```
stefanbuck: Will this prevent the prop from dispalying in the documentation?

HHogg: Yes it will :)

PR #531, opened 2018-04-19
feat(ContextMenuItem): Add icon prop to specify menu item icon

stefanbuck:
screen shot 2018-04-19 at 22 22 21

### DIFF
--- a/packages/axiom-components/src/Context/ContextMenuItem.js
+++ b/packages/axiom-components/src/Context/ContextMenuItem.js
@@ -11,6 +11,7 @@ export default class ContextMenuItem extends Component {
   static propTypes = {
     children: PropTypes.node,
     disabled: PropTypes.bool,
+    icon: PropTypes.string,
     multiSelect: PropTypes.bool,
     onClick: PropTypes.func,
     selected: PropTypes.bool,
@@ -20,6 +21,7 @@ export default class ContextMenuItem extends Component {
     const {
       children,
       disabled,
+      icon,
       onClick,
       multiSelect,
       selected,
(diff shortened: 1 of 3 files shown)


packages/axiom-components/src/Context/ContextMenuItem.js
```diff
@@ -52,6 +54,12 @@ export default class ContextMenuItem extends Component {
             <Icon cloak={ !selected } name="tick" />
           </div>
         ) }
+
+        { icon !== undefined && (
+          <div className="ax-context-menu__item-icon">
+            <Icon name={ icon } />
+          </div>
+        ) }
```
stefanbuck: Ideally, we should combine the `tick` and the custom icon, but I couldn't come up with an appropriate solution.

thatguynamedandy: Tricky one, this approach gives lots of flexibility, but I guess the pattern we are trying to create is actually a component that allows you to move down to the next level of menu items. 

I'm torn between keeping it flexible but having the slight issue you raise, or locking it down to the specific pattern.

HHogg:
```jsx
<ContextMenuItem
    icon="tick"
    iconSize="1rem"
    iconOnClick={ () =>  }
    iconRenderer={ (children) => <Animation>{ children }</Animation> }
    iconProps={ {
      onClick: () => {},
      'data-attribute': 'something',
    } }>
  ...
</ContextMenuItem>
```

As opposed to...

```jsx
<ContextMenuItem>
  <Animation>
    <ContextMenuItemIcon  
        data-attribute="something"
        onClick={ () => {} } 
        name="tick" 
        size="1rem" />
  </Animation>
</ContextMenuItem>
```

Having said that... I think this is ok, it's a nice shortcut and if the it doesn't fit a one off scenario, then it can still be accomplished with the render of ContextMenuItem :)

stefanbuck: I'm happy to rebase on top of #532 and show an arrow-right icon based of the child items.

stefanbuck: @HHogg I'm happy either way, just let me know what works best for you

HHogg: I'm happy with it as it is @stefanbuck 👍

PR #532, opened 2018-04-20
feat(Dropdown): add support for submenus

tomru:
Adds support for submenus into ContextMenus as well as DropdownMenus.

### Visuals

image


packages/axiom-components/src/Context/ContextMenuItem.test.js
```diff
@@ -54,4 +55,14 @@ describe('ContextMenuItem', () => {
...
+
+  it('renders subMenuItems', () => {
+    const subMenuItems = [
+      <ContextSubMenuItem key="1">Subitem 1</ContextSubMenuItem>,
+      <ContextSubMenuItem key="2">Subitem 2</ContextSubMenuItem>,
+    ];
+    const component = getComponent({ subMenuItems });
+    const tree = component.toJSON();
+    expect(tree).toMatchSnapshot();
+  });
```
stefanbuck: Can you add test for the new prop `className` as well please.

packages/axiom-components/src/Context/ContextMenuItem.js
```diff
@@ -10,49 +10,56 @@ export const contextMenuItemSelector = 'data-ax-context-menu-item';
...
-        ) }
-      </Base>
+
+          { !multiSelect && selected !== undefined && (
+            <div className="ax-context-menu__item-icon">
+              <Icon cloak={ !selected } name="tick" />
+            </div>
+          ) }
+        </Base>
+        { subMenuItems }
```
stefanbuck: Just an idea to avoid creating `ContextSubMenuItem` which is just adding a class name. What if we iterate over `subMenutItems` and if the component is from type `ContextSubMenuItem` we apply `ax-context-submenu`.

packages/axiom-components/src/Context/Context.css
```diff
@@ -154,3 +154,7 @@
   flex: 0 0 auto;
   margin-left: var(--spacing-grid-size--x2);
 }
+
+.ax-context-submenu {
+  padding-left: calc(var(--spacing-grid-size) * 10);
```
HHogg: If this is the value that was used for the screenshot, It might be a tad off. 

The designs show the submenu checkbox lining up with the beginning of text from the first level menu option.

packages/axiom-components/src/Context/ContextMenuItem.js
HHogg: We normally try to avoid passing big blocks of react components in as props in order to keep render functions easier to read from top to bottom. Passing in blocks like this, normally require you need to assign a block of components above in the render function and then pass them in at the bottom. Also I think you could avoid having an extra component/class and altering this components render function by having the css just indent the extra menu items when the structure is `ul > li > ul`.

Stripped down example... 

```jsx
const submenuItems =  [
  <DropdownMenuItem>
    Lorem Ipsum
  </DropdownMenuItem>
];
      
<DropdownMenu>
  <DropdownMenuItem subMenuItems={ submenuItems }>
    Lorem Ipsum
  </DropdownMenuItem>
</DropdownMenu>
```

```jsx
<DropdownMenu>
  <DropdownMenuItem>
    Lorem Ipsum
  </DropdownMenuItem>

  <DropdownMenu submenu>
    <DropdownMenuItem>
      Lorem Ipsum
    </DropdownMenuItem>
  </DropdownMenu>
</DropdownMenu>
```

tomru: Great point, that would indeed make most of the changes obsolete.

packages/axiom-components/src/Context/Context.css
```diff
@@ -119,6 +119,10 @@
 
 .ax-context-menu {
   padding: var(--cmp-context-tip-space) 0;
+
+  & .ax-context-menu .ax-context-menu__item {
```
tomru: Is there any more clever way to write this?

Further this rule only applies to the first level of submenus, which isn't ideal either 😕

tomru:
I also updated the docs, to cover sub menu item.s

packages/axiom-components/src/Context/Context.css
HHogg: I think that's ok. Allowing further levels of nesting should go through some design investigations first. For example, they might decide that 3 levels is not great UX and a different pattern should be used instead. It's something that can be tackled as and when it's needed :)

stefanbuck:
This reminds me that I wanted to document how to get Netfily working, which I haven't done yet 😞

packages/axiom-components/src/Context/Context.css
```diff
@@ -166,3 +166,7 @@
   flex: 0 0 auto;
   margin-left: var(--spacing-grid-size--x2);
 }
+
+.ax-context-menu .ax-context-menu .ax-context-menu__item {
```
tomru: @HHogg fyi, stylelint was complaining, needed to move this down.

tomru:
https://youthful-albattani-41ebb2.netlify.com/docs/packages/axiom-components/dropdown for the design review

PR #535, opened 2018-04-23
 Pass closeDropdown action from context down to DropdownSource component

t-sauer:


t-sauer:
Worked on an example but it ended up being messy. Will definitely come back and add one later though.

PR #536, opened 2018-04-25
feat(TextInput): Added tooltipMessage prop which is shown as an error message in an invalid state

t-sauer:
screen shot 2018-04-25 at 15 05 52


packages/axiom-components/src/Form/TextInput.js
```diff
@@ -120,29 +129,42 @@ export default class TextInput extends Component {
...
-            <div className="ax-input__button-container">
-              <div className={ classes(isValid) }>
-                { onClear && value ? (
-                  <TextInputIcon
-                      align="right"
-                      name="cross"
-                      onClick={ onClear } />
-                  ) : icon
-                }
+          <Position isVisible={ invalid && !!tooltipMessage } position="bottom">
```
HHogg: Not every use of TextInput needs inline validation styling, so it might be a bit much to include it in every TextInput. Maybe this could be separated out into an `InlineValidation` (or something similar) component that users can wrap components in? For example...

```jsx
<InlineValidation error="My error message">
  <TextInput invalid />
</InlineValidation>

<InlineValidation error="My error message">
  <TextArea invalid />
</InlineValidation>

<InlineValidation error="My error message">
  <ApplicationSpecificComponent />
</InlineValidation>
```

HHogg: Maybe something that could live with the Validation components also 🤔

HHogg: It would also have the ability to pass props to the Position component like ... 

```jsx
<InlineValidation flip="mirror" offset="start" position="top">
  ...
</InlineValidation>
```

t-sauer: Definitely a good idea, I am on it.

t-sauer:
I've changed it to a separate `InlineValidation` component.

packages/axiom-components/src/Validation/InlineValidation.js
```diff
@@ -0,0 +1,42 @@
...
+import PositionSource from '../Position/PositionSource';
+import TooltipContent from '../Tooltip/TooltipContent';
+import TooltipContext from '../Tooltip/TooltipContext';
+import Text from '../Typography/Text';
+
+export default class InlineValidation extends Component {
+
+  static propTypes = {
+    children: PropTypes.node.isRequired,
+    tooltipMessage: PropTypes.string,
```
thatguynamedandy: Really minor but I wonder about 'message' instead of 'tooltipMessage', feels slightly less verbose.

packages/axiom-components/src/Validation/InlineValidation.js
```diff
@@ -0,0 +1,42 @@
...
+    position: 'bottom',
+  }
+
+  render() {
+    const { children, tooltipMessage, ...rest } = this.props;
+
+    return (
+      <Position isVisible={ !!tooltipMessage } { ...rest }>
+        <PositionTarget>
+          { cloneElement(children, { ...rest }) }
```
thatguynamedandy: I think my comment here https://github.com/BrandwatchLtd/axiom/pull/537#issuecomment-384624159 might apply to this, but I know the `Position` element eventually renders a portal so no real risk of multiple unique attributes being shared across multiple elements.

Does `Position` need the props spread to it, or could it simply be.. 
`<Position isVisible={ !!tooltipMessage }>`

packages/axiom-components/src/Validation/InlineValidation.js
t-sauer: I agree, `message` is better 👍

packages/axiom-components/src/Validation/InlineValidation.js
t-sauer: For passing down the the rest props: We wrap the `InlineValidation` inside of a `DropdownTarget`. Without spreading the props, the clicking behaviour of the `DropdownTarget` didn't work. I assume that only a certain subset of properties need to be passed down, but I am not sure if it is better or worse to pass these individual properties down or just pass down everything.
For passing down the other rest props to `Position`: The idea was that you can manipulate the behaviour the position of the tooltip for example (see https://github.com/BrandwatchLtd/axiom/pull/536#discussion_r184108775).

thatguynamedandy: I see what you mean now, I think I would be more comfortable to pass the specific props down to `Position`, there is only a finite amount of properties that 'Position' expects, and it also gives you the ability to be more opinionated with this component, for example you could  restrict the `position` prop to being just `['top', 'bottom']`, just enable what you need.

t-sauer:
I've did a deployment, an example can be found here: https://naughty-pasteur-6d413a.netlify.com/docs/packages/axiom-components/form/

PR #537, opened 2018-04-26
Pass rest props down to children 

stefanbuck:
This is needed so I can make use of #535

### DIFF
--- a/packages/axiom-components/src/Context/Context.js
+++ b/packages/axiom-components/src/Context/Context.js
@@ -1,5 +1,5 @@
 import PropTypes from 'prop-types';
-import React, { Component } from 'react';
+import React, { isValidElement, cloneElement, Component } from 'react';
 import classnames from 'classnames';
 import Base from '../Base/Base';
 import './Context.css';
@@ -44,7 +44,7 @@ export default class Context extends Component {
         ) }
 
         <div className="ax-context__content" style={ { maxHeight } }>
-          { children }
+          { isValidElement(children) ? cloneElement(children, rest) : children }
         </div>
       </Base>
     );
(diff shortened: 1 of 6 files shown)


thatguynamedandy:
Hey @stefanbuck I'm just trying to get my head around the use case a little, in general do you not already have full control of the children elements to pass them the props that they need? I'm sure there is a need but could do with some help visualising it.

My only concern is just that it sets a new precedent, previously we only spread props to one element so I just want to make sure I full understand any impact of spreading the same props onto multiple components. One particular concern is that if some passes properties like a data attribute for automation tests that attribute would now be on two elements causing some unwanted side affects.

stefanbuck:
It's possible that there is also another way of doing this. What I basically want to do is allowing closing the context menu when the users clicks on the apply button rather than on the menu item itself. I'll send you a DM on Slack with my actual PR


### Pseudo code example

```
<Dropdown>
    <DropdownTarget>Open</DropdownTarget>
    <DropdownSource>
        <DropdownContext>
            <Transition>
                <DropdownMenu>
                    <DropdownMenuItem>Item 1</DropdownMenuItem>
                    <DropdownMenuItem>Item 2</DropdownMenuItem>
                    <DropdownMenuItem>Item 3</DropdownMenuItem>
                </DropdownMenu>
                <Button onClick={() => {
                    closeDropdown();
                }}>Apply</Button>
            </Transition>
        </DropdownContext>
    </DropdownSource>
</Dropdown>
```

stefanbuck:
Using context does the trick. 

```
   static contextTypes = {
        closeDropdown: PropTypes.func.isRequired,
    };
```

PR #538, opened 2018-04-26
feat(CheckBox): Added indeterminate property

t-sauer:
This basically adds a third state to the checkbox.
Livedemo: https://naughty-pasteur-6d413a.netlify.com/docs/packages/axiom-components/form/


t-sauer:
Thanks for the hint, should be good now.

PR #542, opened 2018-05-01
feat(TableRow): adds a hasBorder property to TableRow and TableCell

pmsorhaindo:
Adds a `hasBorder` property to TableRow and TableCell to allow for borderless rows and cells.

### DIFF
--- a/packages/axiom-components/src/Table/TableRow.js
+++ b/packages/axiom-components/src/Table/TableRow.js
@@ -5,6 +5,8 @@ import Base from '../Base/Base';
 
 export default class TableRow extends Component {
   static propTypes = {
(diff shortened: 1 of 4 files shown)


packages/axiom-components/src/Table/TableRow.js
```diff
@@ -1,19 +1,39 @@
...
+
+  constructor(props) {
+    super(props);
+
+    this.renderChildren = this.renderChildren.bind(this);
+  }
+
+  renderChildren() {
+    const { children, hasBorder } = this.props;
+    return Children.map(children, child => {
```
HHogg: As discussed, this can be avoided by just applying a CSS class to the row.

Issue #539, opened 2018-04-27
Between DropdownSubMenu a line is drawn


tomru:
![image](https://user-images.githubusercontent.com/111471/39348656-4b3328e6-49f8-11e8-83d7-d159aaad9a8a.png)

Due to https://github.com/BrandwatchLtd/axiom/blob/master/packages/axiom-components/src/Context/Context.css#L94 which was overseen when introducing the sub menus.

As discussed @HHogg this could maybe changed to add the border to the top when it follows another content item...

PR #544, opened 2018-05-02
Fix redundant separator after submenu

tomru:
[Deploy Link](https://5ae98e0bc965926325846456--youthful-albattani-41ebb2.netlify.com/docs/packages/axiom-components/dropdown)

As discussed with @HHogg - this PR fixes separator lines being drawn after submenus in Dropdowns.

Also fixed the example for dropdowns, where the submenu was wrapped in an own `DropdownMenu`.

### Before:

image

### After: 

image

### DIFF
--- a/packages/axiom-components/src/Context/Context.css
+++ b/packages/axiom-components/src/Context/Context.css
@@ -96,25 +96,25 @@
   padding-bottom: var(--cmp-context-padding-large);
 }
 
-.ax-context-content:not(:last-child)::after {
+.ax-context-content + .ax-context-content::before {
   content: '';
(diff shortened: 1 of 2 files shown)


pmsorhaindo:
Oh hey Harry! 🍺

HHogg:
No biggy... but the second commit shouldn't be a `fix()` as it's purely related to the styleguide documentation so the message doesn't need to make its way into the package release notes.

tomru:
> Nice 👍 Closes #539 (Do your thing Github)

out of interest, doesn't do the commit message in c84de94 do the same ?

PR #546, opened 2018-05-03
feat(Bar): Add new prop striped

stefanbuck:
Adding a new property `fillMode` to apply a different background texture. [Demo](https://brave-keller-55c6a8.netlify.com/docs/packages/axiom-charts/bars/)

### solid (default)
screen shot 2018-05-08 at 10 42 59

### semstriped
screen shot 2018-05-08 at 10 42 04

### striped
screen shot 2018-05-08 at 10 42 13

### DIFF
--- a/packages/axiom-charts/src/Bar/Bar.css
+++ b/packages/axiom-charts/src/Bar/Bar.css
@@ -28,166 +28,223 @@
 }
 
 .ax-bars__bar-rect {
+  position: relative;
   flex: 1 1 0%;
 
-  &.ax-bars__bar-rect--tiny-clanger { background-color: var(--color-product-tiny-clanger); }
-  &.ax-bars__bar-rect--critical-mass { background-color: var(--color-product-critical-mass); }
-  &.ax-bars__bar-rect--fantastic-voyage { background-color: var(--color-product-fantastic-voyage); }
-  &.ax-bars__bar-rect--paradise-lost { background-color: var(--color-product-paradise-lost); }
-  &.ax-bars__bar-rect--serene-sea { background-color: var(--color-product-serene-sea); }
-  &.ax-bars__bar-rect--electric-dreams { background-color: var(--color-product-electric-dreams); }
-  &.ax-bars__bar-rect--giant-leap { background-color: var(--color-product-giant-leap); }
-  &.ax-bars__bar-rect--moon-lagoon { background-color: var(--color-product-moon-lagoon); }
-  &.ax-bars__bar-rect--space-invader { background-color: var(--color-product-space-invader); }
-  &.ax-bars__bar-rect--terra-form { background-color: var(--color-product-terra-form); }
-  &.ax-bars__bar-rect--primeval-soup { background-color: var(--color-product-primeval-soup); }
-  &.ax-bars__bar-rect--sun-maker { background-color: var(--color-product-sun-maker); }
-  &.ax-bars__bar-rect--new-horizon { background-color: var(--color-product-new-horizon); }
-  &.ax-bars__bar-rect--blast-off { background-color: var(--color-product-blast-off); }
-  &.ax-bars__bar-rect--crash-course { background-color: var(--color-product-crash-course); }
-  &.ax-bars__bar-rect--ground-control { background-color: var(--color-product-ground-control); }
-  &.ax-bars__bar-rect--space-oddity { background-color: var(--color-product-space-oddity); }
-  &.ax-bars__bar-rect--deep-thought { background-color: var(--color-product-deep-thought); }
-  &.ax-bars__bar-rect--luna-dust { background-color: var(--color-product-luna-dust); }
-  &.ax-bars__bar-rect--sentiment-positive { background-color: var(--color-sentiment-positive); }
-  &.ax-bars__bar-rect--sentiment-negative { background-color: var(--color-sentiment-negative); }
-  &.ax-bars__bar-rect--sentiment-neutral { background-color: var(--color-sentiment-neutral); }
-  &.ax-bars__bar-rect--social-twitter { background-color: var(--color-social-twitter); }
-  &.ax-bars__bar-rect--social-facebook { background-color: var(--color-social-facebook); }
-  &.ax-bars__bar-rect--social-instagram { background-color: var(--color-social-instagram); }
+  & .ax-bars__bar--solid,
+  & .ax-bars__bar--striped,
+  & .ax-bars__bar--semistriped {
+    width: 100%;
+    height: 100%;
+  }
+
+  & .ax-bars__bar--solid {
+    background-color: currentColor;
+  }
+
+  & .ax-bars__bar-striped,
+  & .ax-bars__bar--semistriped {
+    position: absolute;
+    top: 0;
+    bottom: 0;
+  }
+
+  & .ax-bars__bar--semistriped {
+    opacity: 0.3;
+    background:
+      repeating-linear-gradient(
+        -45deg,
+        rgb(var(--rgb-ui-white)),
+        rgb(var(--rgb-ui-white)) var(--spacing-grid-size--x1),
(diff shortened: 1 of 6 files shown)


packages/axiom-charts/src/Bar/Bar.js
```diff
@@ -55,13 +55,16 @@ export default class Bar extends Component {
     showLabel: PropTypes.bool,
     /** Explicit thickness of the Bar */
     size: PropTypes.string,
+    /** When true the background will be striped */
+    stripedBackground: PropTypes.bool,
```
stefanbuck: In the PR description I was writing `texture` background instead of `striped` which I think sounds better and fits if we decide to use dotted backgrounds in the future.

packages/axiom-charts/src/Bar/Bar.css
```diff
@@ -240,3 +240,13 @@
...
   }
 }
+
+.ax-bars__bar-rect--striped {
+  background-image: repeating-linear-gradient(
+  -45deg,
+  transparent,
+  transparent 10px,
+  rgba(255, 255, 255, .3) 10px,
+  rgba(255, 255, 255, .3) 20px
```
HHogg: Do these colour values need to reference a swatch colour and the size of the bars to the grid system?

packages/axiom-charts/src/Bar/Bar.css
```diff
@@ -240,3 +240,14 @@
     padding-top: var(--cmp-chart-label-margin);
   }
 }
+
+.ax-bars__bar-rect--striped {
+  background-image:
+    repeating-linear-gradient(
+      -45deg,
+      transparent,
```
HHogg: Should this be a translucent colour of the bars colour to handle the exceeded difference situation? (I guess when the difference is less than value the translucent colour will appear transparent against the bar)

packages/axiom-charts/src/Bar/Bar.css
```diff
@@ -240,3 +240,14 @@
...
+      transparent 0.7rem,
+      rgba(255, 255, 255, 0.3) 0.7rem,
+      rgba(255, 255, 255, 0.3) 1.4rem
```
HHogg: Do these values need to correspond to the swatch colours and the grid system?

packages/axiom-charts/src/Bar/Bar.js
HHogg: `striped` Sounds good. Do you think as we just use `color` (not `backgroundColor`) for the background colour configuration, we should just go with `striped` rather than `stripedBackground` for consistency?

stefanbuck: `striped` works for me

packages/axiom-charts/src/Bar/Bar.css
stefanbuck: As I mentioned before, using 30% white was just the quickest way to achieve this. If we want to be explicit I need a list of all corresponding colour values.

HHogg: Yeah, the transparent white I agree is the way to go, but I expect the opacity from the designs  comes from one of the predefined ones in [here](https://github.com/BrandwatchLtd/axiom/blob/master/packages/axiom-materials/src/opacities.css).

packages/axiom-charts/src/Bar/Bar.css
stefanbuck: Ignore my comment above, I mixed up things, sorry.

In the exceeded cases there will be a bar underneath as well which has also a background colour.

HHogg: No worries :) You should be able to avoid defining a colour for every gradient by using the `currentColor` trick.

packages/stylelint-config-axiom/src/stylelintrc.json
```diff
@@ -53,7 +53,7 @@
     "indentation": 2,
     "length-zero-no-unit": true,
     "max-empty-lines": 1,
-    "max-line-length": 110,
+    "max-line-length": 120,
```
stefanbuck: The previous value was conflicting with changes I made in `Bar.css`. I think it's suitable to tweak the rule rather than updating every other line which makes mutli cursor editing way more easier.

packages/axiom-charts/src/Bar/Bar.js
```diff
@@ -62,6 +72,7 @@ export default class Bar extends Component {
     isHidden: false,
     labelStrong: false,
     minSize: '1rem',
+    fillMode: FILL_MODE_SOLID,
```
stefanbuck: @HHogg I sticked with the `fillMode` property to have full control over the bar. For example, in my design I want to have a centered tooltip on the solid bar, not both bars. 

screen shot 2018-05-04 at 11 06 35

packages/axiom-charts/src/Bar/Bar.css
```diff
@@ -28,166 +28,362 @@
...
+        currentColor calc(var(--cmp-grid-gutter-size-tiny) * 2)
+      );
+  }
+
+  & .ax-bars__bar-striped--transparent {
+    opacity: 1;
+  }
+
+  &.ax-bars__bar-rect--tiny-clanger .ax-bars__bar-filled,
+  &.ax-bars__bar-rect--tiny-clanger .ax-bars__bar-striped { color: var(--color-product-tiny-clanger); }
```
HHogg: I think you could probably just go with 

```css
.ax-bars__bar-rect--tiny-clanger { color: var(--color-product-tiny-clanger); }
```

As `.ax-bars__bar-filled` and `.ax-bars__bar-striped` are descendants the `currentColor` will work from the inherited value. The same with the `:hover` and `:active` selectors below. 

This will likely drop those lines lengths too, so the linting rule could be kept at it's original value :)

packages/axiom-charts/src/Bar/Bar.js
HHogg: Ok that makes sense 👍

packages/axiom-charts/src/Bar/Bar.css
```diff
@@ -28,166 +28,362 @@
...
+    top: 0;
+    bottom: 0;
+    opacity: 0.3;
+    background:
+      repeating-linear-gradient(
+        -45deg,
+        rgb(var(--rgb-ui-white)),
+        rgb(var(--rgb-ui-white)) var(--cmp-grid-gutter-size-tiny),
+        currentColor var(--cmp-grid-gutter-size-tiny),
+        currentColor calc(var(--cmp-grid-gutter-size-tiny) * 2)
```
HHogg: Using the `--cmp-grid-gutter` variables won't preprocess because those variables belong to the Grid component. Only variables from the materials package should be used across other files. 

**tldr;** Instead of `--cmp-grid-gutter-size-tiny` use `--spacing-grid-size--x2`

packages/axiom-charts/src/Bar/Bar.css
```diff
@@ -28,166 +28,362 @@
...
+      );
+  }
+
+  & .ax-bars__bar-striped--transparent {
+    opacity: 1;
```
HHogg: I'm not 100% sure this is correct. Looking at the standalone striped designs, they still remain at 0.3

screen shot 2018-05-04 at 10 50 36

packages/axiom-charts/src/Bar/Bar.css
stefanbuck: Done see fixup commit 71703cb8

packages/axiom-charts/src/Bar/Bar.css
stefanbuck: Fixed with fixup 47b5cf42

stefanbuck:
Hi @HHogg, I addressed each of your comments in a separate fixup commit. The last fixup renames some classNames so that those are inline with the props names.

PR #549, opened 2018-05-07
feat(BarChart): Added tooltips to BarChart bars and benchmark line

t-sauer:


t-sauer:
Sure, here you go: https://naughty-pasteur-6d413a.netlify.com/docs/packages/axiom-charts/bar-chart

packages/axiom-charts/src/BarChart/BarChart.css
```diff
@@ -29,7 +29,20 @@
   position: absolute;
   top: 0;
   bottom: 0;
-  border-right: var(--cmp-benchmark-line-width) solid var(--color-theme-text--subtle);
+  margin: 0 -0.25rem;
```
HHogg: It might be better to use a `transform: translateX(-50%)` here so that if the padding below changes the offset position will remain the same too.

packages/axiom-charts/src/BarChart/BarChart.css
```diff
@@ -29,7 +29,20 @@
...
+  padding: 0 0.25rem;
```
HHogg: Should this relate to the grid system (`var(--spacing-grid-size--x1)`)? Also from the designs it looks like there should be vertical padding also.

t-sauer:
I added the grey border thing and adjusted some other minor things. Would be nice if you  could have another look @HHogg

t-sauer:
Changed the border around the benchmark line to only appear on hover.

PR #550, opened 2018-05-07
fix(BarChart): Do not highlight all bars with same color if there is only one bar per row

t-sauer:
Right now if you only use one bar per row, and you hover a row, all rows with the same color are highlighted which in that case is not the desired result (we just want to highlight the actual hovered row).

### DIFF
--- a/packages/axiom-charts/src/BarChart/BarChart.js
+++ b/packages/axiom-charts/src/BarChart/BarChart.js
@@ -60,6 +60,8 @@ export default class BarChart extends Component {
     showBarLabel: PropTypes.bool,
     /** Control for toggling visibility of the key */
     showKey: PropTypes.bool,
+    /** If set to true each color will be handled individually when hovering */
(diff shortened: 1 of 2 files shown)


HHogg:
@t-sauer I'm sure this is fine, but could you provide a gif of the previous behaviour and this fixed behaviour? (Just want to make sure it's not changing the expected behaviour for other scenarios).

t-sauer:
Before:
You can see, that when I hover over "Fashion", "Business" is also highlighted and the other way around. Same goes for "Automotive" and "Technology".
barchart-hover-before
After:
Only the hovered row is highlighted.
barchart-hover-after

The change only takes effect if there is one bar per row, if there are multiple bars per row, everything is unchanged from the previous behaviour (i.e. bars in other rows with the same color are still highlighted).

t-sauer:
Changed it to a property instead of doing it implicit to not screw with the current behaviour.

PR #551, opened 2018-05-08
feat(BarChart) Add option to fade out a bar

siggysamson:
Adding an `isFaded` property to the `data` array enables the whole row to be faded/hidden. Initially I was thinking to pass the `isFaded` property to `BarChartBars` but what we want to achieve is to completely hide the bars and fade out the label.

screen shot 2018-05-08 at 09 35 09

**Note**: Needs removal of the modified `chartData#dotPlotData`

### DIFF
--- a/packages/axiom-charts/src/BarChart/BarChart.js
+++ b/packages/axiom-charts/src/BarChart/BarChart.js
@@ -168,36 +168,39 @@ export default class BarChart extends Component {
             xAxisLabels={ xAxisLabels }
             zoom={ zoom }
             zoomTo={ zoom ? zoomTo : undefined }>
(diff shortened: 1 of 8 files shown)


thatguynamedandy:
Hey @siggysamson I'm not too sure on the use case in the app, but I just wanted to check, that we are ok that the height jumps about as the label switches from faded to not faded?

siggysamson:
Good point! It's about hiding a bar when clicking on the label (in order to adjust the scale if there is an _extrem_ value). I haven't thought about the height yet though.

I'll close this PR for now as I've further discussed the approach and I'll make this part of `BarChart` with some further changes on how `xAxisLabels` works

PR #553, opened 2018-05-11
chore(BarChart): Added tests for BarChart component

siggysamson:
🤷‍♂️ 

I think the `BarChart` component really should have some tests as we are adding more and more functionality to it.
It was really hard to figure out how many tests to write and how extensive they should be but here we are 😄 

As I was writing tests I:
1) Implemented a custom propTypes validation for `lower` as I didn't want to bother writing tests for what happens when `lower` is smaller than `upper` and
2) I separated the zoom calculation into `utils` so that I can write tests for it separately.

I guess some tests are still better than no tests but it was really hard to figure out what's worth writing a test for. I actually might have got lost and forgot to test some properties so if you find anything while reviewing let me know

### DIFF
--- a/packages/axiom-charts/src/BarChart/BarChart.js
+++ b/packages/axiom-charts/src/BarChart/BarChart.js
@@ -13,7 +13,7 @@ import ChartTableRows from '../ChartTable/ChartTableRows';
 import ChartTableVisual from '../ChartTable/ChartTableVisual';
 import DataPoint from '../DataPoint/DataPoint';
 import DataPoints from '../DataPoint/DataPoints';
-import { formatData, flattenValues, hasMultipleValues } from './utils';
+import { formatData, flattenValues, hasMultipleValues, calculateZoom } from './utils';
 import './BarChart.css';
 
 export default class BarChart extends Component {
@@ -53,7 +53,19 @@ export default class BarChart extends Component {
     /** The width of the yAxis labels columns */
     labelColumnWidth: PropTypes.string.isRequired,
     /** Lower value of the data displayed on the chart */
-    lower: PropTypes.number,
+    lower(props, propName, componentName) {
+      if (!props.lower) {
+        return;
+      }
+
+      if (typeof props.lower !== 'number') {
+        return new Error('Invalid prop `lower` supplied to ' + componentName + '. Validation failed.');
+      }
+
+      if (props.upper && props.upper < props.lower) {
+        return new Error('Invalid prop `lower` supplied to ' + componentName + '. `lower` needs to be smaller than `upper`');
+      }
+    },
     /** Spacing applied between Bar groups */
     rowSpace: PropTypes.oneOf(['x1', 'x2', 'x3']),
     /** Option to always show the label next to bars, as opposed to on mouse over  */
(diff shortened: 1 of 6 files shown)


packages/axiom-charts/src/BarChart/BarChart.js
```diff
@@ -53,7 +53,19 @@ export default class BarChart extends Component {
...
+      if (!props.lower) {
+        return;
+      }
+
+      if (typeof props.lower !== 'number') {
+        return new Error('Invalid prop `lower` supplied to ' + componentName + '. Validation failed.');
+      }
+
+      if (props.upper && props.upper < props.lower) {
+        return new Error('Invalid prop `lower` supplied to ' + componentName + '. `lower` needs to be smaller than `upper`');
```
HHogg: Should PropTypes be used for value validation and how far should it be taken is a pretty hot/common topic in React 😄 There are quite a few cases in Axiom and in general react components where value validation with custom propTypes would be useful to stop silly mistakes, but it adds even more maintenance, where those mistakes actually exist in the implements. 

I'm really of the opinion that propTypes should be kept for type validation, and the components should expect they are being used in the correct way, but it's an opinion and open to others. What's your thoughts @thatguynamedandy?

For this specific use case, maybe we could add a helpful swap check instead so lower and upper are the correct way around?

packages/axiom-charts/src/BarChart/barChartFixture.js
```diff
@@ -0,0 +1,57 @@
+export const getXAxisLabels = () => ['0', '10', '20', '30', '40', '50', '60', '70', '80', '90', '100'];
```
HHogg: The build process currently only excludes test files form the final published packages, so with how this current stands this fixture file would end up on NPM. 

I guess there is two options:
* Could move these fixtures into the test final ([similar to the DotPlot](https://github.com/BrandwatchLtd/axiom/blob/master/packages/axiom-charts/src/DotPlotChart/utils.test.js#L19))
* Change the build process to exclude `.fixture.js` files?

packages/axiom-charts/src/BarChart/utils.test.js
```diff
@@ -133,4 +133,112 @@ describe('BarChart (utils)', () => {
...
+          finalLower: 5,
+        })).toBe(33.33333333333333);
+      });
+
+      test('uses dataUpper as the base for the zoom calculation when dataUpper is larger than finalUpper', () => {
+        expect(calculateZoom({
+          dataUpper: 20,
+          finalUpper: 10,
+          finalLower: 5,
+        })).toBe(300);
```
HHogg: I haven't dipped into the zooming stuff in a while but I thought what was produced was always a percentage? _(0-100)_

HHogg: I'm not actually sure, a lot of these tests are needed or the util needs to change a bit... currently the Charts are performing the min/max checks so dataUpper and finalUpper values will never be in this situation. Maybe the min/max checks need moving to the util and instead it takes the data bounds and the explicit bounds from the props? /shrug

lpoulter:
@siggysamson whats the status of this PR?

siggysamson:
I didn't have time to finish this up and I'd prefer to just close it and I'll reopen again when I get to it!

siggysamson:
Thanks for the reminder about this PR 🙂

PR #554, opened 2018-05-14
feat(BarChart): Add new prop showDifferenceArea

stefanbuck:
Adds a new  prop `showDifferenceArea` to the BarChart component. Setting this property to `true` (`false` by default) will cause the BarChart to add a striped bar between the bar and benchmark line (see attached screenshot). 

[Demo](https://brave-keller-55c6a8.netlify.com/docs/packages/axiom-charts/bar-chart)

screen shot 2018-05-14 at 11 46 35

### DIFF
--- a/packages/axiom-charts/src/BarChart/BarChart.js
+++ b/packages/axiom-charts/src/BarChart/BarChart.js
@@ -64,6 +64,8 @@ export default class BarChart extends Component {
     singleSelect: PropTypes.bool,
     /** Thickness of the bars */
     size: PropTypes.string,
+    /**  Control the appearance of the bar */
+    stretch: PropTypes.bool,
     /** Upper value of the data displayed on the chart */
     upper: PropTypes.number,
     /**
@@ -80,6 +82,7 @@ export default class BarChart extends Component {
   static defaultProps = {
     rowSpace: 'x2',
     showKey: true,
+    stretch: false,
   };
 
 
@@ -143,6 +146,7 @@ export default class BarChart extends Component {
       showKey,
       singleSelect,
       size,
+      stretch,
       upper = dataUpper,
       xAxisLabels,
       zoom,
@@ -202,6 +206,7 @@ export default class BarChart extends Component {
                     showBarLabel={ showBarLabel }
                     singleSelect={ singleSelect }
                     size={ size }
+                    stretch={ stretch }
                     upper={ finalUpper }
                     values={ values } />
               </ChartTableVisual>
--- a/packages/axiom-charts/src/BarChart/BarChart.js
+++ b/packages/axiom-charts/src/BarChart/BarChart.js
@@ -58,14 +58,14 @@ export default class BarChart extends Component {
     rowSpace: PropTypes.oneOf(['x1', 'x2', 'x3']),
     /** Option to always show the label next to bars, as opposed to on mouse over  */
     showBarLabel: PropTypes.bool,
+    /**  Control the appearance of the bar */
+    showDifferenceArea: PropTypes.bool,
     /** Control for toggling visibility of the key */
     showKey: PropTypes.bool,
     /** If set to true each color will be handled individually when hovering */
     singleSelect: PropTypes.bool,
     /** Thickness of the bars */
     size: PropTypes.string,
-    /**  Control the appearance of the bar */
-    stretch: PropTypes.bool,
     /** Upper value of the data displayed on the chart */
     upper: PropTypes.number,
     /**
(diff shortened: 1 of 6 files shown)


packages/axiom-charts/src/BarChart/BarChart.js
```diff
@@ -62,6 +62,8 @@ export default class BarChart extends Component {
     showKey: PropTypes.bool,
     /** Thickness of the bars */
     size: PropTypes.string,
+    /**  Control the appearance of the bar */
+    stretch: PropTypes.bool,
```
HHogg: I'm not too keen on the name `stretch`, it's a bit generic on what it is doing. It should probably be something to indicate that the difference area is being highlighted. Something like `showDifferenceArea` perhaps, but also open to other suggestions.

packages/axiom-charts/src/BarChart/BarChartBars.js
```diff
@@ -27,10 +28,15 @@ export default class BarChartBars extends Component {
     onMouseLeave: PropTypes.func.isRequired,
     showBarLabel: PropTypes.bool,
     size: PropTypes.string,
+    stretch: PropTypes.bool,
     upper: PropTypes.number,
     values: PropTypes.array.isRequired,
   };
 
+  static defaultProps = {
+    stretch: false,
```
HHogg: Not an issue, but you could avoid having false default prop here, as it's an internal only component and will always get a value from the BarChart as the default is set there.

packages/axiom-charts/src/BarChart/BarChartBars.js
```diff
@@ -87,16 +121,7 @@ export default class BarChartBars extends Component {
...
-                      color={ color }
-                      data-ax-at={ atIds.BarChart.bar }
-                      isFaded={ isFaded }
-                      isHidden={ hideBars && isFaded }
-                      onMouseEnter={ () => onMouseEnter(color) }
-                      onMouseLeave={ onMouseLeave }
-                      percent={ percent }
-                      showLabel={ false }
-                      size={ size } />
+                  { combineBar || bar }
```
HHogg: We've tended to keep templates together as best as possible so the render functions can be read from top to bottom. 

Something like 

```js
{ stretch ? (
  <CombinedBar
      benchmarkValue={ benchmarkValue }
      color={ color }
      data-ax-at={ atIds.BarChart.bar }
      isFaded={ isFaded }
      isHidden={ hideBars && isFaded }
      onMouseEnter={ () => onMouseEnter(color) }
      onMouseLeave={ onMouseLeave }
      percent={ percent }
      showLabel={ false }
      size={ size } />
) : (
  <Bar
      color={ color }
      data-ax-at={ atIds.BarChart.bar }
      isFaded={ isFaded }
      isHidden={ hideBars && isFaded }
      onMouseEnter={ () => onMouseEnter(color) }
      onMouseLeave={ onMouseLeave }
      percent={ percent }
      showLabel={ false }
      size={ size } />
) }
```

Or as everything is quite similar between the Bars, it might be nicer to do something like...

```jsx
const FinalBar = stretch ? CombinedBar : Bar;

  <FinalBar
     benchmarkValue={ stretch ? benchmarkValue : null }
      color={ color }
      data-ax-at={ atIds.BarChart.bar }
      isFaded={ isFaded }
      isHidden={ hideBars && isFaded }
      onMouseEnter={ () => onMouseEnter(color) }
      onMouseLeave={ onMouseLeave }
      percent={ percent }
      showLabel={ false }
      size={ size } />

```

packages/axiom-charts/src/BarChart/CombinedBar.css
```diff
@@ -0,0 +1,5 @@
+.ax-bar-chart__combined-bar_diff {
```
HHogg: `combined-bar_diff` The single underscore I don't think is valid BEM... maybe just `combined-bar-diff`?

stefanbuck: ✅Addressed in the latest force push

packages/axiom-charts/src/BarChart/BarChartBars.js
stefanbuck: ✅ I went with the second approach.

packages/axiom-charts/src/BarChart/BarChartBars.js
stefanbuck: I see, good point

packages/axiom-charts/src/BarChart/BarChart.js
stefanbuck: ✅ I renamed the prop to `showDifferenceArea`

package.json
```diff
@@ -1,4 +1,6 @@
 {
+  "name": "stefanbuck-axiom-fork",
```
HHogg: 😬 Accidentally left in?

PR #556, opened 2018-05-14
fix(Tab): Only add as a title if the title is a string

HHogg:


packages/axiom-components/src/Tabset/Tab.js
```diff
@@ -34,7 +34,7 @@ export default class Tab extends Component {
           className={ classes }
           textSize="small"
           textStrong
-          title={ title }>
+          title={ typeof title === 'string' ? title : null }>
```
t-sauer: Couldn't the title be a number in theory (in which case it also would makes sense to add a title)?

HHogg: Yes good point, I'll update for the reverse check of a react element 👍

PR #558, opened 2018-05-15
chore(Bar): Omit benchmarkValue when passing props to Base component

t-sauer:
On the current master `benchmarkValue` is passed down to the Base component which results in a react warning (`React does not recognize the `benchmarkValue` prop on a DOM element...`).

### DIFF
--- a/packages/axiom-charts/src/Bar/Bar.js
+++ b/packages/axiom-charts/src/Bar/Bar.js
@@ -1,6 +1,7 @@
 import PropTypes from 'prop-types';
 import React, { Component } from 'react';
 import classnames from 'classnames';
+import omit from 'lodash.omit';
 import { Base, Small } from '@brandwatch/axiom-components';
 import './Bar.css';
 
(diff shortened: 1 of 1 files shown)


packages/axiom-charts/src/Bar/Bar.js
```diff
@@ -127,7 +129,7 @@ export default class Bar extends Component {
     };
 
     return (
-      <Base { ...rest }
+      <Base { ...omit(rest, ['benchmarkValue']) }
```
HHogg: I think this needs to happen further up the chain inside the BarChart component ... [around here](https://github.com/BrandwatchLtd/axiom/blob/master/packages/axiom-charts/src/BarChart/BarChartBars.js#L107)

t-sauer: Yeah that's where it is coming from, but if `FinalBar === CombinedBar` we want `benchmarkValue` to be passed down. Is there a more elegant way to solve this?

HHogg: It's not nice, but a quick way for now could be to conditionally omit and spread it, like...
```jsx
const finalBarProps = omit({ benchmarkValue }, [showDifferenceArea && 'benchmarkValue']);
<FinalBar { ...finalBarProps } />
``` 
...but this would at least keep it scoped to that module. 

Having a deeper look I think the issue might be to do with the structure of CombinedBar and it should use the original Bar component and just handle the difference area.

HHogg:
Closing as is resolved in #571

PR #559, opened 2018-05-15
feat(BarChart): Added rowsConcealable prop that activates show/hide controls for barchart rows

t-sauer:
Deployed here: https://5afaf2fcdd6a542479baef3f--naughty-pasteur-6d413a.netlify.com/docs/packages/axiom-charts/bar-chart
You have to activate `rowsConcealable `

### DIFF
--- a/packages/axiom-charts/src/BarChart/BarChart.js
+++ b/packages/axiom-charts/src/BarChart/BarChart.js
@@ -59,6 +59,7 @@ export default class BarChart extends Component {
     data: PropTypes.arrayOf(PropTypes.shape({
       label: PropTypes.node.isRequired,
       values: PropTypes.object.isRequired,
+      hidden: PropTypes.bool,
     })).isRequired,
     /** The description given to the expand button */
     expandButtonSuffix: PropTypes.string,
@@ -68,6 +69,12 @@ export default class BarChart extends Component {
     labelColumnWidth: PropTypes.string.isRequired,
     /** Lower value of the data displayed on the chart */
     lower: PropTypes.number,
+    /**
+     * If a function is given, rows will have a button to hide them.
+     * If this button is pressed, the passed in function will be called.
+     * In addition, rows with a property `hidden` set to true will be hidden.
+     */
+    onToggleRowVisibility: PropTypes.func,
     /** Spacing applied between Bar groups */
     rowSpace: PropTypes.oneOf(['x1', 'x2', 'x3']),
     /** Option to always show the label next to bars, as opposed to on mouse over  */
@@ -94,6 +101,7 @@ export default class BarChart extends Component {
 
   static defaultProps = {
     isBenchmarkLineFadable: true,
+    onToggleRowVisibility: undefined,
     rowSpace: 'x2',
     showKey: true,
     showDifferenceArea: false,
@@ -140,8 +148,7 @@ export default class BarChart extends Component {
   }
 
   render() {
-    const flatValues = flattenValues(this.props.data);
-
+    const flatValues = flattenValues(this.props.data.filter(({ hidden }) => !hidden));
     const dataLower = Math.min(...flatValues);
     const dataUpper = Math.max(...flatValues);
 
@@ -159,6 +166,7 @@ export default class BarChart extends Component {
       isBenchmarkLineFadable,
       labelColumnWidth,
       lower = dataLower,
+      onToggleRowVisibility,
       rowSpace,
       showBarLabel,
       showKey,
@@ -195,11 +203,14 @@ export default class BarChart extends Component {
             xAxisLabels={ finalXAxisLabels }
             zoom={ zoom }
             zoomTo={ zoom ? zoomTo : undefined }>
-          { formattedData.map(({ values, label, benchmark }, index) =>
+          { formattedData.map(({ values, label, benchmark, hidden }, index) =>
             <ChartTableRow
+                cloakContainer={ !hidden }
                 hover={ index === selectedIndex }
                 key={ index }>
               <ChartTableLabel
+                  disabled={ hidden }
+                  onToggleRowVisibility={ () => onToggleRowVisibility(data[index], index) }
                   textStrong={ index === selectedIndex }
                   width={ labelColumnWidth }>
                 { label }
@@ -218,6 +229,7 @@ export default class BarChart extends Component {
                     hoverColor={ selectedColor }
                     hoverIndex={ selectedIndex }
                     index={ index }
+                    isHidden={ onToggleRowVisibility ? hidden : false }
                     isHovered={ index === selectedIndex }
                     label={ label }
                     lower={ finalLower }
--- a/packages/axiom-charts/src/BarChart/BarChartBars.js
+++ b/packages/axiom-charts/src/BarChart/BarChartBars.js
@@ -2,7 +2,7 @@ import PropTypes from 'prop-types';
 import React, { Component } from 'react';
 import classnames from 'classnames';
 import atIds from '@brandwatch/axiom-automation-testing/ids';
-import { Small } from '@brandwatch/axiom-components';
+import { Base, Small } from '@brandwatch/axiom-components';
 import Bar from '../Bar/Bar';
 import Bars from '../Bar/Bars';
 import CombinedBar from './CombinedBar';
@@ -23,6 +23,7 @@ export default class BarChartBars extends Component {
     hoverColor: PropTypes.string,
     hoverIndex: PropTypes.number,
     index: PropTypes.number,
+    isHidden: PropTypes.bool,
     isHovered: PropTypes.bool.isRequired,
     label: PropTypes.node.isRequired,
     lower: PropTypes.number,
@@ -51,6 +52,7 @@ export default class BarChartBars extends Component {
       hideBars,
       hoverColor,
       hoverIndex,
+      isHidden,
       isHovered,
       index,
       label,
@@ -77,7 +79,7 @@ export default class BarChartBars extends Component {
     }
 
     return (
-      <div className={ classes }>
+      <Base className={ classes } cloak={ isHidden }>
         <Bars direction="right">
           { values.map(({ color, value }) => {
             const isFaded = singleSelect
@@ -140,7 +142,7 @@ export default class BarChartBars extends Component {
                 value={ benchmarkValue } />
           </div>
         ) }
-      </div>
+      </Base>
     );
   }
 }
--- a/site/components/Documentation/Resources/Charts/BarChart.js
+++ b/site/components/Documentation/Resources/Charts/BarChart.js
@@ -9,7 +9,27 @@ import { dotPlotKey, dotPlotData } from './chartData';
 import DropdownContext from './DropdownContext';
 
 export default class Documentation extends Component {
+  constructor(props) {
+    super(props);
+    this.state = {
+      hiddenRows: {},
+    };
+  }
+
+  handleToggleRowVisibility(index) {
+    const { hiddenRows } = this.state;
+
+    this.setState({
+      hiddenRows: {
+        ...hiddenRows,
+        [index]: !hiddenRows[index],
+      },
+    });
+  }
+
   render() {
+    const { hiddenRows } = this.state;
+
     return (
       <DocumentationContent>
         <DocumentationShowCase>
@@ -21,10 +41,11 @@ export default class Documentation extends Component {
               chartKey={ dotPlotKey }
               chartKeyBenchmarkLabel="Benchmark"
               collapsedVisibleRowCount={ 4 }
-              data={ dotPlotData }
+              data={ dotPlotData.map((_, i) => ({ ..._, hidden: hiddenRows[i] })) }
               expandButtonSuffix="Categories"
               labelColumnWidth="11rem"
               lower={ 0 }
+              onToggleRowVisibility={ (a, b, data, i) => this.handleToggleRowVisibility(i) }
               upper={ 100 }
               xAxisLabels={ () => [ '0', '10', '20', '30', '40', '50', '60', '70', '80', '90', '100'] }/>
         </DocumentationShowCase>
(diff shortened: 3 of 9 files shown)


packages/axiom-charts/src/BarChart/BarChart.css
```diff
@@ -50,3 +50,24 @@
 .ax-bar-chart__bar-label--hidden {
   opacity: 0;
 }
+
+.ax-bar-chart__conceal-button {
+  opacity: 0;
+  transition: opacity var(--transition-time-base) var(--transition-function);
```
HHogg: As discussed, you can shortcut all of this css with the `cloak` prop and `textColor` prop

packages/axiom-charts/src/BarChart/BarChart.js
```diff
@@ -89,6 +96,7 @@ export default class BarChart extends Component {
   constructor(props) {
     super(props);
     this.state = {
+      hiddenRows: [],
```
HHogg: How should this component deal with clearing this when the data changes? Working it out might get a bit tricky depending on how this component is getting used (for example a simple `this.props.data !== nextProps.data` check will always be `true` if a new array is always being give).

I'm wondering if the state for this should be held by the component that is rendering the BarChart, and an API like `onToggleRowVisibility()` needs to be used instead?

packages/axiom-components/src/Icon/IconButton.js
```diff
@@ -12,6 +12,7 @@ const sizeMap = {
 
 export default class IconButton extends Component {
   static propTypes = {
+    className: PropTypes.string,
```
HHogg: All of these className props should be avoidable with using the `cloak` props.

packages/axiom-components/src/Button/Button.js
```diff
@@ -12,6 +12,7 @@ export default class Button extends Component {
     active: PropTypes.bool,
     /** Content inserted into the Button */
     children: PropTypes.node.isRequired,
+    className: PropTypes.string,
```
HHogg: All of these className props should be avoidable with using the `cloak` props.

packages/axiom-charts/src/ChartTable/ChartTableRow.js
```diff
@@ -6,12 +6,13 @@ import { Base } from '@brandwatch/axiom-components';
 export default class ChartTableRow extends Component {
   static propTypes = {
     children: PropTypes.node.isRequired,
+    className: PropTypes.string,
```
HHogg: All of these className props should be avoidable with using the `cloak` props.

packages/axiom-charts/src/BarChart/BarChartBars.js
```diff
@@ -116,15 +119,15 @@ export default class BarChartBars extends Component {
                       size={ size } />
                 </ChartContext>
 
-                <div className={ labelClasses } style={ labelStyle }>
+                { !isConcealed && <div className={ labelClasses } style={ labelStyle }>
```
HHogg: If you change the `<div className="ax-bar-chart__bar-container" key={ color }>` to a Base component, you can set `cloak` once for the whole container.

packages/axiom-charts/src/BarChart/BarChart.js
t-sauer: I think I initially talked about going that route with @siggysamson but I thought the decision was to keep it internally?

HHogg: As the data changes though this will need to be cleared? 

Working out when `data` is actually new data or just a new array being created at render will get tricky/impossible, so I think keeping the state for this internally would cause it to continuous clearing on a render.

t-sauer:
I changed this PR so that the handling of which rows are hidden are handled by the consumer of the BarChart. I also changed all custom class names and used the cloak wherever possible.

packages/axiom-charts/src/BarChart/BarChart.js
```diff
@@ -195,14 +204,24 @@ export default class BarChart extends Component {
...
+                }) }
                 hover={ index === selectedIndex }
                 key={ index }>
               <ChartTableLabel
                   textStrong={ index === selectedIndex }
                   width={ labelColumnWidth }>
-                { label }
+                <Grid horizontalGutters="tiny" verticalAlign="middle" verticalGutters={ false }>
+                  <GridCell className="ax-bar-chart__row-label">
+                    <Text onClick={ () => onToggleRowVisibility && onToggleRowVisibility(data[index]) } textColor={ onToggleRowVisibility && hidden && 'disabled' }>{ label }</Text>
```
HHogg: Can the `onClick` be added to the Grid so it's only defined once?

packages/axiom-charts/src/BarChart/BarChart.js
```diff
@@ -195,14 +204,24 @@ export default class BarChart extends Component {
...
+                  </GridCell>
+                  { onToggleRowVisibility && <GridCell shrink>
+                    <IconButton className="ax-bar-chart__conceal-button" name="preview" onClick={ () => onToggleRowVisibility(data[index]) } size="small" style="quaternary" />
```
HHogg: Same here the className on IconButton won't be being applied.

HHogg: Is IconButton the correct component to use here? It looks like from the designs that it's a Icon wrapped in a Link?

packages/axiom-charts/src/BarChart/BarChart.css
```diff
@@ -65,6 +69,28 @@
...
+.ax-bar-chart__bar-container:not(.ax-cloak--visible) .ax-bar-chart__combined-bar-diff,
+.ax-bar-chart__bar-container:not(.ax-cloak--visible) .ax-bar-chart__bar-label {
+  transition: left 0s var(--transition-function) var(--transition-time-base);
+}
+
 .ax-bar-chart__bar-label--hidden {
   opacity: 0;
 }
+
+.ax-bar-chart__conceal-button {
```
HHogg: This class isn't being applied anywhere. So can it be removed?

HHogg: Same with the other blocks of CSS below.

packages/axiom-charts/src/BarChart/BarChart.js
```diff
@@ -195,14 +204,24 @@ export default class BarChart extends Component {
             xAxisLabels={ finalXAxisLabels }
             zoom={ zoom }
             zoomTo={ zoom ? zoomTo : undefined }>
-          { formattedData.map(({ values, label, benchmark }, index) =>
+          { formattedData.map(({ values, label, benchmark, hidden }, index) =>
             <ChartTableRow
+                className={ classnames('ax-bar-chart__table-row', {
+                  'ax-bar-chart__table-row--concealed': onToggleRowVisibility && hidden,
```
HHogg: It doesn't look like this is needed as `ax-bar-chart__table-row--concealed` is being used to target `ax-bar-chart__conceal-button` which isn't being applied.

packages/axiom-charts/src/BarChart/BarChart.js
```diff
@@ -195,14 +204,24 @@ export default class BarChart extends Component {
...
+                }) }
                 hover={ index === selectedIndex }
                 key={ index }>
               <ChartTableLabel
                   textStrong={ index === selectedIndex }
                   width={ labelColumnWidth }>
-                { label }
+                <Grid horizontalGutters="tiny" verticalAlign="middle" verticalGutters={ false }>
```
HHogg: You shouldn't need `horizontalGutters`  or `verticalGutters={ false }` ... vertical gutters only gets applied when sibling cells wrap onto other lines, which we don't want here.

`gutters="tiny" responsive={ false }` Would be better here.

site/components/Documentation/Resources/Charts/BarChart.js
```diff
@@ -8,6 +8,11 @@ import {
 import { dotPlotKey, dotPlotData } from './chartData';
 import DropdownContext from './DropdownContext';
 
+function onToggleRowVisibility(setValue, getValue, changedRow) {
```
stefanbuck: @HHogg ThomasS is off sick this week and I'll continue working on this. I just integrated his latest changes in Analytics and noticed that he relays on `getValue` and `changedRow` which are injected just one the documentation page. 

Maybe I miss something how this should work, but to me it looks like I have to implement this `onToggleRowVisibility` to make it working.

packages/axiom-charts/src/BarChart/BarChart.css
stefanbuck: I address your comment in a8b4e240, which introduced two other issues

1. Using `textColor` prevents the hover color change so it stays grey. 


2. In the documentation `cloak` is using a css transition to toggle the visibly. For some reason I don't see this transition.

packages/axiom-charts/src/ChartTable/ChartTableRow.js
stefanbuck: I revert those changes with af59e987

packages/axiom-charts/src/BarChart/BarChart.css
stefanbuck: My first guess is that React does a re-render because the prop `isConcealed` changes his value.

stefanbuck: Maybe we can't use `cloak` here which is the reason why @t-sauer didn't replaced it there. His last comment says 

>  I also changed all custom class names and used the cloak wherever possible.

stefanbuck: If this is true, the only issue left is how to make this working in Analytics. See my other comment https://github.com/BrandwatchLtd/axiom/pull/559#discussion_r190529038

packages/axiom-charts/src/BarChart/BarChart.js
```diff
@@ -195,45 +205,60 @@ export default class BarChart extends Component {
...
-                    upper={ finalUpper }
-                    values={ values } />
-              </ChartTableVisual>
-            </ChartTableRow>
-          ) }
+          { formattedData.map(({ values, label, benchmark, hidden }, index) => {
+            const Row = hidden ? 'div' : CloakContainer;
+
+            return (<Row key={ index }>
+              <ChartTableRow hover={ index === selectedIndex }>
```
HHogg: The `const Row = hidden ? 'div' : CloakContainer;` isn't needed you can just give `ChartTableRow` the `cloakContainer` prop.

packages/axiom-charts/src/BarChart/BarChart.js
```diff
@@ -195,45 +205,60 @@ export default class BarChart extends Component {
...
+                <ChartTableLabel
+                    textStrong={ index === selectedIndex }
+                    width={ labelColumnWidth }>
+                  <Grid gutters="tiny" onClick={ () => onToggleRowVisibility && onToggleRowVisibility(data[index]) } responsive={ false } verticalAlign="middle">
+                    <GridCell>
+                      <Text textColor={ onToggleRowVisibility && hidden && 'disabled' }>{ label }</Text>
+                    </GridCell>
+                    { onToggleRowVisibility &&
```
HHogg: All of the template change for inside `ChartTableLabel` should probably live in the `ChartTableLabel` component. That way the same behaviour will be available on DotPlots too.

packages/axiom-charts/src/BarChart/BarChartBars.js
```diff
@@ -124,21 +126,22 @@ export default class BarChartBars extends Component {
...
                 </div>
-              </div>
+              </Base>
             );
           }) }
         </Bars>
 
         { benchmarkValue !== undefined && (
-          <div className="ax-bar-chart__benchmark-line-container">
+          <Base className="ax-bar-chart__benchmark-line-container">
```
HHogg: Is this change needed?

packages/axiom-charts/src/BarChart/BarChartBars.js
```diff
@@ -98,7 +100,7 @@ export default class BarChartBars extends Component {
             const FinalBar = showDifferenceArea ? CombinedBar : Bar;
 
             return (
-              <div className="ax-bar-chart__bar-container" key={ color }>
+              <Base className="ax-bar-chart__bar-container" cloak={ isConcealed } key={ color }>
```
HHogg: How come the bar container and benchmark line are being cloak separately as opposed to cloaking this entire component at the top?

packages/axiom-charts/src/BarChart/BarChart.js
```diff
@@ -195,45 +205,60 @@ export default class BarChart extends Component {
...
-          ) }
+          { formattedData.map(({ values, label, benchmark, hidden }, index) => {
+            const Row = hidden ? 'div' : CloakContainer;
+
+            return (<Row key={ index }>
+              <ChartTableRow hover={ index === selectedIndex }>
+                <ChartTableLabel
+                    textStrong={ index === selectedIndex }
+                    width={ labelColumnWidth }>
+                  <Grid gutters="tiny" onClick={ () => onToggleRowVisibility && onToggleRowVisibility(data[index]) } responsive={ false } verticalAlign="middle">
```
HHogg: Would it be enough just to be giving the row index here? The row data seems unnecessary and possibly harder when using this feature. See comment below in the demo component.

site/components/Documentation/Resources/Charts/BarChart.js
```diff
@@ -8,6 +8,11 @@ import {
...
+  changedRow.hidden = !changedRow.hidden;
```
HHogg: Yeah this doesn't look right, we shouldn't be mutating the data fixture. This would be better storing this in state, similar to how it would be done inside an application (local or redux) like ... 

```jsx
const { hiddenRows } = this.state;

const handleToggleRowVisibility = (i) => 
  this.setState({ 
    hiddenRows: {
      ...hiddenRows,
      [i]: !hiddenRows[i],
    },    
  });
 
<BarChart
  data={ data.map((_, i) => ({ ..._, hidden: hiddenRows[i] })) }
  onToggleRowVisibility={ (a, b, i) => handleToggleRowVisibility(i) } />
```

PR #561, opened 2018-05-16
fix(BarChart): vertically center label when size is specified

tomru:
When using `size` the label next to the bar is not vertically centered anymore:

#### Before
image

Set height to `100%` and center it with flexbox.

#### After
image

### DIFF
--- a/packages/axiom-charts/src/BarChart/BarChart.css
+++ b/packages/axiom-charts/src/BarChart/BarChart.css
@@ -57,10 +57,10 @@
 
 .ax-bar-chart__bar-container .ax-bar-chart__bar-label {
   position: absolute;
-  top: 0;
-  align-self: center;
+  top: 50%;
   margin-right: calc(var(--cmp-chart-label-width) * -1);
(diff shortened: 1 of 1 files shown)


packages/axiom-charts/src/BarChart/BarChart.css
```diff
@@ -38,9 +38,11 @@
 }
 
 .ax-bar-chart__bar-container .ax-bar-chart__bar-label {
+  display: flex;
   position: absolute;
   top: 0;
```
HHogg: I'm not sure you can guarantee the contents of the label element is always going to be a single element? Which flexing it might cause more issues down the line... 

Maybe a 

```
top: 50%;
transform: translateY(-50%);
``` 

Might be safer here?

tomru: Good point, will change 👍

tomru: Addressed now in 26fd397

PR #568, opened 2018-05-25
refactor(CombinedBar): To accept tooltip and dropdown props

stefanbuck:
There is a need to be able to add two tooltips on a striped bar. This PR is in preparation for the actual PR which I will raise soon. It moves `ChartContext` into `CombinedBar` and to retain the previous behaviour.


packages/axiom-charts/src/BarChart/BarChartBars.js
```diff
@@ -97,31 +95,26 @@ export default class BarChartBars extends Component {
               left: `${showDifferenceArea && isStretched ? benchmarkValue : percent}%`,
             };
 
-            const FinalBar = showDifferenceArea ? CombinedBar : Bar;
```
stefanbuck: `CombinedBar` is wrapping `Bar` therefore we can use `CombinedBar` all the time.

PR #577, opened 2018-06-06
feat(Dropdown): Remove dropdown mask

lpoulter:
**Affected components**
* Dropdown
* AlertDropdown
* Select
* Datepicker

**Background**
The current implementation adds a mask to the DOM when a dropdown is open. This mask captures click events outside of the dropdown and causes the dropdown to be closed. This caused an undesirable double click when navigating between form elements. We wish to remove the mask, removing a click from the users flow.

**Approach**
This PR has proved challenging and I've tried a few different approaches. The one i've settled on is using a custom HOC and disabling pointer events. I could get no other approach to fully meet all the requirements. 
The `onclickoutside` HOC is used to wrap the dropdown source. This adds an event listener to the DOM which is then used to trigger closing of the dropdown. It was necessary to disable clicks, by disabling pointer events, on the dropdown target when the dropdown source is visible. This is to prevent the reopening of the dropdown once closed by a click-away. The mask previously captured this click.

**Deployed Branch here**
https://5b28fca0c965927b31dc9b9e--nostalgic-curie-a20a02.netlify.com/docs/packages/axiom-components/dropdown/

### DIFF
--- a/packages/axiom-components/src/AlertDropdown/AlertDropdown.js
+++ b/packages/axiom-components/src/AlertDropdown/AlertDropdown.js
@@ -85,9 +85,10 @@ class AlertDropdown extends Component {
       <Position
           { ...rest }
           isVisible={ isVisible }
-          onMaskClick={ () => this.close() }
           position="bottom">
-        <PositionTarget>{ findComponent(children, AlertDropdownTargetRef) }</PositionTarget>
+        <PositionTarget>
+          { findComponent(children, AlertDropdownTargetRef) }
+        </PositionTarget>
         <PositionSource>{ findComponent(children, AlertDropdownSourceRef) }</PositionSource>
       </Position>
     );
--- a/packages/axiom-components/src/Base/Base.js
+++ b/packages/axiom-components/src/Base/Base.js
@@ -34,6 +34,8 @@ export default class Base extends Component {
      * Opposite of `visibleUntil`.
      */
     hiddenUntil: PropTypes.oneOf(['small', 'medium', 'large']),
+    /** disables pointer events */
+    pointerEventsDisabled: PropTypes.bool,
     /** Vertical margins given to the element */
     space: PropTypes.oneOf(['x0', 'x1', 'x2', 'x3', 'x4', 'x5', 'x6', 'x8']),
     /**
@@ -135,6 +137,7 @@ export default class Base extends Component {
       cloakContainer,
       container,
       hiddenUntil,
+      pointerEventsDisabled,
       space,
       sticky,
       textBreak,
@@ -181,6 +184,7 @@ export default class Base extends Component {
       'ax-text--strong': textStrong,
       [`ax-text--underline-${underline}`]: underline,
       [`ax-theme--${theme}`]: theme,
+      'ax-pointer--disabled': pointerEventsDisabled,
     });
 
     if (sticky) {
--- a/packages/axiom-components/src/Base/Base.css
+++ b/packages/axiom-components/src/Base/Base.css
@@ -143,3 +143,7 @@
     pointer-events: all;
   }
 }
+
+.ax-pointer--disabled {
+  pointer-events: none;
+}
--- a/packages/axiom-components/src/DataPicker/__snapshots__/DataPicker.test.js.snap
+++ b/packages/axiom-components/src/DataPicker/__snapshots__/DataPicker.test.js.snap
@@ -52,6 +52,7 @@ exports[`DataPicker renders with defaultProps 1`] = `
                 data-ax-at="9c4d25dadaa7"
                 onClick={[Function]}
                 onFocus={[Function]}
+                pointerEventsDisabled={false}
               >
                 <a
                   className="ax-link ax-link--style-body"
@@ -173,6 +174,7 @@ exports[`DataPicker renders with just DataPickerDropdown 1`] = `
                 data-ax-at="9c4d25dadaa7"
                 onClick={[Function]}
                 onFocus={[Function]}
+                pointerEventsDisabled={false}
               >
                 <a
                   className="ax-link ax-link--style-body"
@@ -358,6 +360,7 @@ exports[`DataPicker renders with onSelectColor 1`] = `
                 data-ax-at="9c4d25dadaa7"
                 onClick={[Function]}
                 onFocus={[Function]}
+                pointerEventsDisabled={false}
               >
                 <a
                   className="ax-link ax-link--style-body"
@@ -479,6 +482,7 @@ exports[`DataPicker renders with value 1`] = `
                 data-ax-at="9c4d25dadaa7"
                 onClick={[Function]}
                 onFocus={[Function]}
+                pointerEventsDisabled={false}
               >
                 <a
                   className="ax-link ax-link--style-body"
--- a/packages/axiom-components/src/Dropdown/Dropdown.js
+++ b/packages/axiom-components/src/Dropdown/Dropdown.js
@@ -88,9 +88,12 @@ export default class Dropdown extends Component {
           { ...rest }
           isVisible={ isVisible }
           offset={ offset }
-          onMaskClick={ () => this.close() }
           position={ position }>
-        <PositionTarget>{ findComponent(children, DropdownTargetRef) }</PositionTarget>
+        <PositionTarget>
+          { React.cloneElement(findComponent(children, DropdownTargetRef), {
+            pointerEventsDisabled: isVisible,
+          } ) }
+        </PositionTarget>
         <PositionSource>{ findComponent(children, DropdownSourceRef) }</PositionSource>
       </Position>
     );
--- a/packages/axiom-components/src/Dropdown/DropdownContext.js
+++ b/packages/axiom-components/src/Dropdown/DropdownContext.js
@@ -39,6 +39,7 @@ export default class DropdownContext extends Component {
   constructor(props) {
     super(props);
     this.handleKeyDown = this.handleKeyDown.bind(this);
+    this.handleClick = this.handleClick.bind(this);
   }
 
   componentWillMount() {
@@ -47,6 +48,17 @@ export default class DropdownContext extends Component {
 
   componentWillUnmount() {
     document.removeEventListener('keydown', this.handleKeyDown);
+    document.removeEventListener('mousedown', this.handleClick);
+  }
+
+  componentDidMount() {
+    document.addEventListener('mousedown', this.handleClick);
+  }
+
+  handleClick() {
+    if (!this.el.contains(event.target)) {
+      return this.context.closeDropdown();
+    }
   }
 
   getFocusedMenuItem() {
--- a/packages/axiom-components/src/Position/Position.css
+++ b/packages/axiom-components/src/Position/Position.css
@@ -6,12 +6,3 @@
 .ax-position--arrow {
   margin: calc((var(--cmp-context-tip-edge-to-edge) * 0.5) + var(--spacing-grid-size--x1));
 }
-
-.ax-position__mask {
-  position: fixed;
-  top: 0;
-  right: 0;
-  bottom: 0;
-  left: 0;
-  z-index: var(--z-index-position);
-}
--- a/packages/axiom-components/src/Position/Position.js
+++ b/packages/axiom-components/src/Position/Position.js
@@ -31,11 +31,6 @@ export default class Position extends Component {
     /** Controls the starting offset of the content */
     offset: PropTypes.oneOf(['start', 'middle', 'end']),
     /**
-     * When provided a mask will be placed behind PositionSource, where this
-     * function is called when clicked.
-     */
-    onMaskClick: PropTypes.func,
-    /**
      * Optional handler that is called, with the new position, when PositionSource
      * has been positioned.
      */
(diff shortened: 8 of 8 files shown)


jdelight:
Nice one @lpoulter 👍 Just had a very quick look - I was expecting to see we were setting pointer events to none at some point... but I can't see it. Might be worth just having a technical explanation of how this works in the PR description. Especially as there's just the one commit.

jdelight:
Actually ignore my last comment - I think it's just the pointer events bit I was wondering about :)

lpoulter:
Is this what your looking for https://github.com/BrandwatchLtd/axiom/pull/577/files#diff-76879590576e94f3f91fa935294d0140R93
I think the Axiom style is just one commit.

package.json
```diff
@@ -59,6 +59,7 @@
     "react-addons-test-utils": "^16.0.0-0",
     "react-dom": "^16.2.0",
     "react-hot-loader": "^3.1.3",
+    "react-onclickoutside": "^6.7.1",
```
HHogg: This should be a dependency, not a dev dependency

packages/axiom-components/src/AlertDropdown/AlertDropdownContext.js
```diff
@@ -1,15 +1,25 @@
 import React, { Component } from 'react';
 import PropTypes from 'prop-types';
 import Context from '../Context/Context';
+import onClickOutside from 'react-onclickoutside';
```
HHogg: We've always tried to stay pretty light on dependencies, and looking at this one it's been pretty quiet for a few months with a few issues around latest react support. 

Have you tried an approach without using a library to see how difficult it would be to achieve?

packages/axiom-components/src/AlertDropdown/AlertDropdownContext.js
```diff
@@ -1,15 +1,25 @@
...
     type: PropTypes.oneOf(['success', 'warning', 'error', 'info']).isRequired,
+    closeDropdown: PropTypes.func.isRequired,
+  };
+
+  handleClickOutside = () => {
+    this.context.closeDropdown();
   };
 
   render() {
+    const onClickOutSideProps = ['eventTypes', 'outsideClickIgnoreClass', 'preventDefault', 'stopPropagation', 'disableOnClickOutside', 'enableOnClickOutside'];
```
HHogg: Omitting an external libraries list of props will take a bit of effort to maintain.

packages/axiom-components/src/AlertDropdown/AlertDropdown.js
```diff
@@ -85,9 +85,11 @@ class AlertDropdown extends Component {
       <Position
           { ...rest }
           isVisible={ isVisible }
-          onMaskClick={ () => this.close() }
           position="bottom">
-        <PositionTarget>{ findComponent(children, AlertDropdownTargetRef) }</PositionTarget>
+        <PositionTarget>
+          { <span style={ { pointerEvents: isVisible ? 'none' : 'auto' } } >
```
HHogg: This `span` will cause issues with layouts as it will introduce an element that will break the margin spacing pattern that is on every Axiom element.

package.json
lpoulter: Ahh yes I wondered about that. I can't see a dependency section in the package.json are they added in some other way?

packages/axiom-components/src/AlertDropdown/AlertDropdownContext.js
lpoulter: This library essentially adds events handlers to the body when the wrapped component is mounted. It also handles some of the ugly ref passing/getting and provides some useful options. It has pretty decent tests also. Rather than re-writing  a solution that works well. I think it's worth using The 16.3 issues don't look like problems to me more suggested refactors to make use of the new features in 16.3.

packages/axiom-components/src/AlertDropdown/AlertDropdownContext.js
lpoulter: Yes it is ugly. I'll see if I can find a better way.

packages/axiom-components/src/AlertDropdown/AlertDropdown.js
lpoulter: Hmm this is a tricky one. I could clone the child and add a `pointerEventsDisabled` prop to it.  However each possible child of `AlertDropdownTarget` and `DropdownTarget` would need to know what to do with it. Is there anyway I can keep the `pointerEvents` logic here and not affect the DOM structure.

lpoulter:
@HHogg I've tried a few different approaches both with and without the library and a HOC seems like the best approach. Is there any particular reason you've avoid HOC's in Axiom and could you suggest any alternative patterns?

jdelight:
@lpoulter / @HHogg I'm not opposed to using a library if there is one that's well written and not bloated with features we don't need (at the moment). 

But it would be worth taking some time to evaluate the available libraries, determine why we would choose one over the others (there are 3 I can see on a after a quick Google: https://github.com/kentor/react-click-outside, https://github.com/tj/react-click-outside, https://github.com/Pomax/react-onclickoutside).

Have you seen this one from TJ: https://github.com/tj/react-click-outside/blob/master/index.js looks very small / simple etc... does it work for what we need? If so is there value in writing one ourselves over using it directly. Maybe our version could be even simpler and use it as inspiration.

I think this is one of those features which doesn't typically need updating lots so I'm not too concerned about activity on the repo if it's "complete". But I agree with @HHogg that it's good to be extra cautious about which dependencies we introduce.

lpoulter:
Thanks for the links James. I've looked at these solutions. The simpler libraries such as TJ's library don't provide a reference to the wrapped component instance. This instance is used to call closeDropdown() within the clickOutside handler and allows for a more decoupled implementation.

The kentor and Pomax versions use a similar method, using the wrapped component instance. However I had some problems implementing the Kentor solution. The Pomax version is also used more widely and unlike kentor adds no transitive dependencies.

It's definitely possible to write our own solution, ideally using the new `React.createRef()` added in `16.3`, axiom is `16.2`.

jdelight:
Very interesting - thanks @lpoulter

lpoulter:
So this might be enough to cover our use case. Will need to add some edge cases and test. https://codesandbox.io/s/o4lr9z7l8y

lpoulter:
@jdelight @HHogg I've added a commit showing the use of my own implementation of an `onClickOutside` are you able to comment on the approach?

packages/axiom-components/src/Base/Base.js
```diff
@@ -34,6 +34,8 @@ export default class Base extends Component {
      * Opposite of `visibleUntil`.
      */
     hiddenUntil: PropTypes.oneOf(['small', 'medium', 'large']),
+    /** disables pointer events */
+    pointerEventsDisabled: PropTypes.bool,
```
HHogg: What's the reason for disabling pointer events?

packages/axiom-components/src/AlertDropdown/AlertDropdownContext.js
```diff
@@ -13,3 +19,5 @@ export default class AlertDropdownContext extends Component {
     );
   }
 }
+
+export default withOnClickOutside(AlertDropdownContext);
```
HHogg: How come we're wrapping each use of Dropdown rather than add the behaviour to the Dropdown component?

packages/axiom-components/src/AlertDropdown/AlertDropdownContext.js
```diff
@@ -1,10 +1,16 @@
...
+import { withOnClickOutside } from '@brandwatch/axiom-utils';
 
-export default class AlertDropdownContext extends Component {
+class AlertDropdownContext extends Component {
   static contextTypes = {
     type: PropTypes.oneOf(['success', 'warning', 'error', 'info']).isRequired,
+    closeDropdown: PropTypes.func.isRequired,
+  };
+
+  handleClickOutside = () => {
```
HHogg: The arrow method syntax was decided against (I can't find the PR) a while ago.

packages/axiom-components/src/Base/Base.js
lpoulter: Explained in the PR desc.

 "It was nessacary to disable clicks, by disabling pointer events, on the dropdown target when the dropdown source is visible. This is to the reopening of the dropdown once closed by a click-away."

packages/axiom-components/src/AlertDropdown/AlertDropdownContext.js
lpoulter: Any reason why. I like using the arrow syntax to avoid having to bind this in every constructor.

packages/axiom-components/src/AlertDropdown/AlertDropdownContext.js
lpoulter: Putting the logic in a HOC allows for it to be used in multiple components. Currently both `Dropdown` and `AlertDropdown` need this functionality.

packages/axiom-components/src/AlertDropdown/AlertDropdownContext.js
HHogg: IIRC it was to do with the spec stage it was in... personally I don't mind it. 

Axiom has always lent towards pattern consistency, so would you mind opening up a discussion to see how others feel and if everyone's happy to to do a full switch over to this pattern?

lpoulter: Sure will do. In the mean time i'll use the standard syntax.

packages/axiom-components/src/Base/Base.js
HHogg: Is this likely to cause issues for elements that still need some sort of mouse interaction (like a hover tooltip)? It just seems like a catch all approach, when the goal is to not reopen the dropdown.

It might be safer to try an approach of disabling the open dropdown method when it's already open or stop the event propagating up (when it's open) to the handler on the document.

lpoulter: "It might be safer to try an approach of disabling the open dropdown method when it's already open or stop the event propagating up (when it's open) to the handler on the document."

I did try these first, I agree they are better options. I can have another go with now i'm using my own implementation.

lpoulter: "Is this likely to cause issues for elements that still need some sort of mouse interaction (like a hover tooltip)?"

I imagine the current mask implementation has this issue also?

jdelight: I was thinking the same @lpoulter - presumably the `pointer-events: none` is only applicable to the dropdown target (i.e. the thing you initially click on). So all other hover / tooltip interactions should work as normal.

packages/axiom-components/src/AlertDropdown/AlertDropdownContext.js
lpoulter: We also want to wrap the `context` components rather than the whole dropwdown components. That way the event listener are only added when the dropdown is open.

packages/axiom-components/src/Base/Base.js
HHogg: Sorry I meant like a tooltip/title inside the target element. Yes the current mask does prevent this, but if we're removing the mask so interactions can happen while the dropdown is open, then it might throw people of as to why the interactions on the target don't work. 

If the behaviour to stop it reopening on clicking the target can be isolated to just that rather than stopping all interactions on the target then it's just less likely to trip up someone down the line.

packages/axiom-components/src/AlertDropdown/AlertDropdownContext.js
HHogg: I wouldn't worry to much about `AlertDropdown` as it was decided to remove that component as it doesn't fit well into the design system, and the adding and removing of event listeners can still happen in the `open` and `close` methods of Dropdown. 

It's just this way it keeps them more inline as a pure react component and we can limit the implementation inside Dropdown (ie not requiring components to implement a `handleClickOutside`).

lpoulter: I can move all the logic to the `Dropdown` component then. We will lose the ability to add `handleClickOutside` functionality to other components but maybe thats ok.

packages/axiom-components/src/Base/Base.js
jdelight: Sorry @HHogg not sure I follow :/ do you have an example of this on Axiom or in Audiences? thanks

jdelight:
@lpoulter the first version (commit) looks a lot better to me! what do you think? much less imperative state management and using event bubbling / propagation is much more how I would expect this to be done.

I think the 2 things @HHogg was after can still be achieved i.e. only applying it to the Dropdown component & not using HOC (I'm not as fussed about HOC but would be interested to see how it would work without using one). 

Perhaps if we can do that we're on to a winner with something like the first commit. 🎉 

@siggysamson / @thatguynamedandy what do you think?

lpoulter:
@thatguynamedandy Would be great to get your input on this now harry is away. I've gone with the HOC approach as the second approach had a bug and the need for extra state to be track.

jdelight:
Thanks for taking a look @thatguynamedandy 👍 @lpoulter what do you think?

lpoulter:
Thanks for taking a look @thatguynamedandy. I've already tried a solution like yours there are a couple of problems with it. Ignoring the click on the dropdown target means that we lose the toggle behaviour, clicking on target when open should close the dropdown.

HOC's are a widely used react pattern with many article and even the offical docs discussing them at length. The benefit of using one here is that the code is re-useable, allowing me to reuse the `clickOutside` functionality in any other component that may need it, both `Alertdropdown` and any future ones. However I understand your point about not wanting to many patterns in Axiom. I've updated my soultion to not use a HOC.

Are you able to take another look?

PR #580, opened 2018-06-11
feat(Dialog): Adds option to add padding to Dialog elements

ogandose:
Adding the possibility to add small, large, or no padding to DialogBody, DialogFooter, and DialogHeader.

### DIFF
--- a/packages/axiom-components/src/Dialog/DialogBody.js
+++ b/packages/axiom-components/src/Dialog/DialogBody.js
@@ -1,18 +1,29 @@
 import PropTypes from 'prop-types';
 import React, { Component } from 'react';
+import classnames from 'classnames';
 import Base from '../Base/Base';
 
 export default class DialogBody extends Component {
   static propTypes = {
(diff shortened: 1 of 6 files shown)


packages/axiom-components/src/Dialog/DialogBody.js
```diff
@@ -6,13 +6,19 @@ export default class DialogBody extends Component {
...
+    padding: 'small',
   };
 
   render() {
-    const { children, ...rest } = this.props;
+    const { children, padding, ...rest } = this.props;
 
     return (
-      <Base { ...rest } className="ax-dialog__body">
+      <Base { ...rest } className={ `ax-dialog__body--padding-${ padding }` }>
```
HHogg: Adding the class like this will break some of the CSS that is relying on the `ax-dialog__body` class. You can achieve the conditional class by using `classnames` _(There are example in other component files)_

Same comment applies to the other Dialog components below.

PR #581, opened 2018-06-11
feat(DialogBody): Adds option to add display:flex to DialogBody

ogandose:
Adding the option to add "display: flex" to DialogBody to have the possibility to display elements next to each other in the DialogBody.

### DIFF
--- a/packages/axiom-components/src/Dialog/Dialog.css
+++ b/packages/axiom-components/src/Dialog/Dialog.css
@@ -29,6 +29,10 @@
   overflow: auto;
 }
 
+.ax-dialog__body--flex {
+  display: flex;
+}
+
 .ax-dialog__footer {
   padding: 0 var(--page-padding-vertical) var(--page-padding-horizontal);
 }
--- a/packages/axiom-components/src/Dialog/DialogBody.js
+++ b/packages/axiom-components/src/Dialog/DialogBody.js
@@ -1,18 +1,24 @@
 import PropTypes from 'prop-types';
 import React, { Component } from 'react';
+import classnames from 'classnames';
 import Base from '../Base/Base';
 
 export default class DialogBody extends Component {
   static propTypes = {
     /** The main content side a Dialog that occupies most of the space */
     children: PropTypes.node,
+    /** Controls if body should be rendered with "display:flex;" */
+    flex: PropTypes.bool,
   };
 
   render() {
-    const { children, ...rest } = this.props;
+    const { children, flex, ...rest } = this.props;
 
+    const classes = classnames('ax-dialog__body', {
+      'ax-dialog__body--flex': flex,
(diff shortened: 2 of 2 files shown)


thatguynamedandy:
Hey @ogandose I think I get the use case, I was just wondering if this could be done already by using a grid as the first child of the DialogBody?

lpoulter:
@thatguynamedandy It would be nice to not have to use a component to achieve flex thou.

jdelight:
What would be the downside of adding the ability to add class names to `DialogBody`? and would that solve your problem @ogandose ?

**grid components**
@thatguynamedandy / @HHogg I think we should have the grid component be optional (as @lpoulter says). For example, the message from the experts is that we don't need Grid frameworks. In the same way we don't need frameworks for other parts of CSS - we can just use CSS Grid or Flexbox as appropriate. https://rachelandrew.co.uk/archives/2017/07/01/you-do-not-need-a-css-grid-based-grid-system/

**using class names**
In this specific PR - it seems we can already pass class names to the `Base` component. So is there any reason why we can't use that approach to set `display: flex;`. I guess allowing the class names to pass-through from `DialogBody`?

What happens when someone needs to use `display: grid;` on the `DialogBody` component. It would follow that we'd add a `grid: true|false` prop to it. This isn't intuitive for developers and doesn't scale as an approach IMHO. Rather we should find a way to allow the correct layout to be implemented without needing changes to Axiom components each time. I.e. use CSS classes.

The fact that this PR exists is a concern... adding `display: flex` to a `div` shouldn't require a PR. 

cc. @ahume / @Binarytales

siggysamson:
I think it's a good thing to add a property to the `DialogBody` rather than allowing passing down a CSS class.
Assuming we allow to add CSS classes to the `DialogBody` people might go completely wild and would even change the look&feel of the Dialog itself which then causes maintenance issues down the line. By limiting the _options_ it makes it easier to roll out changes within Axiom that can be applied everywhere.
I agree it would be good to have the option to make something a flex container or grid container but in that case I'd argue it's something to be added to the `Base` component and every other component where it makes sense should just allow to passing down the properties. Then we'd still limit the usage but allow for a certain flexibility.
I don't see why allowing CSS classes would help us in anything in that regard

kamilio:
I think, allowing custom css classes for Axiom components will cause Axiom to lock down to current html structure and will make any future refactorings/changes super hard. Currently Axiom is not mature enough, maybe in couple of months, a year. 

Let's take the discussion out of this PR to https://github.com/BrandwatchLtd/axiom/issues/585. We need to ship :shipit:

jdelight:
Thanks for your thoughts @thatguynamedandy / @siggysamson 👍 I agree we can move this to an issue or can be one for the working group when it starts.

PR #582, opened 2018-06-11
feat(Colors): Added new product colors

HHogg:
screen shot 2018-06-11 at 12 33 12


packages/axiom-charts/src/Bar/Bar.css
```diff
@@ -73,23 +73,29 @@
       );
   }
 
+  &.ax-bars__bar-rect--forbidden-plaent { color: var(--color-product-forbidden-plaent); }
```
jdelight: should this be "forbidden-plaent"? looks like we have "forbidden-planet" below.

HHogg: 🤦‍♂️ Thanks @jdelight. I'll have another run through to make sure there are no others

pmsorhaindo:
@jdelight are you happy with this now?

PR #583, opened 2018-06-11
fix(List): Tidies up the height of the vertical rule

thatguynamedandy:
This reverts commit 262d494da521302278da7a4a41e3de94c5d8fce9.

Then takes an alternative approach using relative positioning.

### DIFF
--- a/packages/axiom-components/src/List/List.css
+++ b/packages/axiom-components/src/List/List.css
@@ -1,5 +1,6 @@
 :root {
   --cmp-list-item-spacing: calc(var(--spacing-grid-size) * 3);
+  --cmp-list-item-separator: calc(var(--spacing-grid-size) * 3);
 }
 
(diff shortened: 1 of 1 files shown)


packages/axiom-components/src/List/List.css
```diff
@@ -10,13 +11,19 @@
...
   }
 
   & .ax-list__item:not(:last-child)::after {
     content: '';
-    margin: 0 var(--cmp-list-item-spacing) 0 0;
-    padding: 0 var(--cmp-list-item-spacing) 0 0;
+    position: absolute;
+    top: 50%;
+    right: 0;
+    height: var(--cmp-list-item-separator);
```
HHogg: Little bit hesitant on this, over applying vertical margins to get the desired height for the common case, but relatively it's not a massive issue.

PR #586, opened 2018-06-12
Fix Netlify build

stefanbuck:


.travis.yml
```diff
@@ -14,6 +14,7 @@ script:
 - yarn lint:commits
 - yarn lint
 - yarn ci:test
+- yarn build:site
```
stefanbuck: @HHogg let me know what you think about it, but I think it's a nice addition to ensure that we can build docs

HHogg: Nice. Maybe remove the build:site from the publish_site.sh, seems unnecessary to do it twice?

stefanbuck: That's true

PR #591, opened 2018-06-21
feat(Add option to change cursor on text input)

lpoulter:
Currently when a `readOnly` `TextInput` is used as a `DropdownTarget` the onHover cursor is a set to `text`. This is confusing as the input is not editable and onClick triggers the dropdown opening. This PR add an option to the `TextInput` to allow the cursor to be set to `pointer` in this situation. 

https://5b2b8489b13fb1210a8fbb0a--president-randy-42828.netlify.com/docs/packages/axiom-components/dropdown

### DIFF
--- a/packages/axiom-components/src/Form/InputWrapper.css
+++ b/packages/axiom-components/src/Form/InputWrapper.css
@@ -24,3 +24,7 @@
   position: relative;
   flex-direction: column;
 }
+
+.ax-input__wrapper--target .ax-input {
+  cursor: pointer;
+}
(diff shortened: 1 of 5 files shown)


lpoulter:
@siggysamson would be interested in your thoughts on this as it affects the Benchmark component.

thatguynamedandy:
I had two thoughts..
1. Should cursor be a property on `Base` so that we get full flexibility
2. Opposite of the above, should we always enforce the 'pointer' if its `readonly` and has an `onClick`

I don't feel too strongly in favour any of the three options, but thought I would mention them anyway to see what other people thought (cc @siggysamson)

lpoulter:
I thought about 2. I don't think thats a good idea because the text input might be used as read only outside of the dropdown. In that case you wouldn't want a pointer cursor.

lpoulter:
One could work I do worry base is getting really overloaded and just a catch all for CSS tweaks.

thatguynamedandy:
That was pretty much my thought process, I'm happy to go with this.

PR #592, opened 2018-06-21
feat(Tooltip): Adds boolean delay option

thatguynamedandy:
Gives the option of adding a one second delay to the opening of a tooltip (UX requirement)

https://tooltip-display-delay--andyp-axiom.netlify.com/docs/packages/axiom-components/tooltip

### DIFF
--- a/packages/axiom-components/src/Tooltip/TooltipTarget.js
+++ b/packages/axiom-components/src/Tooltip/TooltipTarget.js
@@ -2,10 +2,12 @@ import PropTypes from 'prop-types';
 import { Component, cloneElement } from 'react';
 
 export const TooltipTargetRef = 'TooltipTarget';
+const DELAY_TIMEOUT = 1000;
 
 export default class TooltipTarget extends Component {
   static propTypes = {
     children: PropTypes.node,
+    delay: PropTypes.bool,
     onClick: PropTypes.func,
   };
 
@@ -16,19 +18,29 @@ export default class TooltipTarget extends Component {
 
   static typeRef = TooltipTargetRef;
 
-  handleMouseMove(...args) {
-    const { children } = this.props;
+  handleMouseMove(event, ...args) {
+    const { children, delay } = this.props;
     const { showTooltip } = this.context;
     const { onMouseMove = () => {} } = children.props;
 
-    showTooltip();
-    onMouseMove(...args);
+    if (delay && !this.showTimeout) {
+      this.showTimeout = setTimeout(() => showTooltip(), DELAY_TIMEOUT);
+    }
+
+    if (!delay) showTooltip();
+
+    onMouseMove(event, ...args);
   }
 
   handleMouseLeave(...args) {
     const { children } = this.props;
     const { hideTooltip } = this.context;
     const { onMouseLeave = () => {} } = children.props;
+
+    if (this.showTimeout) {
+      clearTimeout(this.showTimeout);
+      this.showTimeout = undefined;
+    }
 
     hideTooltip();
     onMouseLeave(...args);
--- a/packages/axiom-components/src/Tooltip/TooltipTarget.js
+++ b/packages/axiom-components/src/Tooltip/TooltipTarget.js
@@ -18,7 +18,7 @@ export default class TooltipTarget extends Component {
 
   static typeRef = TooltipTargetRef;
 
-  handleMouseMove(event, ...args) {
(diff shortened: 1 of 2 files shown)


lpoulter:
Doesn't seem to delay the tooltip and getting an error on the local demo `If you want to write it to the DOM, pass a string instead: delay="false" or delay={value.toString()}`

packages/axiom-components/src/Tooltip/TooltipTarget.js
```diff
@@ -16,20 +18,30 @@ export default class TooltipTarget extends Component {
 
   static typeRef = TooltipTargetRef;
 
-  handleMouseMove(...args) {
-    const { children } = this.props;
+  handleMouseMove(event, ...args) {
```
lpoulter: Do we still need `...args` here are they always empty?

thatguynamedandy:
Strange, is this on a local demo on your machine, or the Netlify one. (I have to confess to only testing Chrome)

tooltip-delay

packages/axiom-components/src/Tooltip/TooltipTarget.js
thatguynamedandy: Strange, is this on a local demo on your machine, or the Netlify one. (I have to confess to only testing Chrome)

tooltip-delay

packages/axiom-components/src/Tooltip/TooltipTarget.js
```diff
@@ -16,20 +18,30 @@ export default class TooltipTarget extends Component {
...
 
-    showTooltip();
-    onMouseMove(...args);
+    if (delay && !this.showTimeout) {
+      this.showTimeout = setTimeout(() => showTooltip(), DELAY_TIMEOUT);
+    }
+
+    if (!delay) showTooltip();
+
+    onMouseMove(event, ...args);
```
lpoulter: What's the purpose of `onMouseMove` on the child here? Is it so we can define a custom callback? If so is it a problem that the callback will fire before `showTooltip`?

packages/axiom-components/src/Tooltip/TooltipTarget.js
thatguynamedandy: Strange, is this on a local demo on your machine, or the Netlify one. (I have to confess to only testing Chrome)

tooltip-delay

thatguynamedandy:
Strange, is this on a local demo on your machine, or the Netlify one. (I have to confess to only testing Chrome)

tooltip-delay

packages/axiom-components/src/Tooltip/TooltipTarget.js
thatguynamedandy: Ah, good point, I changed that and never put it back to how it was.

packages/axiom-components/src/Tooltip/TooltipTarget.js
thatguynamedandy: I think it's so if there is a use case for more mouse movement it can be handled. To be honest that was existing functionality which I just replicated.

packages/axiom-components/src/Tooltip/TooltipTarget.js
```diff
@@ -17,11 +19,16 @@ export default class TooltipTarget extends Component {
...
 
   handleMouseMove(...args) {
-    const { children } = this.props;
+    const { children, delay } = this.props;
     const { showTooltip } = this.context;
     const { onMouseMove = () => {} } = children.props;
 
-    showTooltip();
+    if (delay && !this.showTimeout) {
+      this.showTimeout = setTimeout(() => showTooltip(), DELAY_TIMEOUT);
```
lpoulter: I think this can be simplified to `setTimeout(showTooltip, DELAY_TIMEOUT);` The `this` context is already bound in `Tooltip`

thatguynamedandy:
@siggysamson I think on the delay it's good that it's either an on or off delay. I thought about allowing people to specify a time, but I think consistency is nice across the board. It's been through a quick round of design review, but I'll double check 1 second is the most appropriate time.

packages/axiom-components/src/Tooltip/TooltipTarget.js
lpoulter: Interesting and not really for this PR but it seems to just confuse things and may not fire when the user expects with the delay.

PR #594, opened 2018-06-28
fix(Dropdown): Restore pointer events

lpoulter:
In the work to remove the dropdown mask it was nessacary to disable pointer events from on the Dropdown Target. This was nessacary to avoid the double open bug caused by simultaneous firing of a `focus` & `click` event. 

This cause two problems

1. When the child of the Dropdown target was a DOM element it was being passed the prop `pointerEventsDisabled` this caused an error and did not disable pointer events.

2. On hover affects on the Dropdown target were lost when the dropdown was open.

This PR addresses these issues by replacing the `pointerEventsDisabled` approach with some state within `DropdownTarget` to avoid the double open bug.

Preview
https://5b34b634dd28ef2de9fff4a1--nostalgic-curie-a20a02.netlify.com/docs/packages/axiom-components/dropdown

### DIFF
--- a/packages/axiom-components/src/Dropdown/DropdownContext.js
+++ b/packages/axiom-components/src/Dropdown/DropdownContext.js
@@ -34,6 +34,7 @@ export default class DropdownContext extends Component {
 
   static contextTypes = {
     closeDropdown: PropTypes.func.isRequired,
+    dropdownRef: PropTypes.func.isRequired,
   };
 
   constructor(props) {
@@ -55,8 +56,9 @@ export default class DropdownContext extends Component {
     document.addEventListener('mousedown', this.handleClick);
   }
 
-  handleClick() {
-    if (!this.el.contains(event.target)) {
+  handleClick(event) {
+    const dropdownRef = this.context.dropdownRef();
+    if (!dropdownRef.contains(event.target) && !this.el.contains(event.target)) {
       return this.context.closeDropdown();
     }
   }
--- a/packages/axiom-components/src/Dropdown/DropdownTarget.js
+++ b/packages/axiom-components/src/Dropdown/DropdownTarget.js
@@ -10,14 +10,33 @@ export default class DropdownTarget extends Component {
 
   static contextTypes = {
     openDropdown: PropTypes.func.isRequired,
+    toggleDropdown: PropTypes.func.isRequired,
   };
 
   static typeRef = DropdownTargetRef;
 
-  handleOpen(cb, ...args) {
+  constructor(props) {
+    super(props);
+    this.focusEventHasFired = false;
+  }
+
+  handleClick(cb, event) {
+    const { toggleDropdown } = this.context;
+
+    if (this.focusEventHasFired) {
+      this.focusEventHasFired = false;
+      return;
+    }
+
+    toggleDropdown(event);
+    if (cb) cb(event);
+  }
+
+  handleFocus(cb, event) {
     const { openDropdown } = this.context;
-    openDropdown();
-    if (cb) cb(...args);
+    this.focusEventHasFired = true;
+    openDropdown(event);
+    if (cb) cb(event);
   }
 
   render() {
@@ -25,8 +44,8 @@ export default class DropdownTarget extends Component {
 
     return cloneElement(children, {
       ...rest,
-      onClick: this.handleOpen.bind(this, children.props.onClick),
-      onFocus: this.handleOpen.bind(this, children.props.onFocus),
+      onClick: this.handleClick.bind(this, children.props.onClick),
+      onFocus: this.handleFocus.bind(this, children.props.onFocus),
     });
   }
 }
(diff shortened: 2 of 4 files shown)


packages/axiom-components/src/Dropdown/DropdownTarget.js
```diff
@@ -10,23 +10,43 @@ export default class DropdownTarget extends Component {
...
-  handleOpen(cb, ...args) {
+  constructor(props) {
+    super(props);
+    this.isFocus = false;
+  }
+
+  handleClick(cb, event) {
+    const { toggleDropdown } = this.context;
+
+    if (this.isFocus) {
```
jdelight: NABD but I would suggest this being at the top of this function (i.e. return as early as possible) `toggleDropdown` is only required if this is false.

packages/axiom-components/src/Dropdown/DropdownTarget.js
```diff
@@ -10,23 +10,43 @@ export default class DropdownTarget extends Component {
...
 
   render() {
     const { children, ...rest } = this.props;
 
     return cloneElement(children, {
       ...rest,
-      onClick: this.handleOpen.bind(this, children.props.onClick),
-      onFocus: this.handleOpen.bind(this, children.props.onFocus),
+      onClick: this.handleClick.bind(this, children.props.onClick),
+      onFocus: this.handleFocus.bind(this, children.props.onFocus),
```
jdelight: Makes sense to have two separate handlers (and I guess that's what this solution is based on).
However, does it have a dependency on the order in which events are fired? 

Seems like we're expecting the `focus` event to fire before `click` so we can make `click` a no-op.

Is this behaviour consistent between browsers? or is it something React abstracts away and makes consistent (via its SyntheticEvent model)?

packages/axiom-components/src/Dropdown/DropdownTarget.js
```diff
@@ -10,23 +10,43 @@ export default class DropdownTarget extends Component {
...
+    if (this.isFocus) {
+      this.isFocus = false;
+      return;
+    }
+
+    toggleDropdown(event);
+    if (cb) cb(event);
+  }
+
+  handleFocus(cb, event) {
```
jdelight: Looks like we only have one side of this event at the moment. I'd be interested to know if we should be registering an `onBlur` event handler to effectively restore the default state. Seems to work though (i.e. if I tab onto a dropdown then tab away it closes... not sure how that's currently working though :)

packages/axiom-components/src/Dropdown/DropdownTarget.js
```diff
@@ -10,23 +10,43 @@ export default class DropdownTarget extends Component {
 
   static contextTypes = {
     openDropdown: PropTypes.func.isRequired,
+    closeDropdown: PropTypes.func.isRequired,
```
jdelight: Can't see this used anywhere at the moment. Maybe one for `onBlur`?

packages/axiom-components/src/Dropdown/DropdownContext.js
```diff
@@ -56,7 +57,8 @@ export default class DropdownContext extends Component {
   }
 
   handleClick() {
-    if (!this.el.contains(event.target)) {
+    const dropdownRef = this.context.dropdownRef();
+    if (!dropdownRef.contains(event.target) && !this.el.contains(event.target)) {
```
jdelight: Ok so this is how we're effectively doing `blur`?

packages/axiom-components/src/Dropdown/DropdownTarget.js
jdelight: https://patrickhlauke.github.io/getting-touchy-presentation/ looks useful as do his tests.
https://patrickhlauke.github.io/touch/

lpoulter: We still need the Click function as we need to listen for click events after the target has been focused. I've done a bit of reading and it seems that the order is not given in the spec. However it seems all browser do fire the events in this order, i've tested in chrome, firefox and safari. I'll check Edge & ie11 also.

packages/axiom-components/src/Dropdown/DropdownTarget.js
lpoulter: tab into: dropdown fires.

packages/axiom-components/src/Dropdown/DropdownTarget.js
jdelight: https://www.w3.org/TR/pointerevents/#compatibility-mapping-with-mouse-events

> The relative ordering of these high-level events (click, contextmenu, focus, blur, etc.) with pointer events is undefined and varies between user agents.

This is worth noting - however, looking at the tests on Patrick Lauke's website I can only see focus firing before click which is good. Seems like all browsers we care about have (so far)  agreed on the event precedence.

packages/axiom-components/src/Dropdown/DropdownTarget.js
lpoulter: Intresting it looks as if tab out closes the dropdown because the switch statement in `handleKeyDown` falls through to the default case because we don't return from `Tab`. @HHogg do you remember why we don't handle the `Tab` case and can you see this being a problem?

packages/axiom-components/src/Dropdown/DropdownTarget.js
lpoulter: Yes it's not in the spec but all the browser seem to follow this convention. It's not ideal but I would think it very unlikely browser implementors would change the order now.

PR #595, opened 2018-06-28
 feat(StatusBadge): Added new StatusBadge component

HHogg:
screen shot 2018-06-28 at 15 48 58

[Netlify Deploy](https://status-badges--hhogg-axiom.netlify.com/docs/packages/axiom-components/status-badge/)


packages/axiom-components/src/StatusBadge/StatusBadge.css
```diff
@@ -0,0 +1,3 @@
+.ax-status-badge {
+  display: block;
```
jdelight: Would `inline-block` work here instead? I've not seen any designs for how this fits in with other things so I've not got anything to base this suggestion on other than the use-cases in Analytics where it's using `inline-block`.

image
image

HHogg: The scenarios I've seen where these will be used look to be more flex based layouts where the badges are aligned vertically (with content on either side) off of the baseline, with a strict `.5rem/0.75rem` left and right margins. 

But `inline-block` shouldn't affect that, so I'm happy to update.

jdelight: Makes sense - good to know that block seems to be a better choice if we're primarily using flex based layouts (for the line-height reason). 👍

packages/axiom-components/src/StatusBadge/StatusBadge.js
```diff
@@ -0,0 +1,40 @@
...
+    ]).isRequired,
+  };
+
+  render() {
+    const { status, ...rest } = this.props;
+    const { body, height, viewBox, width } = BADGE_MAP[status];
+
+    return (
+      <Base { ...rest }
+          Component="svg"
```
jdelight: Interesting!... would be good to know how what the alternative is here. I.e. I would expect this to just be an SVG component. Do we get something extra by using Base with a Component prop? Just curious :)

HHogg: The main case for using the Base component is the use of the `space` prop, but in general all Axiom components will use a Base component to provide current and future the top level utility styling.

jdelight: I see - thanks for the explanation 👍

PR #599, opened 2018-06-29
feat(Dropdown): Change context menu focus design

lpoulter:
The current dropdown focus style is very strong and the same as the hover style. This creates confusing states like the one below. I think a good solution would be to make the focus state a little more subtle. 

[Demo](https://5b36373167610c6f1711dd23--nostalgic-curie-a20a02.netlify.com/docs/packages/axiom-components/dropdown)

Old focus style
screen shot 2018-06-29 at 14 39 56

Changed focus style
screen shot 2018-06-29 at 14 48 06

### DIFF
--- a/packages/axiom-components/src/Context/Context.css
+++ b/packages/axiom-components/src/Context/Context.css
@@ -144,10 +144,10 @@
 
 .ax-context-menu__item:focus {
   outline: 0;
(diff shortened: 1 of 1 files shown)


jdelight:
Has there been more though / progress on this @lpoulter. We're looking using Axiom dropdowns in frontend this sprint.

PR #600, opened 2018-07-02
docs(TooltipContext): add documented support for color prop

TomWija:
https://5b39ff71dd28ef612e68618b--axiomwija.netlify.com/docs/packages/axiom-components/tooltip


packages/axiom-components/src/Tooltip/TooltipSource.js
```diff
@@ -6,6 +6,7 @@ export const TooltipSourceRef = 'TooltipSource';
 export default class TooltipSource extends Component {
```
HHogg: I know this works because of the cloning but should it go on TooltipContext to match the DropdownContext and Context components?

PR #601, opened 2018-07-02
fix(Position): Boundary element to be viewport

HHogg:
This is a fix to positioning issue in various places. Popper defaults to it's boundaries element being the nearest scroll container. However when ever overflow is used it will effect the positioning by trying to keep it within those bounds. 

#### Issue example
This contextual menu should be displayed central as there is more than enough room on either side.

example1 

#### Issue example

This contextual menu is targeting an element inside the console (a scroll area) however it should be positioned out over the canvas area where there is enough room. 

example2

#### Fix

To change the default from the nearest scroll parent to the viewport. This seems like the most common use case for positioning for our layouts. I'm not sure it's needed but we can always add an `boundariesElement` prop to the `Position` component to allow it to be configured, but I'm not convinced it is needed. 

This is an example of the changes, which position correctly for inside and outside of the platforms canvas, and other overflow areas. 

example4

This change also still allows it to respond correct when scrolling an element or the window. 

example5

### DIFF
--- a/packages/axiom-components/src/Position/Position.js
+++ b/packages/axiom-components/src/Position/Position.js
@@ -15,6 +15,13 @@ import './Position.css';
 export default class Position extends Component {
   static propTypes = {
     /**
+     * Allows the boundary element for the positioning to be set.
+     */
+    boundariesElement: PropTypes.oneOfType([
+      PropTypes.object,
+      PropTypes.oneOf(['scrollParent', 'viewport', 'window']),
+    ]),
+    /**
      * Children inside Position this should contain all of and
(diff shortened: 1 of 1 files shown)


jdelight:
@HHogg I think this should be customisable to allow for future use-cases - especially has it doesn't cost us anything as we're wrapping Popper.js.

See this PR for some more thoughts: https://github.com/BrandwatchLtd/axiom/pull/462

HHogg:
Thanks @jdelight. I've added the prop mentioned in that PR.

jdelight:
Thanks @HHogg

PR #602, opened 2018-07-02
feat (Dialog) : Added the ability to set width on the Dialog componen…

MariamBashir:
**feat (Dialog) : Added the ability to set width on the Dialog component in axiom**

The PR implements and touches the following features

1. Add the ability to set width for the dialog and keep the fullscreen functionality.
2. Update documentation with the relevant changes.
3. Add test for the Dialog

**BREAKING CHANGE: 'removes' the size ['small', 'large', 'medium'] property from the Dialog**


HHogg:
A long needed change. Nice one @MariamBashir. Have you got Netlify setup and/or a link so it can be looked at?

SeaOfTea:
I'm not 100% sure this is possible, but could we possibly keep the 'small', 'medium' and 'large' for backwards compatibility? But yet still have the functionality of using an explicit width if one of those is not used?

HHogg:
@SeaOfTea that was my first though as well, and I guess it would be possible... but from what I see of the designs, the sizes of the dialogs are mostly always custom on a case by case basis.

We could have these values, but will they ever get used if the values don't match the designs?

MariamBashir:
@HHogg 
Yeah i am on it. Its still deploying. Also i wanted to mention about the tests. I faced a problem and i had to mock ```createPortal```. May be this (https://github.com/facebook/react/issues/11565) will provide some more context into it.

MariamBashir:
@HHogg 

Here's the deploy branch. https://5b3a23d5792f89751c5c3e1e--affectionate-mcclintock-42eb2e.netlify.com/docs/packages/axiom-components/dialog

Sorry for not having done this in the beginning. Had some confusions.

packages/axiom-components/src/Dialog/Dialog.js
```diff
@@ -24,14 +24,14 @@ export default class Dialog extends Component {
...
-    size: PropTypes.oneOf(['small', 'medium', 'large']),
     /** Theme of the dialog */
     theme: PropTypes.oneOf(['day', 'night']),
+    /** Custom width for Dialog */
+    width: PropTypes.string,
   };
 
   static defaultProps = {
-    size: 'small',
+    width:'50rem',
```
HHogg: I would probably avoid the default here for a couple of reasons. 
- It current breaks the `fullscreen` property being used without setting the `width`. Because it sets `width: 100vw;` in the CSS. 
- With a default being provided, it might be used instead of identifying the correct width for the designs... likewise is the `50rem` related to any designs?

MariamBashir: @HHogg 

1. I completely agree with you on that, Will fix that right away.
2. No its not related to any design, it was ```medium``` before so i just got the corresponding value.

PR #603, opened 2018-07-02
Chore(CI): fix github auth for package publishing

tomru:
- [x] created new `encrypted_7c2951e40ae6_key` and `encrypted_7c2951e40ae6_iv` values and set them in Travis
- [x] created a new ssh key dedicated for automatic publishing via Travis
- [x] encrypted that private key with the key and initialisation vector from above and updated the deploy key in the repo
- [x]  added the public ssh key as a deploy key to the axiom repo (with read/write access)
- [x] re-enabled the publish script


tomru:
Note: the branch build won't publish, hence it does not test github auth on the PR build

HHogg:
Awesome @tomru. Can you think of a way to test the SSH in the `scripts` section? So if something breaks in future, it's a bit more visible.

scripts/publish_packages.sh
```diff
@@ -10,6 +10,14 @@ chmod 600 github_deploy_key
 eval `ssh-agent -s`
 ssh-add github_deploy_key
 
+# Test SSH connection
+if ! ssh git@github.com 2>&1 | grep "BrandwatchLtd/axiom" > /dev/null
```
tomru: This is a little hacky, as testing via `ssh git@github.com` has a non-zero exit code and outputs to stderr. 

To verify that we have access, we could either check if we get `Permission denied (publickey).` or the username returned. For Deploy keys, which we are using here, it returns the repository, e.g. `BrandwatchLtd/axiom`.

To avoid confusion with other key's, I went the route of verifying that we are using a correct deploy key.

PR #606, opened 2018-07-03
feat(Dropdowns): Set show arrow to true in Dropdown and Position comp…

lpoulter:
…onent

AX-35. 
This PR removes the need for `showArrow={ false }` to be set on `Dropdown` components. Dropdown was used with this setting the majority of the time so it is now set to default.

I also considered other defaults we might want to set. I looked though the Brandwatch usage of each component in the `charts` & `components` packages. Whilst there was a few usages of setting a `true` default to `false` only the Dropdown and position component usage consistently overwrote this default.

showArrow shall remain the default for the chart and tooltips component.

BREAKING CHANGE: changes the default arrow behaviour for the Dropdown component.

[DEMO](https://5b3ccc2ffdd72a1f935d4f25--nostalgic-curie-a20a02.netlify.com/)


HHogg:
@lpoulter I think there's a few cases of Dropdown around Axiom in examples and also within the charts  that will need `showArrow` adding to them.

lpoulter:
ahh good point.

packages/axiom-components/src/Tooltip/Tooltip.js
```diff
@@ -75,7 +75,7 @@ export default class Tooltip extends Component {
     const { isVisible } = this.state;
 
     return (
-      <Position { ...rest } isVisible={ isVisible } position={ position }>
+      <Position { ...rest } isVisible={ isVisible } position={ position } showArrow>
```
HHogg: Is this one correct? Before it was configurable with the `{ ...rest }`, this would overwrite it's configuration. Should it be treated the same as `Dropdown` where it's default by false?

packages/axiom-composites/src/UserMenu/UserMenu.js
```diff
@@ -53,7 +53,7 @@ export default class UserMenu extends Component {
     const { axiomLanguage } = this.context;
 
     return (
-      <Dropdown position="bottom" { ...rest }>
+      <Dropdown position="bottom" { ...rest } showArrow>
```
HHogg: There's been a recent design request for `UserMenu` not to have the arrow, so you can avoid adding this one.

packages/axiom-components/src/Tooltip/Tooltip.js
lpoulter: I thought it would as with the charts the `Tooltip` would always want the arrow.

PR #607, opened 2018-07-03
AX-14 - usage hints on form components

TomWija:
https://5b3b9728fdd72a4dc35d4f14--axiomwija.netlify.com/docs/packages/axiom-components/form/


packages/axiom-components/src/Form/UsageHint.js
```diff
@@ -0,0 +1,34 @@
...
+import Dropdown from '../Dropdown/Dropdown';
+import DropdownContent from '../Dropdown/DropdownContent';
+import DropdownContext from '../Dropdown/DropdownContext';
+import DropdownSource from '../Dropdown/DropdownSource';
+import DropdownTarget from '../Dropdown/DropdownTarget';
+
+import Icon from '../Icon/Icon';
+import Link from '../Typography/Link';
+
+const UsageHint = ({ usageHint, usageHintPosition }) => (
```
HHogg: At the moment Axiom consistently uses stateful components across the board. At the time, this was because of some of the tooling didn't work with the stateless functional components, PR diffs when changing between the two was a nightmare to review the actual changes and because there wasn't really any noticeable performance differences. 

For all these reasons I would say for now still use the same format, but all these reasons/opinions around them might have changed. You can also use the PureComponent, which is being used in Axiom in some places, probably less than they could/should be.

packages/axiom-components/src/Form/UsageHint.js
```diff
@@ -0,0 +1,34 @@
...
+  <Dropdown position={ usageHintPosition }>
+    <DropdownTarget>
+      <Link onClick={ e => e.preventDefault() } style="subtle" >
```
HHogg: Do you need the preventDefault here?

packages/axiom-components/src/index.js
```diff
@@ -143,5 +143,6 @@ export { default as TooltipSource } from './Tooltip/TooltipSource';
 export { default as TooltipTarget } from './Tooltip/TooltipTarget';
 export { default as Transition } from './Transition/Transition';
 export { default as Underline } from './Typography/Underline';
+export { default as UsageHint } from './Form/UsageHint';
```
HHogg: Maybe if this is going to be hinting pattern for application, it should be inside it's own module. I can definitely think of places in applications already where this pattern is used not in the context of a form.

site/components/Documentation/Resources/Components/Form.js
```diff
@@ -26,9 +27,11 @@ export default class Documentation extends Component {
           <GridCell>
             <DocumentationShowCase centered>
               <TextInput
+                  label="label"
```
HHogg: A nicer label example might be nice here and below... it looks a bit messy in the docs, all lower case and short.

packages/axiom-components/src/Form/UsageHint.js
```diff
@@ -0,0 +1,34 @@
...
+          { usageHint }
+        </DropdownContent>
+      </DropdownContext>
+    </DropdownSource>
+  </Dropdown>
+);
+
+UsageHint.propTypes = {
+  usageHint: PropTypes.string,
+  usageHintPosition: PropTypes.oneOf(['top', 'bottom', 'left', 'right']),
```
HHogg: As the component is already prefixed to UsageHint, perhaps this could just be `position`?

packages/axiom-components/src/Form/UsageHint.js
```diff
@@ -0,0 +1,34 @@
...
+        <DropdownContent>
+          { usageHint }
+        </DropdownContent>
+      </DropdownContext>
+    </DropdownSource>
+  </Dropdown>
+);
+
+UsageHint.propTypes = {
+  usageHint: PropTypes.string,
```
HHogg: It might be safer in the long run for this to be as children. It could then also better/nicer support more complex text structures (which I have already seen in apps). Like ... 

```jsx
<UsageHint>
  <TextIcon name="smiley-face" />  Lorem ipsum <Strong>dolar amor</Strong> 
</UsageHint>
```

site/components/Documentation/Resources/Components/Form.js
```diff
@@ -103,6 +119,16 @@ export default class Documentation extends Component {
           require('!!axiom-documentation-loader!@brandwatch/axiom-components/src/Form/RadioButton'),
           require('!!axiom-documentation-loader!@brandwatch/axiom-components/src/Form/RadioButtonGroup'),
         ] } />
+
+        <DocumentationShowCase>
+          <UsageHint>
```
HHogg: No biggy but this example isn't needed anymore and the UsuageHint is still living inside the Form folder.

Flawwles:
👍 Design is fine with me

PR #610, opened 2018-07-04
fix(Typography): increase line heights for display and display2

pmsorhaindo:
Increase line heights for display and display2 text types to prevent clipping in descender characters.

https://5b3ca0993813f00ed4f16e69--librarian-chimpanzee-34483.netlify.com/docs/packages/axiom-components/typography


Flawwles:
👍 Looks good to me

PR #611, opened 2018-07-04
chore(Site): Reenable site publishing

HHogg:


scripts/publish_site.sh
```diff
@@ -10,8 +10,6 @@ curl ci-utils.bwcom.io/gcloud/auth | bash
 
 export COMPRESSION_PLUGIN=true
 
-yarn build:site
```
lpoulter: Did you mean to delete this?

HHogg: Yeah, in a previous PR this got moved to the scripts section of travis (so breakages in the site building would fail a build). So by the time it reaches the publish script, the site has already been built... no need to build it again.

PR #613, opened 2018-07-05
Optional mask

HHogg:
- Adds onMaskClick back into Position 
- Adds withMask property to Dropdown and AlertDropdown 
- Changes Chart dropdown to use a mask.


packages/axiom-charts/src/ChartContext/ChartContext.js
```diff
@@ -54,7 +55,8 @@ export default class ChartContext extends Component {
         <Dropdown
             onRequestClose={ onDropdownClose }
             onRequestOpen={ onDropdownOpen }
-            position="top">
+            position="top"
+            withMask>
```
lpoulter: Do we still need the mask if there is no tooltip context?

jdelight:
Works nicely and solves the BarChart issue - nice one @HHogg 

Only thought I have is one of semantics - mask implies something over the top whereas backdrop implies something behind... I'd vote for `withBackdrop` or `withMaskBehind` but its NABD :)

lpoulter:
FWIW I like mask here it is more the convention and it's covering something.

packages/axiom-charts/src/ChartContext/ChartContext.js
HHogg: There are is some styling on DotPlots and BarChart that needs to be retained when the Dropdown is open. Without the mask that styling is lost when hovering another part of the chart... maybe those interactions need to be turned off _within_ the chart instead though?

PR #614, opened 2018-07-05
feat(DatePicker): add quick selection

tomru:
## Adds quick selection ranges to the Date Picker.

The DatePicker can now offer _quick selections_ for ranges to the user.

These quick selections can be specified by setting the prop `rangeQuickSelections` in the form of e.g. 
```js
[
  {
    label: 'Today',
    range: 'P1D',
  },
  {
    label: '7D',
    range: 'P7D',
  },
  {
    label: '14D',
    range: 'P14D',
  },
  {
    label: '1M',
    range: 'P1M',
  },
  {
    label: '2M',
    range: 'P2M',
  },
]
```

The quick selection ranges are only shown if the component has
* `rangeSelect`
* `rangeQuickSelections` not set to `null` or an empty array.

When the user clicks on of these the `onSelect` callback is called with the `range` of the quick selection clicked.  It's then up to the application to convert the `range` into `startDate` and `endDate` and update these properties on the `DatePicker` so that the clicked range is selected.

This allows us to keep the application logic outside of axiom and keep the implementation completely agnostic of what these ranges mean in an application. (i.e. does 7D means 7*24h ago until now or start of the day 7 days ago until now....)

single

double

~~[Netlify Link](https://5b71551b4ed62f6d0c396c7e--youthful-albattani-41ebb2.netlify.com/docs/packages/axiom-components/date-picker)~~
~~[Netlify Link](https://5b72854dc96592106b141531--youthful-albattani-41ebb2.netlify.com/docs/packages/axiom-components/date-picker)~~
~~[Netlify Link](https://5b728f081f12b744b73c03d0--youthful-albattani-41ebb2.netlify.com/docs/packages/axiom-components/date-picker/)~~
[Netlify Link](https://5b87e5191f12b70f1a6bfb06--brandwatch-axiom-tomru.netlify.com/docs/packages/axiom-components/date-picker/)

### DIFF
--- a/packages/axiom-components/src/DatePicker/DatePicker.js
+++ b/packages/axiom-components/src/DatePicker/DatePicker.js
@@ -22,14 +22,25 @@ export default class DatePicker extends Component {
     onApply: PropTypes.func,
     /** Callback for when the cancel button has been clicked */
     onCancel: PropTypes.func,
-    /** Callback for when a date or range has been selected */
+    /** Callback for when a date, range with endDate and startDate, or a range
+     * out of rangeSelections has been selected */
     onSelect: PropTypes.func.isRequired,
     /** Whether a date range can be selected */
     rangeSelect: PropTypes.bool,
+    /** Predefined date ranges offered for selection. label will be shown and
+     * range is passed to onSelect when clicked. */
+    rangeSelections: PropTypes.arrayOf(
+      PropTypes.shape({
+        label: PropTypes.string.isRequired,
+        range: PropTypes.string.isRequired,
+      }),
+    ),
     /** A single date that appears selected */
     selectedDate: PropTypes.instanceOf(Date),
     /** The date selected at the end of the range */
     selectedEndDate: PropTypes.instanceOf(Date),
+    /** The selected range, specified as it's label */
+    selectedRange: PropTypes.string,
     /** The date selected at the start of the range */
     selectedStartDate: PropTypes.instanceOf(Date),
     /** Configuration for a single date picker view or two pickers side by side */
--- a/packages/axiom-components/src/DatePicker/DatePickerContext.js
+++ b/packages/axiom-components/src/DatePicker/DatePickerContext.js
@@ -2,19 +2,10 @@ import PropTypes from 'prop-types';
 import React, { Component } from 'react';
 import DropdownContent from '../Dropdown/DropdownContent';
 import DropdownContext from '../Dropdown/DropdownContext';
-import Grid from '../Grid/Grid';
-import GridCell from '../Grid/GridCell';
 import DatePickerControls from './DatePickerControls';
-import DatePickerHeaderControl from './DatePickerHeaderControl';
-import DatePickerViewMonth from './DatePickerViewMonth';
-import {
-  addMonths,
-  dateOrNow,
-  isAfterDay,
-  isBeforeDay,
-  isSameMonth,
-  orderDates,
-} from './utils';
+import DatePickerSelection from './DatePickerSelection';
+import DatePickerRangeSelection from './DatePickerRangeSelection';
+import SelectedDateOrRangePropTypes from './SelectedDateOrRangePropTypes';
 import './DatePicker.css';
 
 export default class DatePickerContext extends Component {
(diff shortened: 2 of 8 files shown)


tomru:
> Note: The quick selection Today does not set a dateEnd just a dateStart analog to the behaviour when you just click a single day.

Just noticing that you can set a range of the same day, when you click twice on the day. So 'TODAY' could do the same.

packages/axiom-components/src/DatePicker/DatePicker.js
```diff
@@ -39,6 +45,28 @@ export default class DatePicker extends Component {
   static defaultProps = {
     view: 'single',
     rangeSelect: false,
+    rangeQuickSelections: [
+      {
+        label: 'Today',
+        daysBeforeNow: 0,
```
tomru: I would have loved to go with [ISO 8601 durations](https://en.wikipedia.org/wiki/ISO_8601#Durations) here but date-fns does not yet support these :(

HHogg: @stefanpearson recommended us switching over to [luxon](https://github.com/moment/luxon) which does have support for it. Maybe now's a good time?

packages/axiom-components/src/DatePicker/DatePicker.js
```diff
@@ -24,6 +24,12 @@ export default class DatePicker extends Component {
     onCancel: PropTypes.func,
     /** Callback for when a date or range has been selected */
     onSelect: PropTypes.func.isRequired,
+    /** Ranges Quick Selections. Default is Today, 7D, 14D, 1M, 2M */
+    rangeQuickSelections: PropTypes.arrayOf(PropTypes.shape({
+      label: PropTypes.string.isRequired,
+      daysBeforeNow: PropTypes.number,
+      monthsBeforeNow: PropTypes.number,
```
HHogg: I feel like switching over to something that supports the ISO 8601 durations would also make these props a bit nicer, as they could be a bit more generic and could support future ranges, years, hours, minutes etc...

packages/axiom-components/src/DatePicker/DatePicker.js
tomru: Good call, will have a look into that 👍

packages/axiom-components/src/DatePicker/DatePickerQuickSelection.js
```diff
@@ -0,0 +1,38 @@
...
+    onRangeQuickSelection: PropTypes.func.isRequired,
+    rangeQuickSelections: PropTypes.arrayOf(PropTypes.shape({
+      label: PropTypes.string.isRequired,
+      daysBeforeNow: PropTypes.number,
+      monthsBeforeNow: PropTypes.number,
+    })).isRequired,
+  };
+
+  constructor(props) {
+    super(props);
```
HHogg: You don't need this constructor here :)

packages/axiom-components/src/DatePicker/DatePickerQuickSelection.css
```diff
@@ -0,0 +1,41 @@
...
+    width: 0.1875rem;
+    height: 0.1875rem;
+    margin-left: var(--spacing-grid-size--x3);
+    transform: translate(-50%, -50%);
+    border-radius: 50%;
+    background-color: var(--color-product-deep-thought);
+  }
+
+  &:hover {
+    color: var(--color-ui-accent);
```
HHogg: It would be nice to have some transitions here between the colours... ie 

```css
transition: color var(--transition-time-fast) var(--transition-function);
```

HHogg:
I think the double view is removed below a certain breakpoint.... maybe we should look into how to get the range selections to work on the single view. Ping @mozzling @Flawwles 

screen shot 2018-07-05 at 14 45 18

tomru:
@Flawwles struggling a little bit with this as the behaviour of quick selection is unclear to me

The design assumes the quick selection is a short cut for selecting date ranges manually, hence the selected date range should be shown as if you would have selected them manually. But this raises the question what these ranges actually mean, e.g. 
- is `Today`, start of today to now or start of today to end of today
- is `7D`,  today and the last 6 full days, or as in analytics today + the last 7 full days or even exactly 7 days ago until now (as far as I can tell vizia does that different currently as well @TomWija right?)

This is IMO this is down to the application logic what these quick selection mean. Hence we should not define in Axiom what these ranges are.

On top we can just visually select the day the range starts and the day the range ends. The user has no clue yet, what time on that day the range actually starts.

To get around defining what Today, 7D, etc. means can we make the quick selection rather a first class selection? Meaning when you select 7D, we won't highlight what 7D means in the picker?

In that case I think the design needs to emphasise a bit more that we have selected a range instead of a start and end date. Maybe show a textual description, like `Last 7 days` where the start and end date would be shown when you select them manually?

tomru:
@Flawwles You can ignore my last comment, as I've now changed the implementation to be free of deciding what e.g. 7D means. This is now up to the application consuming the DatePicker. Can you please review this on the netlify link in the description please?

@HHogg I've refactored this quite a bit now to keep the application logic out of axiom. This also means that in the DatePicker we do not actually use luxon to interpret the ISO 8061 durations anymore.

site/components/Documentation/Resources/Components/DatePicker.js
```diff
@@ -28,12 +68,16 @@ export default class Documentation extends Component {
...
-                setValue('DatePicker', 'selectedEndDate', dateEnd);
-                setValue('DatePicker', 'selectedStartDate', dateStart);
-                this.setState({ date, dateEnd, dateStart });
-              } }>
+              <DatePicker
+                  onSelect={ (setValue, getValue, { date, dateStart, dateEnd, range }) => {
+                    this.handleSelect({ date, dateEnd, dateStart, range });
+                  } }
+                  rangeSelect
+                  rangeSelections={ rangeSelections }
```
tomru: I've added the `rangeSelections` to both examples. 

As far as I can see on example is showing single view the other double view. Wondering if I should rather make 4 examples covering all variants, or rather drop one and rely on _you need to modify the settings_.

I think to show how to use it, it would be good add the `rangeSelections` at least at one place though...

tomru:
Now rebased after https://github.com/BrandwatchLtd/axiom/pull/628 was merged.

packages/axiom-components/src/DatePicker/DatePickerRangeSelection.css
```diff
@@ -0,0 +1,45 @@
+.ax-date-range-selection {
```
lpoulter: There's a bit of jumping around here when the month spans 5 or 6 rows.

packages/axiom-components/src/DatePicker/DatePicker.js
```diff
@@ -22,14 +22,25 @@ export default class DatePicker extends Component {
...
+    /** Predefined date ranges offered for selection. label will be shown and
+     * range is passed to onSelect when clicked. */
+    rangeSelections: PropTypes.arrayOf(
+      PropTypes.shape({
+        label: PropTypes.string.isRequired,
+        range: PropTypes.string.isRequired,
+      }),
+    ),
     /** A single date that appears selected */
     selectedDate: PropTypes.instanceOf(Date),
```
lpoulter: Getting some prop type errors on local. `selectedDate` is being set to a blank string when in range mode I think. Which is then passed into the child components.

packages/axiom-components/src/DatePicker/DatePickerContext.js
```diff
@@ -59,6 +68,22 @@ export default class DatePickerContext extends Component {
     };
   }
 
+  componentDidUpdate(prevProps) {
+    const { selectedStartDate, selectedEndDate, view } = this.props;
+
```
lpoulter: Not sure I really understand why we need to mirror the props `selectedStartDate` & `selectedEndDate` into their `active` equivalents state here?

packages/axiom-components/src/DatePicker/DatePickerContext.js
```diff
@@ -127,7 +167,46 @@ export default class DatePickerContext extends Component {
     const [ startDate, endDate ] = orderDates(selectedStartDate, selectedEndDate);
     const [ earliestDate, latestDate ] = orderDates(earliestSelectableDate, latestSelectableDate);
 
-    const isDoubeView = view === 'double';
+    const isDoubleView = view === 'double';
+    const showRangeSelection = rangeSelect && rangeSelections && rangeSelections.length;
+
+    const rangeSelection = showRangeSelection && (isDoubleView ? (
```
lpoulter: Theres a lot of logic the pivots on weather `rangeSelect` is applied. Have you tried splitting out some of this into a separate component? The switching between a single date, `selectedDate` and a range of dates, `selectedStartDate` & `selectedEndDate` is quite hard to follow.

packages/axiom-components/src/DatePicker/DatePickerContext.js
```diff
@@ -105,6 +130,19 @@ export default class DatePickerContext extends Component {
     });
   }
 
+  handleRangeSelection(range) {
+    const {
+      onSelect,
+      rangeSelect,
+    } = this.props;
+
+    if (!rangeSelect || !range) {
```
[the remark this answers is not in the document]
lpoulter: Look like we are already checking this in the render.

packages/axiom-components/src/DatePicker/DatePickerContext.js
lpoulter: `showRangeSelection && (isDoubleView ? (` is the `(` before isDoubleView doing any thing here?

lpoulter: Perhaps this template could be simplified if we had a `DatePickerView` component that rendered one or two DatePickerHeaderControl/DatePickerViewMonth pairs based on a double prop.

packages/axiom-components/src/DatePicker/DatePickerContext.js
```diff
@@ -59,6 +68,22 @@ export default class DatePickerContext extends Component {
...
   }
 
+  componentDidUpdate(prevProps) {
+    const { selectedStartDate, selectedEndDate, view } = this.props;
+
+    if (selectedStartDate !== prevProps.selectedStartDate) {
+      this.setState({ activeStartDate: selectedStartDate });
+    }
+
+    if (view === 'double' && selectedEndDate !== prevProps.selectedEndDate) {
```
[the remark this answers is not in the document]
tomru: As this part only sets `activeEndDate` which affects only the second month grid, I think it makes sense to check if we are in double view. I agree it makes sense to check for `rangeSelect` additionally to keep the state clean.

packages/axiom-components/src/DatePicker/DatePickerContext.js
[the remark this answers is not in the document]
tomru: Depends, if you pass `rangeSelections` according to the prop types. But in that case it's probably your fault as well, when `onSelect` is called without a range value. 

Dropping 👍

packages/axiom-components/src/DatePicker/DatePickerContext.js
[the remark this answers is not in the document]
tomru: Yeah, it's not pretty, but couldn't think of another way of doing this currently.

packages/axiom-components/src/DatePicker/DatePickerRangeSelection.css
tomru: True, I've noticed that as well. Though this hasn't really change from how it was before as far as I can tell.

Even when we would reserve the space for 6 rows to get rid of the jumping, it would look weird in the 4 col case (which applies more often).

Any thoughts on this @Flawwles ?

packages/axiom-components/src/DatePicker/DatePickerContext.js
tomru: The `active[Start|End]Date` do control which month is shown on both sides of the picker.

When setting `selected[Start|End]Date` props on the picker, after selecting a month, the user should see the start and end of the selected range. Hence I need to update both moth grids as well.

packages/axiom-components/src/DatePicker/DatePickerContext.js
tomru: `&&` has a higher operator precedence than the conditional operator, so I'm forced to use a bracket here if I'm not missing anything.

tomru: Might be worth splitting it out, but I'm concerned it won't reduce the complexity a lot, it would just move it to another file. I think to reduce the complexity it might be worth to to split the date picker into a `SingleDatePicker` and a `RangeDatePicker` to reduce the different cases.

The `SingleDatePicker` could be free of all the range related stuff, as well as the `RangeDatePicker` could be free of the single date stuff ?

lpoulter: I've not seen this sort of ternary before. This is confusing but I think there is a bug here.
```
let showRangeSelection = false;
let isDoubleView = false;
showRangeSelection && (isDoubleView ? 'trueCase' : 'falseCase');
returns false
```

lpoulter: Either way I think it would be clear if it was split out.

tomru: doh! brackets belong here `(showRangeSelection && isDoubleView) ? ... ` 

Thanks for pointing that out!

tomru:
@lpoulter @HHogg I've split up `DatePickerContext` a bit, by
- moving the gnarly conditional for responsiveness down into `DatePickerRangeSelection`
- Extracted displaying single and double _day in month pickers_ into `DatePickerSelection` - I'm still unsure about the the name tbh

Could you have another look please ?

tomru:
@Flawwles IIRC the thing left was the question if the documentation is ok ( one opinion [here](https://github.com/BrandwatchLtd/axiom/pull/614#discussion_r209947160) ). Anything else that needs to be addressed ?

packages/axiom-components/src/DatePicker/SelectedDateOrRangePropTypes.js
```diff
@@ -0,0 +1,44 @@
...
+}
+
+function isEmptyOrDate(props, propName) {
+  const value = props[propName];
+
+  if (!value) {
+    return;
+  }
+
+  if (value instanceof Date && !isNaN(value)) {
```
HHogg: Could these custom propTypes make use `PropTypes.instanceOf` somehow?

tomru: They could, but it wouldn't detect
- invalid dates
- invalid combinations of `rangeSelect`, `selectedDate`, `selectedStartDate` and `selectedEndDate`

So I think this should be superiorly

PR #615, opened 2018-07-05
feat(Slider): Added slider component

HHogg:
slider

### DIFF
--- a/packages/axiom-components/src/Slider/Slider.css
+++ b/packages/axiom-components/src/Slider/Slider.css
@@ -1,0 +1,82 @@
+:root {
+  --ax-slider-track-size--small: var(--spacing-grid-size--x1);
+  --ax-slider-track-size--medium: var(--spacing-grid-size--x2);
+  --ax-slider-thumb-size--small: var(--spacing-grid-size--x4);
+  --ax-slider-thumb-size--medium: var(--spacing-grid-size--x6);
+  --ax-slider-thumb-border-width: 0.0625rem;
+}
+
+.ax-slider {
+  position: relative;
+  outline: none;
+  user-select: none;
+}
+
+.ax-slider__track {
+  position: absolute;
+  top: 50%;
+  right: 0;
+  left: 0;
+  transform: translateY(-50%);
+  background-color: var(--color-theme-background--subtle);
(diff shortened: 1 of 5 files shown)


packages/axiom-components/src/Slider/Slider.js
```diff
@@ -0,0 +1,147 @@
...
+    case 'ArrowDown':
+      delta = -1;
+      break;
+    case 'ArrowRight':
+    case 'ArrowUp':
+      delta = 1;
+      break;
+    }
+
+    onChange(Math.max(min, Math.min(value + step * delta, max)));
```
lpoulter: Out of curiosity why do we need the delta here?

packages/axiom-components/src/Slider/Slider.css
```diff
@@ -0,0 +1,81 @@
...
+  user-select: none;
+}
+
+.ax-slider__track {
+  position: absolute;
+  top: 50%;
+  right: 0;
+  left: 0;
+  transform: translateY(-50%);
+  background-color: var(--color-theme-background--subtle);
```
lpoulter: I wonder if we want `cursor: pointer` here also as clicking on the track moves the slider.

lpoulter:
Looks great. We have a slider in the `Word cloud` app that has a text input box to allow greater precision. It looks like that could be added at a later date thou.

packages/axiom-components/src/Slider/Slider.js
HHogg: It sets the direction for when using the arrow keys. It was to shortcut writing `onChange(Math.max(min, Math.min(value +/- step, max)));` twice. Happy to change if the other way is more clearer though.

lpoulter: Ok thanks for the explanation. The second way is a little clearer to me.

packages/axiom-components/src/Slider/Slider.css
HHogg: Good shout. Have added in 👍

HHogg:
Ooops I forgot to add [Netlify link](https://slider--hhogg-axiom.netlify.com/)

PR #616, opened 2018-07-05
AX-31 progress indicators on form components

TomWija:
https://5b754700dd28ef26c0ced7d1--axiomwija.netlify.com/docs/packages/axiom-components/form

### DIFF
--- a/packages/axiom-components/src/Form/InputWrapper.js
+++ b/packages/axiom-components/src/Form/InputWrapper.js
@@ -14,12 +14,19 @@ const labelSizeMap = {
 export default class InputWrapper extends Component {
   static propTypes = {
     children: PropTypes.node,
+    disabled: PropTypes.bool,
+    hasFocus: PropTypes.bool,
+    inputType: PropTypes.string,
+    invalid: PropTypes.bool,
     isTarget: PropTypes.bool,
+    isValid: PropTypes.bool,
     label: PropTypes.node,
     size: PropTypes.oneOf(['small', 'medium', 'large']),
     space: PropTypes.string,
+    style: PropTypes.string,
     usageHint: PropTypes.node,
     usageHintPosition: PropTypes.oneOf(['top', 'bottom', 'left', 'right']),
+    valid: PropTypes.bool,
   };
 
   static defaultProps = {
@@ -28,10 +35,18 @@ export default class InputWrapper extends Component {
   };
 
   render() {
-    const { children, isTarget, label, size, space, usageHint, usageHintPosition, ...rest } = this.props;
-
+    const { children, disabled, hasFocus, invalid, isTarget, isValid, label, size, space, style, usageHint, usageHintPosition, valid, ...rest } = this.props;
     const classes = classnames('ax-input__wrapper', {
       'ax-input__wrapper--target': isTarget,
+    });
+
+    const innerClasses = (isValid) => classnames('ax-input__icon-container', {
+      [`ax-input__icon-container--${size}`]: size,
+      [`ax-input__icon-container--${style}`]: style,
+      'ax-input__icon-container--disabled': disabled,
+      'ax-input__icon-container--focused': hasFocus,
+      'ax-input__icon-container--valid': valid,
+      'ax-input__icon-container--invalid': invalid || isValid === false,
     });
 
     return (
@@ -62,7 +77,9 @@ export default class InputWrapper extends Component {
             </Base>
           )
         }
-        { children }
+        <div className={ innerClasses(isValid) }>
+          { children }
+        </div>
       </Base>
     );
   }
--- a/packages/axiom-components/src/Form/TextArea.css
+++ b/packages/axiom-components/src/Form/TextArea.css
@@ -1,6 +1,6 @@
 .ax-textarea {
   padding: var(--component-padding-vertical-medium) var(--component-padding-horizontal-small);
-  border: var(--component-border-width) solid var(--color-theme-border);
+  border: 0;
   border-radius: var(--component-border-radius);
   background-color: var(--color-theme-background);
   resize: none;
@@ -23,4 +23,10 @@
   &::placeholder {
     color: var(--color-ui-error);
   }
+}
+
+.ax-textarea__progress {
+  align-self: start;
+  padding-top: var(--component-padding-horizontal-small);
+  padding-right: var(--component-padding-horizontal-small);
 }
--- a/packages/axiom-components/src/Form/TextArea.js
+++ b/packages/axiom-components/src/Form/TextArea.js
@@ -4,6 +4,7 @@ import classnames from 'classnames';
 import Base from '../Base/Base';
 import Validate from '../Validation/Validate';
 import InputWrapper from './InputWrapper';
+import ProgressInfinite from '../Progress/ProgressInfinite';
 import './TextArea.css';
 
 export default class TextArea extends Component {
@@ -16,8 +17,14 @@ export default class TextArea extends Component {
     height: PropTypes.number,
     /** Applies styling to indicate the users input was invalid */
     invalid: PropTypes.bool,
+    /** Adds a progress indicator the the right of the text input */
+    isInProgress: PropTypes.bool,
     /** Descriptive label that is placed with the input field */
     label: PropTypes.node,
+    /** Handler for when the input field is blurred */
+    onBlur: PropTypes.func,
+    /** Handler for when the input field is focused */
+    onFocus: PropTypes.func,
     /** See Validate[patterns] */
     patterns: PropTypes.array,
     /**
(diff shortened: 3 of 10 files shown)


packages/axiom-components/src/Form/TextArea.js
```diff
@@ -44,12 +51,30 @@ export default class TextArea extends Component {
     height: 150,
   };
 
+  constructor(props) {
+    super(props);
+    this.state = { hasFocus: false };
+  }
+
```
lpoulter: NABD. We don't need to add `props` the `constructor` signature nor add them to the `props` call if we don't edit them with in the constructor. I'd prefer setting state on the class directly in this case. Not sure how that fits in with `Axiom` style.

packages/axiom-components/src/UsageHint/UsageHint.js
```diff
@@ -14,16 +14,22 @@ export default class UsageHint extends PureComponent {
...
   render() {
     const {
       children,
       position,
+      showArrow,
     } = this.props;
 
     return (
-      <Dropdown position={ position }>
+      <Dropdown position={ position } showArrow={ showArrow }>
```
lpoulter: I think this can just be

packages/axiom-components/src/Form/TextArea.js
TomWija: actually based this off of the usage of `hasFocus` that existed in the TextInput component already: https://github.com/BrandwatchLtd/axiom/blob/master/packages/axiom-components/src/Form/TextInput.js#L73

packages/axiom-components/src/Form/TextArea.js
```diff
@@ -44,12 +51,30 @@ export default class TextArea extends Component {
...
+    this.setState({ hasFocus: false });
+    onBlur(...arguments);
+  }
+
+  handleOnFocus() {
+    const { onFocus = () => {} } = this.props;
+    this.setState({ hasFocus: true });
+    onFocus(...arguments);
+  }
+
```
lpoulter: Why are we adding onBlur and focus to add a progress or tooltip, are they just things that should have been added before?

TomWija: The addition of a progress indicator required me to add some wrapping divs so that it could look like it was inside the text area itself. This just made it require focus/blur logic similar to TextInput which already accounted for this

packages/axiom-components/src/UsageHint/UsageHint.js
HHogg: Also everything dropdown can be configured (offsets, boundary element... etc) ideally needs spreading on for the future. 

```jsx
const  { children, ...rest } = this.props;

<Dropdown { ...rest } showArrow>...</Dropdown>
```

packages/axiom-components/src/Form/TextInput.js
```diff
@@ -157,6 +160,13 @@ export default class TextInput extends Component {
                     onFocus={ this.handleOnFocus.bind(this) }
                     type={ type }
                     value={ value } />
+                {
+                  isInProgress && (
```
HHogg: Does the conditional on the `<TextInputIcon>` needs adjusting slightly so the icon and the progress are not rendered at the same time?

packages/axiom-components/src/Form/TextArea.js
```diff
@@ -74,13 +108,26 @@ export default class TextArea extends Component {
...
         { (isValid) =>
           <InputWrapper label={ label } usageHint={ usageHint } usageHintPosition={ usageHintPosition }>
-            <Base { ...rest }
-                Component="textarea"
-                className={ classes(isValid) }
-                disabled={ disabled }
-                placeholder={ placeholder }
-                style={ style }
-                value={ value } />
+            <div className="ax-input__button-container">
```
HHogg: `ax-input__button-container` doesn't really make sense here as there is no button.

packages/axiom-components/src/Form/TextArea.js
```diff
@@ -60,8 +85,17 @@ export default class TextArea extends Component {
...
+    const classes = (isValid) => classnames('ax-input__icon-container', {
+      'ax-input__icon-container--disabled': disabled,
+      'ax-input__icon-container--focused': hasFocus,
+      'ax-input__icon-container--valid': valid,
+      'ax-input__icon-container--invalid': invalid || isValid === false,
+    });
+
+    const textareaClasses = (isValid) => classnames('ax-textarea', {
       'ax-textarea--valid': valid,
       'ax-textarea--invalid': invalid || isValid === false,
```
HHogg: These classes are no longer needed as the border color changes is added by the classes above right?

packages/axiom-components/src/Form/TextArea.js
```diff
@@ -60,8 +85,17 @@ export default class TextArea extends Component {
       value,
       ...rest } = this.props;
 
+    const { hasFocus } = this.state;
+
     const style = { minHeight: height };
-    const classes = (isValid) => classnames('ax-textarea', {
+    const classes = (isValid) => classnames('ax-input__icon-container', {
```
HHogg: We've kept to specifying classes in two places, once in a single component and in a single CSS file. This helps to reduce maintenance when the classes/templates/styles change. 

Is there a way a general component (similar to the InputWrapper) can be  created to get this same behaviour/styling for both Input and TextArea?

packages/axiom-components/src/Form/TextArea.js
HHogg: It's to handle adding focus border-color to the parent wrapping the actual input... [we don't need focus styling in IE/Edge right?](https://caniuse.com/#feat=css-focus-within)  😄

packages/axiom-components/src/Form/TextArea.js
TomWija: I thought so too, but the focus stuff didn't work for the color of the text without this 🤔

packages/axiom-components/src/Form/InputWrapper.js
```diff
@@ -14,7 +14,10 @@ const labelSizeMap = {
 export default class InputWrapper extends Component {
   static propTypes = {
     children: PropTypes.node,
+    classProps: PropTypes.object,
```
HHogg: This is a bit of an odd pattern. Why not add the props onto the component in the standard way? 

```jsx
<InputWrapper disabled={ disabled } focused={ focused } size={ size }  />
```

packages/axiom-components/src/Form/TextArea.css
```diff
@@ -24,3 +24,9 @@
     color: var(--color-ui-error);
   }
 }
+
+.ax-textarea-progress {
```
HHogg: Following BEM the textarea would be the block and progress would be the element. So this should be more like 

```
.ax-textarea__progress
```

packages/axiom-components/src/Form/TextArea.css
```diff
@@ -1,6 +1,6 @@
 .ax-textarea {
   padding: var(--component-padding-vertical-medium) var(--component-padding-horizontal-small);
-  border: var(--component-border-width) solid var(--color-theme-border);
+  border: 0;
```
HHogg: There are `border-color` styles specified below which are no longer being used.

packages/axiom-components/src/Form/TextArea.js
```diff
@@ -73,14 +108,23 @@ export default class TextArea extends Component {
           required={ required }
           value={ value }>
         { (isValid) =>
-          <InputWrapper label={ label } usageHint={ usageHint } usageHintPosition={ usageHintPosition }>
+          <InputWrapper classProps={ classProps } isValid={ isValid } label={ label } usageHint={ usageHint } usageHintPosition={ usageHintPosition }>
             <Base { ...rest }
                 Component="textarea"
-                className={ classes(isValid) }
+                className={ textareaClasses(isValid) }
```
HHogg: This can now just be set to `className="ax-textarea"` as the the dynamic classes have been offloaded onto the InputWrapper and are no longer doing anything.

packages/axiom-components/src/Form/TextInput.css
```diff
@@ -18,6 +18,10 @@
   }
 }
 
+.ax-input-progress {
```
HHogg: Same comment here about the BEM syntax `ax-input__progress`

packages/axiom-components/src/Form/TextInput.js
```diff
@@ -91,13 +94,27 @@ export default class TextInput extends Component {
...
 
+  handleShowIcon(isInProgress, onClear, value, icon) {
+    if (!isInProgress && onClear && value) {
+      return (
+        <TextInputIcon
+            align="right"
+            name="cross"
+            onClick={ onClear } />
+      );
+    } else if (icon && (!isInProgress || icon.props.align === 'left')) {
```
HHogg: We tend to not split up the rendering logic away from the rending function so it's easier to see when reading from top to bottom what is being output without needing to jump around the file.

HHogg:
The template changes to the input have slightly 
screen shot 2018-07-11 at 16 01 50
broken the right icon alignment.

packages/axiom-components/src/Form/TextArea.js
lpoulter: Clarified this in person. This was to fix an ie bug with focus. Would be good to put this sort of stuff in the description, makes it easier to review.

PR #621, opened 2018-07-16
feat(ArrowHead): Add an ArrowHead component to axiom.

MariamBashir:
feat(ArrowHead): Added an ArrowHead component.

This PR implements a new component which lets you add an arrow head on another component. 

The arrow head can be configured to strick on any side and position. The user also has the ability to change the colour of the component.

### DIFF
--- a/packages/axiom-components/src/ArrowHead/ArrowHead.js
+++ b/packages/axiom-components/src/ArrowHead/ArrowHead.js
@@ -1,0 +1,116 @@
+import PropTypes from 'prop-types';
+import React, { Component } from 'react';
+import { findDOMNode } from 'react-dom';
+import classnames from 'classnames';
+import Base from '../Base/Base';
+import './ArrowHead.css';
+
+/* eslint-disable react/no-find-dom-node */
+export default class ArrowHead extends Component {
+  static propTypes = {
+    /** Contents to be displayed in the arrow container */
+    children: PropTypes.node.isRequired,
+    /** Background color for the arrow */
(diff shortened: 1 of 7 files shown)


lpoulter:
Strange how it show's me as authored on this PR. Do you have a netlify link?

MariamBashir:
@lpoulter i think may be i messed up something. I had to rebase and then made the push. And i dont have the link because it failed to build there too. Ill try to fix it. Sorry for that.

packages/axiom-components/src/ArrowHead/ArrowHead.js
```diff
@@ -0,0 +1,116 @@
...
+
+  componentDidMount() {
+    if (this._content && this._arrow) {
+      const { position, direction } = this.props;
+      this.updateArrowPlacement();
+      this.updateArrowPosition(position, direction);
+    }
+  }
+
+  componentWillReceiveProps(nextProps) {
```
lpoulter: This lifecycle method is being Deprecated you might want to use `getDerivedStateFromProps ` instead https://reactjs.org/docs/react-component.html#unsafe_componentwillreceiveprops. You will also want to check the props that have changed and that they are in-fact different from the last values.

MariamBashir: Yeah i saw that and wanted to replace it. I am closing this PR. I think ill rebase properly and add the changes with the new ```react``` version. Thanks.

PR #622, opened 2018-07-17
feat(ArrowHead): Add an ArrowHead component to axiom.

MariamBashir:
feat(ArrowHead): Added an ArrowHead component.

This PR implements a new component which lets you add an arrow head on another component.

The arrow head can be configured to point to any direction and position. The user also has the ability to change the colour of the arrow head component.

I made a PR before and majorly messed up the rebasing and pushed code that was in a completely bad state.  I am sorry for that. 

Here is the netlify link : https://5b55cfa8c6aed668d7a5e539--affectionate-mcclintock-42eb2e.netlify.com/
Thanks.


packages/axiom-components/src/ArrowHead/ArrowHead.js
```diff
@@ -0,0 +1,116 @@
...
+import Base from '../Base/Base';
+import './ArrowHead.css';
+
+/* eslint-disable react/no-find-dom-node */
+export default class ArrowHead extends Component {
+  static propTypes = {
+    /** Contents to be displayed in the arrow container */
+    children: PropTypes.node.isRequired,
+    /** Background color for the arrow */
+    color: PropTypes.string.isRequired,
```
lpoulter: We will probably want to restrict this color to `onOff` the Axiom one's in `@brandwatch/axiom-materials`

packages/axiom-components/src/ArrowHead/ArrowHead.js
```diff
@@ -0,0 +1,116 @@
...
+    /** Sets the direction where the arrow points */
+    direction: PropTypes.oneOf(['top', 'bottom', 'left', 'right']),
+    /** Relative position of the arrow with respect to the left or top depending on the position
+     * The value should be between 0 and 1.*/
+    position: PropTypes.number.isRequired,
```
lpoulter: It would be nice to validate this number. You should be able to generate a between a range array 0-1 and use `onOff`

packages/axiom-components/src/ArrowHead/ArrowHead.js
```diff
@@ -0,0 +1,116 @@
...
+    this.setState({
+      relativePosition: (position * content.height) - (position * arrow.height),
+    });
+  };
+
+  componentDidMount() {
+    if (this._content && this._arrow) {
+      const { position, direction } = this.props;
+      this.updateArrowPlacement();
+      this.updateArrowPosition(position, direction);
```
lpoulter: I doesn't look like these args are used in the `updateArrowPosition` method.

packages/axiom-components/src/ArrowHead/ArrowHead.js
```diff
@@ -0,0 +1,116 @@
...
+      });
+    }
+
+    this.setState({
+      relativePosition: (position * content.height) - (position * arrow.height),
+    });
+  };
+
+  componentDidMount() {
+    if (this._content && this._arrow) {
```
lpoulter: I don't think you need this check here. You shouldn't if you use callback refs more here. https://reactjs.org/docs/refs-and-the-dom.html#callback-refs

packages/axiom-components/src/ArrowHead/ArrowHead.js
```diff
@@ -0,0 +1,116 @@
...
+    const arrowClasses = classnames('ax-arrow-head--arrow', {
+      'ax-arrow-head--arrow--top': direction === 'top',
+      'ax-arrow-head--arrow--left': direction === 'left',
+      'ax-arrow-head--arrow--right': direction === 'right',
+      'ax-arrow-head--arrow--bottom': direction === 'bottom',
+    });
+
+    return (
+      <Base { ...rest }>
+        <Base className={ arrowClasses } ref={ (node) => {
```
lpoulter: I think it's best practice to add these functions to the class would make it more readble also.

lpoulter: I think it's best practice to add these ref functions to the class would make it more readable also.

packages/axiom-components/src/ArrowHead/ArrowHead.js
```diff
@@ -0,0 +1,116 @@
...
+    case 'bottom':
+      return { left: relativePosition, borderTopColor: color, bottom: containerOffset };
+    case 'left':
+      return { top: relativePosition, borderRightColor: color, left: containerOffset };
+    case 'right':
+      return { top: relativePosition, borderLeftColor: color, right: containerOffset };
+    }
+  };
+
+  updateArrowPlacement = () => {
```
lpoulter: Both this and `updateArrowPosition` are very similarly named and seem to do simular things. Is it possible to refactor to have one method that returns a position object. Somthing like

packages/axiom-components/src/ArrowHead/ArrowHead.js
MariamBashir: Was suppose to remove them. 👎 Sorry.

packages/axiom-components/src/ArrowHead/ArrowHead.js
MariamBashir: I am concerned about the resolution of such an array? Do you mean something like this 
```PropTypes.oneOf([0, 0.1, 0.2, 0.3, 0.4, 0.5, 0.6, 0.7, 0.8, 0.9, 1])```?

Comment (lpoulter) [reply packages/axiom-components/src/ArrowHead/ArrowHead.js]: Yeah you can generate it if you want smaller increments. Something like `const zeroToOneRange = Array.from({length: 100}, (ele, idx) => (idx * 0.01).toFixed(2))`

File: packages/axiom-components/src/ArrowHead/ArrowHead.js (hunk 4 above)
Comment (MariamBashir) [reply packages/axiom-components/src/ArrowHead/ArrowHead.js]: i checked and for some reason my test always fails if i dont keep the check.

**Comment by MariamBashir** [comment]
@lpoulter Thank you for the feedback. I have made some changes based on that. I was hoping to know if you can point in the right direction regarding the design review. Who might be the responsible person?

**Comment by lpoulter** [comment]
@mozzling or Matt Evens might be able with the design review.

File: packages/axiom-components/src/ArrowHead/ArrowHead.js (hunk 4 above)
Comment (MariamBashir) [reply packages/axiom-components/src/ArrowHead/ArrowHead.js]: Actually i cant use ```createRef``` because my netlify build keeps failing if i do that. I am going to use callback refs made in class like you suggested earlier. 

screen shot 2018-07-17 at 18 06 59

Comment (lpoulter) [reply packages/axiom-components/src/ArrowHead/ArrowHead.js]: You will need to rebase, I recently bumped out react dep, to use `createRef` as it was added in `16.3` using the callback version is fine also. Feel free to slack me if you need a hand.

Comment (lpoulter) [reply packages/axiom-components/src/ArrowHead/ArrowHead.js]: You might also need to update `react-test-renderer` if you use later react features. https://github.com/BrandwatchLtd/axiom/pull/620/files#diff-b9cfc7f2cdf78a7f4b91a753d10865a2L64

File: packages/axiom-components/src/ArrowHead/ArrowHead.js [hunk 7]
```diff
@@ -0,0 +1,135 @@
...
+      this.setArrowPosition();
+    }
+  }
+
+  render() {
+    const { color, children, direction, opacity, ...rest } = this.props;
+
+    const arrowClasses = classnames('ax-arrow-head--arrow', {
+      [`ax-arrow-head--${color}`]: color,
+      'ax-arrow-head--arrow--top': direction === 'top',
```
lpoulter: you should be able to use `direction` directly in the string to simplify this e.g. 'ax-arrow-head--arrow--top': direction === 'top',

packages/axiom-components/src/ArrowHead/ArrowHead.js
```diff
@@ -0,0 +1,132 @@
...
+    /** Background color for the arrow */
+    color: PropTypes.oneOf([
+      'faded',
+      'blast-off',
+      'crash-course',
+      'solar-rust',
+      'ground-control',
+      'space-oddity',
+      'rocky-planet',
+      'deep-thought',
```
HHogg: This is quite a cut down list of colours that are available. Inside context the colours only seem to be the alerting colours. Is there a need for the product colours at all, and do we need to allow for all of the product colours to be selectable? (Probably a question for design)

packages/axiom-components/src/ArrowHead/ArrowHead.js
```diff
@@ -0,0 +1,132 @@
...
+    ]).isRequired,
+    /** Sets the direction where the arrow should point */
+    direction: PropTypes.oneOf(['top', 'bottom', 'left', 'right']),
+    /**  Adjusts the opcaity value for the Arrow head background color*/
+    opacity: PropTypes.number.isRequired,
+    /** Relative position of the arrow with respect to the left or top depending on the position
+     * The value should be a string between 0 and 1.*/
+    position: PropTypes.oneOf(['0', '0.1', '0.2', '0.3', '0.4', '0.5', '0.6', '0.7', '0.8', '0.9', '1']),
```
HHogg: I don't think this is the most flexible way to allow the positions to be defined. If it allows just take a standard CSS value/unit (px/rem) it can be places anywhere along the axis (Take a look at the current context component).

packages/axiom-components/src/ArrowHead/ArrowHead.js
```diff
@@ -0,0 +1,132 @@
...
+    }
+  };
+
+  setArrowContenOffset = () => {
+    const { direction } = this.props;
+    const contentTop = getContainerDimensions(this._content)[direction];
+    const arrow = getContainerDimensions(this._arrow);
+
+    if (direction === 'top' || direction === 'bottom') {
+      return this.setState({ containerOffset: contentTop - (arrow[direction] + arrow.height) });
```
HHogg: I'm not sure we need to use state for this, as we don't need to cause a rerender. As you have the element under refs you could add the styles on directly.

packages/axiom-components/src/ArrowHead/ArrowHead.js
```diff
@@ -0,0 +1,132 @@
...
+    /** Relative position of the arrow with respect to the left or top depending on the position
+     * The value should be a string between 0 and 1.*/
+    position: PropTypes.oneOf(['0', '0.1', '0.2', '0.3', '0.4', '0.5', '0.6', '0.7', '0.8', '0.9', '1']),
+  };
+
+  static defaultProps = {
+    direction: 'top',
+    position: '0.4',
+    color: 'faded',
+    opacity: 0.4,
```
HHogg: So we need an opacity on this? Axiom generally controls all the UI colour opacities through the provided UI colours.  [For example](https://github.com/BrandwatchLtd/axiom/blob/master/packages/axiom-materials/src/theme-day.css#L28)

packages/axiom-components/src/ArrowHead/ArrowHead.css
```diff
@@ -0,0 +1,139 @@
+.ax-arrow-head--arrow {
```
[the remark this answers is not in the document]
MariamBashir: Makes sense. Ill update that.

packages/axiom-components/src/ArrowHead/ArrowHead.js
MariamBashir: I actually had percentages before but @lpoulter suggested that i should probably provide a an array of values to select from with a suitable resolution.

packages/axiom-components/src/ArrowHead/ArrowHead.js
MariamBashir: Yeah i will try to confirm that and then update this list. I had quite a lot of doubts about these myself.

packages/axiom-components/src/ArrowHead/ArrowHead.css
```diff
@@ -0,0 +1,139 @@
...
+  position: absolute;
+  width: 0;
+  height: 0;
+}
+
+.ax-arrow-head--arrow--top {
```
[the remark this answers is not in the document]
MariamBashir: Ill check this and update it. Should have read about the BEM syntax more before using it in the code. Sorry!

packages/axiom-components/src/ArrowHead/ArrowHead.js
lpoulter: My concern was that there should be some validation to reflect the fact we are only accepting numbers between 0-1.

lpoulter: % are fine also  but we should change the prop type and description to reflect that.

MariamBashir:
@HHogg Just so i am clear about what you are saying, i would like to run it through you before i make any changes. 
 ```popperJS``` is only used in the position component as per my knowledge.  You are suggesting that we could remove the logic for creating popper and replace it with the ```Tip Component``` ?  And the ```Tip Component``` stays  like a separate component which can be leveraged at other places too.

packages/axiom-components/src/ArrowHead/ArrowHead.js
MariamBashir: @lpoulter I actually agree with the both ways and have no issues. And you are right, the description definitely needs some work to better reflect whats the acceptable configuration and why.

MariamBashir:
In this commit i add a ```Tip``` component. This is a temporary solution to get an idea if the direction is right. I would like to add more implementation on top.

MariamBashir:
@lpoulter and @HHogg Please have another look at the PR, it would be great if we can move this faster because we need to add this to the component picker and the deadline is already on Tuesday. Thanks.

packages/axiom-components/src/Tip/__snapshots__/Tip.test.js.snap
```diff
@@ -0,0 +1,1276 @@
+// Jest Snapshot v1, https://goo.gl/fbAQLP
+
+exports[`Tip renders with color 1`] = `
+ShallowWrapper {
```
tomru: I'd expect it to just have markup, instead of the `ShallowWrapper` instance. Not sure how this happened...

packages/axiom-components/src/Context/Context.js
```diff
@@ -16,13 +18,15 @@ export default class Context extends Component {
   };
 
   static defaultProps = {
+    arrowPlacement: '5rem',
```
tomru: Wouldn't a default of `35%` be saver, in case someone set's a width of e.g. `4rem` ?

packages/axiom-components/src/Tip/Tip.js
```diff
@@ -0,0 +1,54 @@
...
+  static propTypes = {
+    arrowRef: PropTypes.func,
+    children: PropTypes.node.isRequired,
+    color: PropTypes.oneOf(['success', 'warning', 'error', 'info', 'faded']),
+    placement: PropTypes.string,
+    position: PropTypes.oneOf(['top', 'bottom', 'left', 'right']),
+  };
+
+  static defaultProps = {
+    placement: '6.5rem',
```
tomru: Same as above, I'd rather go with `50%` as a default ?

packages/axiom-components/src/Tip/Tip.js
```diff
@@ -0,0 +1,54 @@
+import PropTypes from 'prop-types';
+import React, { Component } from 'react';
+import classnames from 'classnames';
+import Base from '../Base/Base';
+import './Tip.css';
+
+export default class Tip extends Component {
+  static propTypes = {
+    arrowRef: PropTypes.func,
```
tomru: Shouldn't `arrowRef` be required ?

packages/axiom-components/src/Tip/Tip.js
```diff
@@ -0,0 +1,54 @@
...
+    position: 'top',
+    color: 'faded',
+  };
+
+  getArrowStyles = () => {
+    const { placement, position } = this.props;
+
+    switch (position) {
+    case 'top':
+      return { left: placement };
```
tomru: NABD, Not sure what others think, though for matching multiple cases I like
```js
switch (position) {
    case 'top':
    case 'bottom':
      return { left: placement };
...
```

packages/axiom-components/src/Tip/Tip.js
MariamBashir: Should be required.

packages/axiom-components/src/Tip/Tip.js
MariamBashir: Oh thats my bad! It should be like that.

packages/axiom-components/src/Context/Context.js
```diff
@@ -35,17 +39,19 @@ export default class Context extends Component {
...
-          <span className={ arrowClasses } ref={ arrowRef } />
-        ) }
-
-        <div className="ax-context__content" style={ { maxHeight } }>
-          { children }
-        </div>
+        { arrowRef ?
+          <Tip arrowRef={ arrowRef } color={ color } placement={ arrowPlacement } position={ position }>
+            <div style={ { maxHeight } }>
+              { children }
```
HHogg: I think this could be even simpler, if the Tip component is just that a Tip. With it taking content, it's doubling up behaviour on how content is rendered with it, for example here the way Tip renders content is being influenced to how Context does. 

Something like ... 

```jsx
<Tip color="red" direction="up" position="14rem" />
```

Would then allow it to be used in a much more flexible way.

packages/axiom-components/src/Tip/Tip.js
MariamBashir: Ill test with ```%``` and see the results. I saw all other unit were in ```rem``` so i thought ill just stay consistent with it.  It would require us to handle some edge cases though.

packages/axiom-components/src/Context/Context.js
MariamBashir: Yeah that is much cleaner. Ill update that.

packages/axiom-components/src/Tip/Tip.css
```diff
@@ -0,0 +1,68 @@
...
+
+.ax-tip--bottom .ax-tip__arrow {
+  transform: translateY(calc(var(--cmp-tip-size) * -1)) rotate(45deg);
+}
+
+.ax-tip--left .ax-tip__arrow {
+  transform: translateX(var(--cmp-tip-size)) rotate(45deg);
+}
+
+.ax-tip-content {
```
HHogg: I guess this CSS isn't needed anymore?

packages/axiom-components/src/Tip/Tip.js
```diff
@@ -0,0 +1,48 @@
...
+    const { arrowRef, color, direction, ...rest } = this.props;
+
+    const classes = classnames(`ax-tip--${direction}`, {
+      [`ax-tip--${color}`]: color,
+    });
+    const arrowClasses = classnames('ax-tip__arrow');
+
+    return (
+      <Base className={ classes } { ...rest }>
+        <span className={ arrowClasses } ref={ arrowRef } style={ this.getArrowStyles() } />
```
HHogg: Can these elements be merged into a single element now? 🤔

packages/axiom-components/src/Tip/Tip.js
```diff
@@ -0,0 +1,48 @@
...
+export default class Tip extends Component {
+  static propTypes = {
+    arrowRef: PropTypes.func.isRequired,
+    color: PropTypes.oneOf(['success', 'warning', 'error', 'info', 'shade-1', 'shade-2', 'shade-3', 'shade-4']),
+    direction: PropTypes.oneOf(['top', 'bottom', 'left', 'right']),
+    position: PropTypes.string,
+  };
+
+  static defaultProps = {
+    placement: '10%',
```
HHogg: I think this is a case where it's fine not to have a default. It can't really be used without some sort of positioning, and with the case of it being used with the Position components, it's applied by the Popper library through the ref :)

packages/axiom-components/src/Context/Context.js
```diff
@@ -2,27 +2,31 @@ import PropTypes from 'prop-types';
 import React, { Component } from 'react';
 import classnames from 'classnames';
 import Base from '../Base/Base';
+import Tip from '../Tip/Tip';
 import './Context.css';
 
 /* eslint-disable react/no-find-dom-node */
 export default class Context extends Component {
   static propTypes = {
+    arrowPlacement: PropTypes.string,
```
HHogg: This isn't needed as it's applied by the Popper library through the ref

packages/axiom-components/src/Context/Context.js
```diff
@@ -2,27 +2,31 @@ import PropTypes from 'prop-types';
...
     arrowRef: PropTypes.func,
     children: PropTypes.node,
-    color: PropTypes.oneOf(['success', 'warning', 'error', 'info']),
+    color: PropTypes.oneOf(['success', 'warning', 'error', 'info', 'shade-1', 'shade-2', 'shade-3', 'shade-4']),
```
HHogg: > `'shade-1', 'shade-2', 'shade-3', 'shade-4'` 

Should these only be defined on the Tip? Context currently doesn't support these as background colours right?

packages/axiom-components/src/Tip/Tip.js
MariamBashir: I tried but couldn't make it work.

packages/axiom-components/src/Tip/Tip.css
MariamBashir: Its not. Will remove.

lpoulter:
Does this need a page on the components documentation?

MariamBashir:
I have updated the PR again with yet another commit. I tested this over the weekend (because obviously i have no life :P)  and realised that i cant make it work without making sure that the content that follow the ```Tip``` are always relatively positioned and use ``` -webkit-mask-image: -webkit-radial-gradient(white, black)``` and   ```background-color: var(--color-theme-background);``` to hide half of the arrow that is behind them. @HHogg , its unfortunate that i couldn't make it work like you had suggested because i myself found that much cleaner.  
If the content is not using the above two css properties this is how it looks.

screen shot 2018-07-23 at 09 24 39

I did however make sure that no other component thats use context are affected by it. 

I am also adding their screenshots 

DropDown:

screen shot 2018-07-23 at 09 18 33

Positioning:

screen shot 2018-07-23 at 09 17 26

Alert Dropdown
screen shot 2018-07-23 at 09 17 14


I am open to any work arounds and others way to make it work. I apologize for the unprecedented back and forth on this PR. 

@lpoulter 
I have added the documentation as you mentioned. Earlier, this looked like a localised change but after taking the```content``` as ```children``` to the ```Tip``` it makes sense to add the documentation.

I used it with the component picker and added a configurations to handle the ```shadow```. The tip has ```shadow``` by default but because we dont need that in the component picker, we can set it to false and our use case is also handled.

lpoulter:
He @MariamBashir thanks for the updates. Will take a look now one thing i've noticed is the direction of the arrow is opposite to what I would have expected. i.e when `direction = left` the arrow is to the right of the context.

MariamBashir:
@lpoulter yes it is, and thats because it is used that way in many different places by the context component. I was a little bit worried if i should touch all those files. I also thought about may be mapping it so if a user sets a ```left``` i can map it to ```right``` and show accordingly. What do you think about that?

lpoulter:
I see. It looks with the Context component the `position` prop refers to the relative position of the context to the context target. It looks like the Tip `direction` prop refers to the position of the tip context to whatever it's pointing at. 
I'd change the description of the `direction` prop to better reflect this.

MariamBashir:
@lpoulter Yes, that the use cases are different. ill update the description. 👍

PR #628, opened 2018-07-20
chore(dates): replace date-fns with luxon

tomru:
In preparation for https://github.com/BrandwatchLtd/axiom/pull/614 replace `date-fns` with [luxon](https://github.com/moment/luxon/blob/master/changelog.md).

luxon supports ISO 8601 durations and intervals which is needed for #614


packages/axiom-components/src/DatePicker/utils.js
```diff
@@ -1,30 +1,18 @@
...
-  startOfMonth,
-} from 'date-fns';
+import { DateTime } from 'luxon';
+
+const { fromJSDate } = DateTime;
 
 export const buildMonthGrid = (date) => {
-  const startMonth = setHours(startOfMonth(date), 12);
-  const startDay = getDay(startMonth) === 0 ? 6 : getDay(startMonth) - 1;
+  const startMonth = fromJSDate(date).startOf('month').set({ hours: 12 });
```
HHogg: cc @stefanpearson as he queried the setting hours to 12 before. IIRC it was the case of just ignoring hours all together and passing around timezone-_less_ date strings.... but not blocking this if it's a pain to sort out now.

tomru: Discussed with @stefanpearson - this seems odd to both of us. We don't see a reason to change set the hours to 12. Removing this...

PR #630, opened 2018-07-24
Remove default color from tip

MariamBashir:
- This PR removes the default color from the tip. This introduced a bug into all components that were using Context. 

Here's the deploy branch. : https://5b572ad2e39e7c1d2253a11c--affectionate-mcclintock-42eb2e.netlify.com/


HHogg:
Thanks for fixing this @MariamBashir ... would you mind labelling the commit as a `fix` so it gets the correct version bump and also rebasing against master... not sure what those merge commits will do to the semantic version.

MariamBashir:
@HHogg I updated the commit message and i pulled from master just before creating this branch.Not sure why i would need to do that again ?

PR #632, opened 2018-07-27
fix(InputWrapper) Using 'node' prop-type for label prop in InputWrapper.

SeaOfTea:


lpoulter:
Can you add some description, what the idea here? What other types of label would you be setting?

SeaOfTea:
Hey @lpoulter , yeah sorry. I did have some description in here, but realised it was a link to invision, and as axiom is a public repo I wasn't 100% sure it was a good idea to have it in the description. Was in a rush so just removed the description altogether. The reasoning for this PR is that a design requires a clickable link that will change some state (in my case hiding a textarea).

screen shot 2018-07-27 at 16 06 31

I hadn't thought about possibly using oneOf, maybe `oneOf([PropTypes.object, PropTypes.string])` ?

pmsorhaindo:
I think `oneOf([PropTypes.string, PropTypes.node])` is what's needed here.

packages/axiom-components/src/Form/InputWrapper.js
```diff
@@ -15,7 +15,7 @@ export default class InputWrapper extends Component {
   static propTypes = {
     children: PropTypes.node,
     isTarget: PropTypes.bool,
-    label: PropTypes.string,
+    label: PropTypes.any,
```
HHogg: `PropTypes.node` as it would cover strings and React elements.

SeaOfTea:
Have updated this PR to use `node` instead of `any`.

PR #634, opened 2018-07-31
chore(CI): fail build when publishing packages fails

tomru:
@HHogg as discussed, this should fail the build when publishing packages fails for some reason...


tomru:
Before we merge this the NPM token should be fixed though...

tomru:
closing for now until the work on publishing is done

PR #635, opened 2018-07-31
feat(BarChart): highlight row when mouse is anywhere inside the row

tomru:
When you hovered over the benchmark line other rows have lost their fade state.

The root cause for this is that the selected row is set when you hover over a bar and the benchmark line is not part of the bar.

This PR fixes this by introducing a separate state, when you hover over the over a row, by just setting `selectedIndex`. When you hover over over a bar the `selectedColor` then set.

This made it possible to keep the other rows faded, when you hovered over the benchmark line.

### Before
highlight-row-bug

### After
highlight-row-fix

(Sorry about the grey background artifacts, seems licecap is bugged out by them when recording the gifs)

~~[netlify link](https://5b603b22b13fb10a278ecf2b--youthful-albattani-41ebb2.netlify.com/docs/packages/axiom-charts/bar-chart)~~

[netlify link](https://5b62f443dd28ef3382c20440--youthful-albattani-41ebb2.netlify.com/docs/packages/axiom-charts/bar-chart)

### DIFF
--- a/packages/axiom-charts/src/ChartTable/ChartTable.css
+++ b/packages/axiom-charts/src/ChartTable/ChartTable.css
@@ -66,6 +66,7 @@
   top: 0;
   right: 0;
   bottom: 0;
+  pointer-events: none;
 }
 
 .ax-chart-table__grid-line {
--- a/packages/axiom-charts/src/BarChart/BarChart.js
+++ b/packages/axiom-charts/src/BarChart/BarChart.js
@@ -138,18 +138,38 @@ export default class BarChart extends Component {
     });
   }
 
-  handleMouseEnter(selectedIndex, selectedColor) {
+  handleMouseEnterRow(selectedIndex) {
     this.setState({
-      selectedColor,
       selectedIndex,
     });
   }
 
-  handleMouseLeave() {
+  handleMouseLeaveRow() {
     this.setState(({ isDropdownOpen }) => isDropdownOpen ? {} : ({
-      selectedColor: null,
       selectedIndex: null,
     }));
+  }
+
+  handleMouseEnterBar(selectedIndex, selectedColor) {
+    this.setState({
+      selectedIndex,
+      selectedColor,
+    });
+  }
+
+  handleMouseLeaveBar() {
+    this.setState(({ isDropdownOpen }) => isDropdownOpen ? {} : ({
+      selectedColor: null,
+    }));
+  }
+
+  handleToggleRowVisibility(data, index, hidden) {
+    this.setState({
+      selectedIndex: hidden ? index : null,
+    });
+
+    const { onToggleRowVisibility } = this.props;
+    onToggleRowVisibility && onToggleRowVisibility(data, index);
   }
 
   render() {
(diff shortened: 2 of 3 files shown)


packages/axiom-charts/src/ChartTable/ChartTable.css
```diff
@@ -62,6 +62,7 @@
 }
 
 .ax-chart-table__grid {
+  pointer-events: none;
```
tomru: There was a weird thing happening when you moved the mouse from one row to the next one. The `mousleave` event was triggered before the mouse actually left the row and the `mouseenter` was fired a bit after you enter the row. This cause quite some flickering if you move the mouse fast over the chart rows.

bad

I've poked around quite a bit in the DOM to find what is causing that. Weirdly it seems to be the table grid laying behind the rows that's causing it.

I have no solid explanation for why it happens, but it can be observer in Chrome and Firefox. Setting `pointer-events: none` on the table grid solves the issue:

good

tomru:
Hey @mozzling could you look at this one - @lilybrandwatch reviewed it already. The netlify link shows the benchmark component use case.

Let me know if you would like to have a look at the multi-bar case as well.

Thanks

mozzling:
Hey @tomru - we initially decided against highlighting the whole row (unless when interacting directly with the data point) as we were worried that it would look quite flickery when scrolling down the page or just moving the cursor about. It's so subtle that perhaps it's time for a rethink - I like this so let's go ahead. Just one thing to note - when you hide a row with the eye icon - I think the row's hover bar should also be disabled - otherwise we're highlighting a row with nothing in it.

tomru:
Hiya @mozzling,
thanks for having a look. Very good point about removing the highlight when you hide the row. That's now addressed [here](https://github.com/BrandwatchLtd/axiom/pull/635/commits/b018d744eda53ed47af9e4b8d4cbca512d5e7580#diff-279ce1861c91c7de1797523377744c37R226) and can [be reviewed](https://5b695e59c965923101556e92--youthful-albattani-41ebb2.netlify.com/docs/packages/axiom-charts/bar-chart)

mozzling:
Hey @tomru - Apologies, I noticed one more thing. The hidden row still has some interaction on hover - the label turns bold and the benchmark lines on non-hidden rows also dim. Both of these interactions should be disabled.
aug-07-2018 10-57-43

tomru:
Hey @mozzling,

apologies from my side not catching that - fixed on [here](https://5b697b743813f0030321bafa--youthful-albattani-41ebb2.netlify.com/docs/packages/axiom-charts/bar-chart)

mozzling:
Hey @tomru - I'm still seeing the benchmark bars fading on non-hidden rows when hovering over a hidden row. Can you disable this also?
aug-07-2018 15-10-55

tomru:
stupid me  - sorry for wasting your time @mozzling now both are fixed on [here](https://5b6a9f1f3813f04cba25e8ae--youthful-albattani-41ebb2.netlify.com/docs/packages/axiom-charts/bar-chart/)

@HHogg `selectedIndex` will now never be set for a hidden row, hence we don't need to add the `!hidden` condition in multiple places

mozzling:
Hi @tomru - no problem at all. All good to go.

tomru:
Thanks @lpoulter - addressed now

PR #636, opened 2018-07-31
fix(Reveal): Add check for inner if component is removed before callb…

MariamBashir:
In this PR, a check has been added for `inner` which is used to set the height of the `Reveal` component.  This problem was identified during the keyboard navigation in component picker when the reveal component started throwing this exception. 

screen shot 2018-07-30 at 14 37 02

It works perfectly if we use mouse click. 
If this solution doesn't sound good, other suggestions are appreciated. 

Deployed Branch: https://5b606c76b3127438e1f38461--affectionate-mcclintock-42eb2e.netlify.com/

### DIFF
--- a/packages/axiom-components/src/Reveal/Reveal.js
+++ b/packages/axiom-components/src/Reveal/Reveal.js
@@ -48,11 +48,15 @@ export default class Reveal extends Component {
     }
   }
 
+  componentWillUnmount() {
+    window.cancelAnimationFrame( this._frameId );
+  }
+
   conceal() {
     this.setState({ overflow: 'hidden' });
-    window.requestAnimationFrame(() => {
+    this._frameId = window.requestAnimationFrame(() => {
       this.setState({ height: this.inner.offsetHeight });
-      window.requestAnimationFrame(() => {
+      this._frameId = window.requestAnimationFrame(() => {
         this.setState({ height: 0, opacity: 0 });
       });
     });


tomru:
I'm not really familiar with `requestAnimationFrame`, but if this happens when the component is removed before the animationFrame is executed, maybe we should rather store the request id returned from `window.requestAnimationFrame` and cancel that via `window.window.cancelAnimationFrame(requestID)` in a `componentWillUnmount` ?

MariamBashir:
@tomru  Not sure about it myself, i will check if these two problems are actually related. However, what you are suggesting should be done anyway.

MariamBashir:
@tomru Thank You for the suggestion, i added the change. It fixes the problem without the need for checking for `inner` in an `if` condition.

packages/axiom-components/src/Reveal/Reveal.js
```diff
@@ -48,9 +48,13 @@ export default class Reveal extends Component {
...
+  componentWillUnmount() {
+    window.cancelAnimationFrame( this._frameId );
+  }
+
   conceal() {
     this.setState({ overflow: 'hidden' });
-    window.requestAnimationFrame(() => {
+    this._frameId = window.requestAnimationFrame(() => {
       this.setState({ height: this.inner.offsetHeight });
       window.requestAnimationFrame(() => {
```
tomru: I'd also overwrite `this._frameId` here again, as otherwise the second call might still fail the same way in another case.

MariamBashir: You mean for the nested `window.requestAnimationFrame`?

tomru: yep, exactly - or is there something I've overlooked ?

MariamBashir: Not at all. It makes sense.

MariamBashir:
@tomru Please have another look. Your suggestions have been quite handy today. Thanks.

PR #637, opened 2018-08-01
fix(ColorPicker): cascades size property to main ColorPickerOption

samjmiele:


packages/axiom-components/src/ColorPicker/ColorPicker.js
```diff
@@ -105,10 +105,13 @@ export default class ColorPicker extends Component {
     ]),
     /** Width of the color selection container */
     width: PropTypes.string,
+    /** Diameter of the colour picker button */
+    size : PropTypes.string,
   };
 
   static defaultProps = {
     disabledOptions: [],
+    size: ".75rem",
```
samjmiele: TBF I should probably rely on the ColorPickerOptions default size. But I wasn't sure if passing down undefined would mess with everything

packages/axiom-components/src/ColorPicker/ColorPicker.js
```diff
@@ -143,6 +146,7 @@ export default class ColorPicker extends Component {
           <ColorPickerOption
               color={ selected }
               data-ax-at={ atIds.ColorPicker.dropdown }
+              size = { size }
```
thatguynamedandy: This should probably fail linting, it seemingly doesn't but could be nice to bring it inline `size={ size }`

thatguynamedandy:
Just  a few lint failures to fix @samjmiele, aside from that all good by me.

PR #638, opened 2018-08-03
chore(npmAuth): fix npm config call

tomru:
This is a followup on https://github.com/BrandwatchLtd/axiom/pull/624

The line to set npm config variable was faulty.
- It was missing the config key `registry`
- npm nowadays will throw an error when the protocol is missing

Hope this fixes the publishing 😓

### DIFF
--- a/scripts/publish_packages.sh
+++ b/scripts/publish_packages.sh
@@ -24,7 +24,7 @@ git config --global user.name "Brandwatch (via TravisCI)"
 # travis sets origin to https. Let set up a second remote for ssh
 git remote add upstream git@github.com:BrandwatchLtd/axiom.git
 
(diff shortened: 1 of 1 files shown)


tomru:
> Has there been a change to NPM because I'm pretty sure we've had successful publishing with this config before 🤔 ...

I'm not sure what changed and how it was working before.
 
I've gone back to npm@3.10.8 and set the config via `npm config set "//registry.npmjs.org/:_authToken=\${NPM_API_KEY}"` and verified with `npm i -dd` that npm does not use the specified registry.

Though with `npm config set registry "https://registry.npmjs.org/:_authToken=\${NPM_API_KEY}"`  that version and newer ones do use the specified registry.

PR #647, opened 2018-08-13
AX-16 tooltip sizes

TomWija:
https://5b72963fc6aed67cf32b61df--axiomwija.netlify.com/docs/packages/axiom-components/tooltip/

### DIFF
--- a/packages/axiom-components/src/Slider/Slider.js
+++ b/packages/axiom-components/src/Slider/Slider.js
@@ -163,7 +163,7 @@ export default class Slider extends Component {
 
           <PositionSource>
             <TooltipContext theme="night" width="auto">
-              <TooltipContent textSize="small" textStrong>
+              <TooltipContent size="tiny" textStrong>
                 { valueFormatter(value) }
               </TooltipContent>
             </TooltipContext>
--- a/packages/axiom-components/src/Tooltip/Tooltip.js
+++ b/packages/axiom-components/src/Tooltip/Tooltip.js
@@ -75,7 +75,7 @@ export default class Tooltip extends Component {
     const { isVisible } = this.state;
 
     return (
-      <Position { ...rest } isVisible={ isVisible } position={ position } showArrow>
+      <Position { ...rest } isVisible={ isVisible } position={ position } showArrow >
         <PositionTarget delay={ delay } onClick={ onClick }>
           { findComponent(children, TooltipTargetRef) }
         </PositionTarget>
--- a/packages/axiom-components/src/Tooltip/TooltipContent.js
+++ b/packages/axiom-components/src/Tooltip/TooltipContent.js
@@ -1,10 +1,21 @@
+import PropTypes from 'prop-types';
 import React, { Component } from 'react';
 import ContextContent from '../Context/ContextContent';
 
 export default class TooltipContent extends Component {
+  static propTypes = {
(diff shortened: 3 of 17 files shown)


packages/axiom-components/src/Tooltip/TooltipContent.js
```diff
@@ -1,10 +1,21 @@
+import PropTypes from 'prop-types';
 import React, { Component } from 'react';
 import ContextContent from '../Context/ContextContent';
 
 export default class TooltipContent extends Component {
+  static propTypes = {
+    padding: PropTypes.oneOf(['small', 'medium', 'large']),
```
HHogg: Even though the lower level Context component uses padding, it might make more sense to use `size` here as it does more than just affect padding (text size). 

Also I think there's only two variations of sizes for tooltip.

packages/axiom-components/src/Slider/Slider.js
```diff
@@ -163,7 +164,7 @@ export default class Slider extends Component {
 
           <PositionSource>
             <TooltipContext theme="night" width="auto">
-              <TooltipContent textSize="small" textStrong>
+              <TooltipContent size="small" textSize="small" textStrong>
```
HHogg: No biggy but, I think you could remove the `textSize="small"` from here

packages/axiom-components/src/Slider/Slider.js
```diff
@@ -151,6 +151,7 @@ export default class Slider extends Component {
         </div>
 
         <Position
+            arrowSize="small"
```
HHogg: This isn't needed any more right as this prop doesn't exist and is enforced inside Context?

site/components/Documentation/Resources/Components/Tooltip.js
```diff
@@ -21,21 +23,42 @@ export default class Documentation extends Component {
...
 
-            <TooltipSource theme="night" width="auto">
-              <TooltipContext>
-                <TooltipContent>Tooltip content</TooltipContent>
-              </TooltipContext>
-            </TooltipSource>
-          </Tooltip>
+                <TooltipSource theme="night" width="auto">
+                  <TooltipContext>
+                    <TooltipContent size="small">50%</TooltipContent>
```
lpoulter: I think 50% is slightly confusing here. Why not use size - `small`

packages/axiom-components/src/Tooltip/Tooltip.js
```diff
@@ -75,7 +75,7 @@ export default class Tooltip extends Component {
     const { isVisible } = this.state;
 
     return (
-      <Position { ...rest } isVisible={ isVisible } position={ position } showArrow>
+      <Position arrowSize="small" { ...rest } isVisible={ isVisible } position={ position } showArrow >
```
lpoulter: ` arrowSize="small"` should this be here? I can't see anywhere this prob is used.

site/components/Documentation/Resources/Components/Tooltip.js
TomWija: Yeah I was thinking that. Was trying to convey the usage of the tiny tooltips which are to show small indiviudal values (like the slider tooltip), guess this is where dos and don'ts will come in handy to convey reasoning

PR #648, opened 2018-08-14
AX-26: chore(Dropdown): Update to use new context API

lpoulter:
The old context API will likely be deprecated in the next Major react version. We will want to convert over components to the new API. I've started with the dropdown as it's fairly complicated and should give us a good pattern for how to do the rest.

As the new context api uses the render props pattern we are a little restricted in how we can use it. If we want to reference the context outside of the render method we can pass the context in as an argument i.e.

```
 <DropdownReactContext.Consumer>
	{
      context => <btn onClick={() => this.handleClick(context)}
    }
</DropdownReactContext.Consumer>
```

If we want to access in a lifecycle method we have to pass as a prop, which is what I've done for the `DropdownContext` component.

[Netlify](https://nostalgic-curie-a20a02.netlify.com/docs/packages/axiom-components/dropdown)


packages/axiom-components/src/Dropdown/Dropdown.js
```diff
@@ -101,19 +102,24 @@ export default class Dropdown extends Component {
...
-          offset={ offset }
-          onMaskClick={ withMask ? () => this.close() : null }
-          position={ position }
-          ref={ (el) => this.el = ReactDOM.findDOMNode(el) }>
-        <PositionTarget>
-          { React.cloneElement(findComponent(children, DropdownTargetRef)) }
-        </PositionTarget>
-        <PositionSource>{ findComponent(children, DropdownSourceRef) }</PositionSource>
-      </Position>
+      <DropdownReactContext.Provider value={ this.getChildContext() }>
```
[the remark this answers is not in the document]
tomru: Also maybe rename `getChildContext()` to `getDropdownContext()` ?

packages/axiom-components/src/Dropdown/DropdownContext.js
```diff
@@ -16,8 +16,10 @@ export default class DropdownContext extends Component {
     arrowRef: PropTypes.func,
     /** Content to be inserted in the contextual area */
     children: PropTypes.node,
+    closeDropdown: PropTypes.func.isRequired,
```
tomru: Should we define the context proptypes in one place, so they can be shared across components?

HHogg:
I'm still not sure whether we should be upgrading to the new Context API or removing context completely. My understanding of it is provides a way to communicate between components through _an unknown number of levels_. 

Reacts new context components use the functions as children props, and I'm wondering if over all the context uses we have at the moment they can either be removed in favour for just using css (like with the Menu/MenuItem) and for cases like the Dropdown where we know that the number of levels is 1 where we skip the context and go straight to function as children? 

What's your thoughts @lpoulter?

lpoulter:
@HHogg You raise a good point. I think you are right in that the Context API is best used to avoid prop drilling and the more levels deep this drilling is the greater the need for a state management system beyond props. 

I considered just using just props but felt it might add more noise to the code. I think ultimately it is a style choice. The down sides of prop drilling is increased noise in the code and having to use Omit in a few more places, I'm also unsure if the `findComponent` approach might cause complications. However using the context API may introduce more complexity and unnecessary abstraction. 

I'll add a new commit which removes the context so that we can compare.

packages/axiom-components/src/Context/Context.js
```diff
@@ -36,7 +37,7 @@ export default class Context extends Component {
       [`ax-context--${color}`]: color,
     });
     const content = () => <div className="ax-context__content" style={ { maxHeight } }>
-      { children }
+      { cloneElement(children, { closeDropdown }) }
```
lpoulter: This will fail if `children` is not a react element, i'm not sure if that is possible.

packages/axiom-components/src/Dropdown/DropdownMenu.js
```diff
@@ -1,19 +1,24 @@
 import PropTypes from 'prop-types';
-import React, { Component } from 'react';
+import React, { Component, cloneElement } from 'react';
 import ContextMenu from '../Context/ContextMenu';
 
 export default class DropdownMenu extends Component {
   static propTypes = {
     /** DropdownMenuItems */
     children: PropTypes.node.isRequired,
+    closeDropdown: PropTypes.node.func,
```
lpoulter: There are some issues here with prop types I have not been able to resolve. This issue stems for the fact `cloneElement` happens after the component is already rendered once, without the `closeDropdown` prop. This was resolved by added a default previously but seems to not help here. Perhaps because it's rendered in the demo rather than a parent component.

packages/axiom-components/src/Dropdown/DropdownMenu.js
```diff
@@ -1,19 +1,24 @@
...
   };
 
   render() {
-    const { children, ...rest } = this.props;
+    const { children, closeDropdown, ...rest } = this.props;
 
     return (
       <ContextMenu { ...rest }>
-        { children }
+        { children.map( child => cloneElement(child, { closeDropdown })) }
```
lpoulter: Each `DropdownMenuItem` needs `closeDropdown` injected into it.

packages/axiom-components/src/Context/Context.js
lpoulter: Also this is kinda gross. Not sure if there is a better way.

lpoulter:
@HHogg I've put up a commit which removes context from dropdown altogether. It looks good for the target. However passing `closeDropdown` to the more nested dropdownMenuItem got pretty ugly. There are a couple of warnings I've not been able to shake. 
Let me know what you think.

HHogg:
Oh right yeah I agree the clone element approach is kinda gross, and probably has some sort of impact on performance. I was more suggesting the function as children approach like ... 


```jsx
class Lorem extends Component {
  station propTypes = {
    children: PropTypes.func.isRequired,
  };

  render() {
    return (
      <div>{ this.props.children({ onCloseDropdown }) }</div>
    );
  }
}
```

... which would be used like... 

```jsx
<Lorem>
  { ({ onCloseDropdown }) => (
    <button onClick={ onCloseDropdown }>Close</button> 
  ) }
</Lorem>
```

It's the same pattern that the new Context components use, but without the context. 

We might also avoid breaking changes by providing both children as function or nodes. 

```jsx
class Lorem extends Component {
  station propTypes = {
    children: PropTypes.oneOfType([
      PropTypes.func,
      PropTypes.node,
    ]).isRequired,
  };

  render() {
    return (
      <div>
        { typeof this.props.children === 'function' 
          ? this.props.children({ onCloseDropdown }) }
          : node 
        }
      </div>
    );
  }
}
```

I've never done the multi-approach before but I guess it would work 🤔

PR #650, opened 2018-08-16
Fix(card) removed unused color prop

samjmiele:


packages/axiom-components/src/Card/CardContent.js
```diff
@@ -7,12 +7,12 @@ export default class CardContent extends Component {
   static propTypes = {
     /** Content to be inserted inside the Card */
     children: PropTypes.node.isRequired,
-    /** Color variation */
-    color: PropTypes.oneOf(['default', 'dark', 'darker']),
```
HHogg: Normally a removal of a prop would be classed as a breaking change, but the functionality was removed aggeees ago and the prop got left behind.

PR #652, opened 2018-08-27
precise mode in line chart x axis

tomru:
## Background

`LineChart` uses `ChartLayoutLabels` to position the labels on the x- and y-axis. 

On the y-axis it supports the [`precise mode`](https://github.com/BrandwatchLtd/axiom/blob/master/packages/axiom-charts/src/ChartLayout/ChartLayoutLabels.js#L33), i.e. you specify the exact position with 
```js
[
  {
    label: <text>,
    value: <number>,
  },
  ...
]
```

The value is then converted, based on `upper` and `lower`, to the range of `0-100%`.

For the x-axis the `precise mode` is not supported. Instead the passed labels are [positioned with flexbox](https://github.com/BrandwatchLtd/axiom/blob/8c021a1f5aee0443756efc3e9fd9a6b39e76a945/packages/axiom-charts/src/ChartLayout/ChartLayout.css#L30-L31). Though that seems wrong as
* it enforces the consumer to use labels, that match the positioning of flexbox
* positioning with `justify-content: space-around` never aligns with the chart, even when you pass `n-2` labels for `n` data points (the line is always drawn from position 0% to 100%)

image

Perhaps I'm lacking the use-case for the positioning with flex-box ?

Nevertheless, I'd like to enable the `precise mode` in this PR on the x-axis, as then a consumer is then free to position the labels, instead of matching the labels to the internal positioning by axiom.

It's not yet defined what the domain of the x-axis could be. It does not have an upper or lower limit, nor is it bound to be just numeric - time is a quite common use-case (see benchmark component).

The line does not really care about the domain of the x-axis, as it simply spreads all data points equally, beginning with position `0%` up to `100%` for the last data point.

Instead of introducing another `upper` and `lower` for the x-axis I'm proposing to position in the domain used by the line already `0-100%`. 

## Changes in this PR

* labels of the y-axis were positioned from the top, hence with `${100-normalizedValue}`. I switched that, so both axis can be positioned relative the the origin (lower left corner, being 0%,0%)
* `lower` and `upper` are not passed to `ChartLayoutLabels` for the x-axis. Define them as `0%`, `100%` in the default props.
* update the example with aligned x-axis labels

## not tackled inside this PR

* non-precise mode, as I'm unsure if I missed a use-case and will not remove it until discussed

## Visuals
image

[Netlify](https://5b84114b67610c267eeed09a--youthful-albattani-41ebb2.netlify.com/docs/packages/axiom-charts/line-chart)

The example has a data point for each day, from Dec 1st to Dec 27th, and labels for each second day.

### DIFF
--- a/packages/axiom-charts/src/ChartLayout/ChartLayout.css
+++ b/packages/axiom-charts/src/ChartLayout/ChartLayout.css
@@ -57,9 +57,14 @@
   position: relative;
 }
 
-.ax-chart-layout__label {
+.ax-chart-layout__y-label {
   position: absolute;
-  transform: translateY(-50%);
+  transform: translateY(50%);
+}
+
+.ax-chart-layout__x-label {
+  position: absolute;
+  transform: translateX(-50%);
 }
 
 .ax-chart-layout__x-labels .ax-chart-layout__label { top: 0; }
--- a/packages/axiom-charts/src/ChartLayout/ChartLayoutLabels.js
+++ b/packages/axiom-charts/src/ChartLayout/ChartLayoutLabels.js
@@ -18,6 +18,11 @@ export default class ChartLayoutLabels extends Component {
     upper: PropTypes.number,
   };
 
+  static defaultProps = {
+    lower: 0,
+    upper: 100,
+  };
+
   render() {
     const { axis, labels, lower, upper, ...rest } = this.props;
     const side = axis === 'y' ? 'bottom' : 'left';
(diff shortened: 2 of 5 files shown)


tomru:
@lilybrandwatch and @mozzling in the example I've used also a label for the first and last data point. The last one is always cut off and I'm unsure if we should not in the _docs_
* do not position a label to go over the right edge of the chart
* or rather add more spacing after the last data point inside the chart

tomru:
Thanks @HHogg - agreed, will align the prop types then 👍

tomru:
As it is now up to the consumer to choose the label, I think this can go in as it does not change anything design related.

PR #653, opened 2018-08-28
fix(Dropdown): Implement native style hover and focus

roburidge:
Related ticket: [AX-53](https://office.brandwatch.com/jira/browse/AX-53)
Fixes: [TOP-853](https://office.brandwatch.com/jira/browse/TOP-853) & [STS-160](https://office.brandwatch.com/jira/browse/STS-160)

##### Menus in a Dropdown now have the same hover and focus styles as a native select menus.

[Netlify](https://5b8eb280dd28ef7b7f4716af--priceless-benz-91b798.netlify.com/docs/packages/axiom-components/dropdown)

Previously hover and focus styles were separate and you could get into a state where both were bing applied on two menu items. Now the blue highlight can only be applied to a single item even when switching between mouse and keyboard menu navigation.

old:
e8nl9cv6h2

new:
cnuyiwdzzc

### DIFF
--- a/packages/axiom-components/src/Dropdown/DropdownContext.js
+++ b/packages/axiom-components/src/Dropdown/DropdownContext.js
@@ -5,6 +5,10 @@ import omit from 'lodash.omit';
 import Context from '../Context/Context';
 import { contextMenuItemSelector } from '../Context/ContextMenuItem';
 
+if (typeof window !== 'undefined') {
+  require('element-closest');
+}
+
 const isFocusableMenuItem = (element) =>
   element && element.hasAttribute(contextMenuItemSelector) &&
     !element.disabled;
@@ -41,6 +45,7 @@ export default class DropdownContext extends Component {
     super(props);
     this.handleKeyDown = this.handleKeyDown.bind(this);
     this.handleClick = this.handleClick.bind(this);
+    this.handleMouseMove = this.handleMouseMove.bind(this);
   }
 
   componentWillMount() {
@@ -50,10 +55,12 @@ export default class DropdownContext extends Component {
   componentWillUnmount() {
     document.removeEventListener('keydown', this.handleKeyDown);
     document.removeEventListener('mousedown', this.handleClick);
+    document.removeEventListener('mousemove', this.handleMouseMove);
   }
 
   componentDidMount() {
     document.addEventListener('mousedown', this.handleClick);
+    document.addEventListener('mousemove', this.handleMouseMove);
   }
 
   handleClick(event) {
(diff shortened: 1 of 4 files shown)


lpoulter:
Looks like a simple and effective solution. Works nicely, good stuff.

packages/axiom-components/src/Dropdown/DropdownContext.js
```diff
@@ -71,6 +74,18 @@ export default class DropdownContext extends Component {
     return [...this.el.querySelectorAll(`[${contextMenuItemSelector}]:not(:disabled)`)];
   }
 
+  focusMenuItem(element) {
+    if (isFocusableMenuItem(element)) {
+      element.focus();
+    } else if (isFocusableMenuItem(element.parentNode)) {
```
HHogg: Does this work for several nested elements inside a menu item or does this need to walk up the DOM tree until it finds a `isFocusableMenuItem`?

roburidge: The nested menu groups still work fine as we're picking up the event on the buttons with "data-ax-context-menu-item" but now you mention it I wasn't totally happy with this solution.

The events we care about are either triggered on the button or any of it's children. This conditional catches events from the button and from one of it's immediate children however it doesn't work when hovered over the check boxes as they have another level of nesting. I'll change the implementation to be less dependent on the DOM structure.

HHogg:
@roburidge Looks like a nice polyfill. 

You'll need to conditionally require it for SSR/static generators. Something like ... 

```js
if (typeof window !== 'undefined') {
  require('element-closest');
}
```

jdelight:
Hi @HHogg - I (or someone else on the team) will need to pick this up as Rob is away now. Regarding your [comment](https://github.com/BrandwatchLtd/axiom/pull/653#issuecomment-416978708) - I'm not sure I fully understand so would be good to catch up in person perhaps. Or if you could clarify on the PR that would be good to.

I was just looking for examples of this in the repo and didn't see many so was wondering if this is something for the documentation or for the component itself. 

I've also not used SSR much; does it currently break with the addition of this polyfill? and also how are we/others using SSR with axiom (just so we can be aware of this use case for the future). Thanks.

HHogg:
@jdelight  The polyfill works by adding the missing methods (matches, closest) onto [`window.Element.prototype`](https://github.com/jonathantneal/closest/blob/master/element-closest.js#L31). 

`window` is not available during server side rendering or when static generating pages during a build process, because it happens in node, where `window` isn't a thing (unless it is being spoofed with something like JSDom).

Axiom site is generated using a static site generator, which is the reason @roburidge is seeing the `window is not defined` error. 

As this is an interacting thing, and something that isn't required during SSR the way to resolve it is to exclude the polyfill being loaded in those environments. This can be done by changing... 

```js
import 'element-closest';
```

to 

```js
if (typeof window !== 'undefined') {
  require('element-closest');
}
```

jdelight:
@HHogg perfect thanks - I couldn't see where the SSR was used but:

> Axiom site is generated using a static site generator, which is the reason @roburidge is seeing the window is not defined error.

Cleared that up for me 👍

lpoulter:
Could we just adapt the [polyfill](https://developer.mozilla.org/en-US/docs/Web/API/Element/closest#Polyfill) to check for window and add to helpers?

roburidge:
I think I'd prefer to keep the `window` check in this project as the polyfill is only used in IE11 so we can remove it when we no longer need it. I'd rather that than duplicate it into helpers, might try to put the conditional logic for checking window in the webpack config, not sure if it's better but putting the check in with the import looks weird to me 🤷‍♀️

roburidge:
@HHogg hey, I'm back and picking this up again, I've implemented the suggested solution. Could you give this another review? Really want to get this in, thanks 👍

PR #657, opened 2018-08-31
fix(Slider): Fix onSlideEnd being unrecognised

TomWija:


packages/axiom-components/src/Slider/Slider.js
```diff
@@ -140,6 +140,7 @@ export default class Slider extends Component {
           onFocus={ disabled ? null : this.handleFocus }
           onMouseLeave={ () => this.setState({ isMouseOver: false }) }
           onMouseOver={ () => this.setState({ isMouseOver: true }) }
+          onSlideEnd={ onSlideEnd }
```
HHogg: This is going to cause the same propType warning... it needs to be omitted from the rest spread. 

``` js
omit(rest, ['onSlideEnd']);
```

HHogg: Thanks for adding the omit line... but this line still needs removing to stop the proptype warning.

PR #660, opened 2018-09-06
feat(Platform): Allows use of shade-3 and shade-4 in the Platform component

SeaOfTea:


packages/axiom-components/src/Platform/Platform.js
```diff
@@ -12,7 +12,7 @@ export default class Platform extends Component {
     openConsolePosition: PropTypes.oneOf(['left', 'right']),
     openConsoleWidth: PropTypes.string,
     responsive: PropTypes.bool,
-    shade: PropTypes.oneOf(['shade-1', 'shade-2']),
+    shade: PropTypes.oneOf(['shade-1', 'shade-2', 'shade-3']),
```
HHogg: You'll need to add the CSS to go along with this... 

https://github.com/BrandwatchLtd/axiom/blob/master/packages/axiom-components/src/Platform/Platform.css#L117

PR #663, opened 2018-09-11
fix(SparkLine): Make linepoint size configurable

samjmiele:


HHogg:
The commit should probably be a `feat` though.

PR #664, opened 2018-09-12
Fix input container spacing

TomWija:
This fixes an issue that was introduced in 4.11.0 (https://github.com/HHogg/axiom/commit/bc3072c) where because the inputWrapper was moved inside of the TextInput's input container, the bottom-margin of `ax-space--x4` would no longer appear as it would always be the last child.

This adds a `space` property to the TextInput's container to rectify this. 

I've also made it so the documentation showcases for TextInputs and TextAreas now have at least one case of two inputs in a showcase area so that it is easier to notice these issues in development

### DIFF
--- a/packages/axiom-components/src/Form/TextInput.js
+++ b/packages/axiom-components/src/Form/TextInput.js
@@ -69,6 +69,7 @@ export default class TextInput extends Component {
 
   static defaultProps = {
     size: 'medium',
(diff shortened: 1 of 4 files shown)


stefanpearson:
> the bottom-margin of ax-space--x4 would no longer appear as it would always be the last child.

do we still need another rule for last-child?

PR #666, opened 2018-09-13
Support className on `Grid` and `GridCell`

stefanbuck:
Following scenario. 

I want to create a table using Axiom Grid. The table consist of two columns ID and name. The ID column needs to have a background image. Axiom doesn't except the `className` property for `GridCell` which is the reason why I added a `div` as a children of `GridCell`. Not very nice, but it works. 

```js
<Grid>
    <GridCell>
        <Grid>
            <GridCell>
                <div className="pattern-background">2</div>
            </GridCell>
            <GridCell>Smith</GridCell>
        </Grid>
    </GridCell>
    <GridCell>
        <Grid>
            <GridCell>
                <div className="pattern-background">3</div>
            </GridCell>
            <GridCell>Williams</GridCell>
        </Grid>
    </GridCell>
    <GridCell>
        <Grid>
            <GridCell>
                <div className="pattern-background">5</div>
            </GridCell>
            <GridCell>Rodriguez</GridCell>
        </Grid>
    </GridCell>
</Grid>
```

However, now there is this other requirement that the last row should not have a background image. I don't want to do this programmatically and would prefer using css for that and here the problem starts.

I can not rely on Axiom class names because they can change without notice. Therefore I would like to be able to add additional classes for `Grid` and `GridCell` so I can write use Axiom like that

```js
<Grid>
    <GridCell className="table-row">
        <Grid>
            <GridCell className="pattern-background">2</GridCell>
            <GridCell>Smith</GridCell>
        </Grid>
    </GridCell>
    <GridCell className="table-row">
        <Grid>
            <GridCell className="pattern-background">3</GridCell>
            <GridCell>Williams</GridCell>
        </Grid>
    </GridCell>
    <GridCell className="table-row">
        <Grid>
            <GridCell className="pattern-background">5</GridCell>
            <GridCell>Rodriguez</GridCell>
        </Grid>
    </GridCell>
</Grid>
```

```css

.table-row:last-child .pattern-background {
    background: none;
}

```

### DIFF
--- a/packages/axiom-components/src/Grid/Grid.js
+++ b/packages/axiom-components/src/Grid/Grid.js
@@ -8,6 +8,8 @@ export default class Grid extends Component {
   static propTypes = {
     /** <GridCells>. */
     children: PropTypes.node,
+    /** Class name to be appended to the element */
+    className: PropTypes.string,
     /**
      * Applies fill styling for all GridCell children.
      * See GridCell for fill explanation
@@ -61,6 +63,7 @@ export default class Grid extends Component {
   render() {
     const {
       children,
+      className,
       fill,
       fit,
       full,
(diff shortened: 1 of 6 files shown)


stefanbuck:
To be honest, I think there is no need for such a component at all. Let the  developer decide if he wants to use FlexBox, CSS Grid or a table. Spacing can be ensured with CSS-Vars

https://github.com/BrandwatchLtd/axiom/blob/9f3007d3a56738612038bd1904f6cb22491dc336/packages/axiom-materials/src/sizing.css#L5-L11

PR #668, opened 2018-09-17
fix(TextInput): Add ability to focus TextInput in axiom.

MariamBashir:
This PR adds the ability to focus TextInput in axiom. This is to avoid looking for the `input` tag and using `findDOMNode` in every app that uses `TextInput` and instead providing a way to do this in axiom.

### DIFF
--- a/packages/axiom-components/src/Form/TextInput.js
+++ b/packages/axiom-components/src/Form/TextInput.js
@@ -1,5 +1,6 @@
 import PropTypes from 'prop-types';
 import React, { Component } from 'react';
+
 import { findComponent } from '@brandwatch/axiom-utils';
 import Base from '../Base/Base';
 import Validate from '../Validation/Validate';
@@ -17,6 +18,8 @@ export default class TextInput extends Component {
     disabled: PropTypes.bool,
     /** See Validate[error] */
     error: PropTypes.func,
+    /** Pass this prop to get ref to the Text Component instance. */
+    inputRef: PropTypes.object,
     /** Applies styling to indicate the users input was invalid */
     invalid: PropTypes.bool,
     /** Adds a progress indicator the the right of the text input */
@@ -116,6 +119,7 @@ export default class TextInput extends Component {
       usageHint,
       usageHintPosition,
       value,
+      inputRef,
       ...rest
     } = this.props;
 
@@ -154,6 +158,7 @@ export default class TextInput extends Component {
               { !showOnClear && showIcon && icon }
               <Base { ...rest }
                   Component="input"
+                  baseRef={ inputRef }
                   className="ax-input"
                   disabled={ disabled }
                   onBlur={ this.handleOnBlur.bind(this) }
--- a/packages/axiom-components/src/Base/Base.js
+++ b/packages/axiom-components/src/Base/Base.js
@@ -19,6 +19,8 @@ export default class Base extends Component {
       PropTypes.string,
       PropTypes.func,
     ]),
+    /** Pass this prop to get ref to the Base Component instance. */
+    baseRef: PropTypes.object,
     /** Class name to be appended to the element */
     className: PropTypes.string,
     /** Adds ability to make an element invisible */
@@ -161,6 +163,7 @@ export default class Base extends Component {
       textUnderline,
       theme,
       visibleUntil,
+      baseRef,
       ...rest
     } = this.props;
 
@@ -202,7 +205,7 @@ export default class Base extends Component {
     }
 
     return (
-      <Component { ...rest } className={ classes } />
+      <Component { ...rest } className={ classes } ref={ baseRef } />
     );
   }
 }


packages/axiom-components/src/Form/TextInput.js
```diff
@@ -78,6 +82,11 @@ export default class TextInput extends Component {
     this.state = { hasFocus: false };
   }
 
+  focusTextInput(){
```
lpoulter: `focusTextInput` how is this called? Or am I missing something?

lpoulter: We probably want a focus option passed in? Having more than one textInput with the option would cause issues thou. I think it would pick the last focused in the DOM.

MariamBashir: yeah i should have left some explanation. You can get the `TextInput` in a ref and you will call this method on the ref, something like below. I was testing it with `ComponentPicker`.

```
class Search extends Component {
    focusSearch() {
        this._searchRef.focusTextInput();
    }

    render() {
        const { onSearchChange, onKeyPress, searchValue, onIconClick } = this.props;
        const iconProps = searchValue
            ? { align: 'right', name: 'cross', onClick: onIconClick }
            : { align: 'left', name: 'magnify-glass' };
        return (
            <TextInput
                onChange={event => onSearchChange(event.target.value)}
                onKeyDown={event => onKeyPress(event.key)}
                placeholder="Search Component by name"
                value={searchValue}
                ref={el => (this._searchRef = el)}
            >
                <TextInputIcon {...iconProps} />
            </TextInput>
        );
    }
}
```

lpoulter: Ah ok I see is it not better to pass as an option? and call `focusTextInput` from within `TextInput`

MariamBashir: You mean a prop like a  `bool` value to `focus` when its `true`?

lpoulter: yeah exactly. Sometihng like this
```
  componentDidMount(){
    const { focus } = this.props;
    if (focus) {
      this.focusTextInput();
    }
  }

Comment (MariamBashir) [reply packages/axiom-components/src/Form/TextInput.js]: But this means that this would only happen once in the `componentDidMount`. what if there is a use case where the `TextInput` has to be focused when a user presses a key?

**Comment by lpoulter** [comment]
Just to echo discussion elsewhere `React.forwardRef` should be a nice solution. Allowing us to target the underlying DOM element from outside Axiom. Example here.

https://codesandbox.io/s/mzo1yq2yr9`

File: packages/axiom-components/src/Form/TextInput.js [hunk 2]
```diff
@@ -152,12 +156,13 @@ export default class TextInput extends Component {
...
               }
               { !showOnClear && showIcon && icon }
-              <Base { ...rest }
+              <BaseWithRef { ...rest }
                   Component="input"
                   className="ax-input"
                   disabled={ disabled }
                   onBlur={ this.handleOnBlur.bind(this) }
                   onFocus={ this.handleOnFocus.bind(this) }
+                  ref={ forwardedRef }
```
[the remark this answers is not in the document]
lpoulter: Yes you are right Rob. The way I understand it is using `forwardRef()` allows us to use the ref prop at a higher level and pass/foreward it down to lower level component. Allowing syntax like `<TextInput ref={myRef}>` with myRef being forwarded on rather that being assigned to the `TextInput` component. However we are only partially doing that here. As it seems to be the convention forwarding the ref from `TextInput`, to allow `<TextInput ref={myRef}>` might also be a good idea.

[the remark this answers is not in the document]
lpoulter: Your way, using `<TextInput forwardedRef={myRef}>` and not needing to use `React.forwardRef` might be better thou. What do you think?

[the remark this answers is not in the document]
roburidge: Yeah, that makes sense. However the way this is implemented my understanding is ref will only be passed forward on the `BaseWithRef` component.
`<BaseWithRef ref={myRef}>` Will pass on the ref to the underlying component
`<TextInput ref={myRef}>` But this will not work, the ref will point to the `TextInput` component, not the underlying HTML `input`
I think we would need to use `React.forwardRef` in the `TextInput` component to capture the ref and pass it down. Which might be nice for apps using `TextInput` however it would be a breaking change to Axiom.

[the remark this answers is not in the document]
lpoulter: Yeah I was saying we are kinda only half doing the `ref` approach.

lpoulter:
The react docs say it's a breaking change to add foreward ref but i'm not sure it is if we are not currently using refs in this component? Either way I think we should go with the less drastic approach and not use foreward. So this format
```<TextInput inputRef={myRef}> ->
<Base baseRef={inputRef}> ->
<input ref={myRef}>.```
What do you think @MariamBashir @roburidge

roburidge:
@lpoulter I'm happy with that approach. I would love to add forward ref to `Base` and `TextInput` however if ref has been used to point to either of these instances somewhere else it would break that as it would change the ref to point at the underlying html `input` rather than the component instance.

I vote we go with custom ref props like @lpoulter suggests unless we want to do the extra work making sure noting is currently using a ref to specifically point at these component instances.

MariamBashir:
@lpoulter Yes i can definitely this approach a try. I'll try to update with the changes on Monday. 
Thanks.

MariamBashir:
@roburidge and @lpoulter, I updated the solution as suggested. Please have another look, Thanks.

PR #673, opened 2018-09-25
feat(Position): Add targetElement property that acts as a general target element

t-sauer:
Right now it is only possible to attach the `Position` component to an element within an existing React app. This PR makes it possible, to attach the `Position` component to any DOM element through the `targetElement` property, which makes it possible to place tooltips outside of the react tree for example.


packages/axiom-components/src/Dropdown/DropdownContext.js
```diff
@@ -64,6 +64,10 @@ export default class DropdownContext extends Component {
   }
 
   handleClick(event) {
+    if (!this.context.dropdownRef || !this.context.closeDropdown) {
```
t-sauer: This is actually broken right now in master as well and I wasn't sure if this fix is the correct way to do it. Basically if `DropdownContext` is placed in `Position` (like the docs are doing it), the context that is expected here is not provided (it is only provided if you use `DropdownContext` within `Dropdown`).

jdelight:
This is great @t-sauer - exactly what we need for the DOM. However, I wonder if we can use this opportunity to support both concept of a target HTML element **and** a reference object. 

https://popper.js.org/popper-documentation.html#referenceObject

This is particularly useful for us and the word cloud component which is a canvas element (i.e. each word doesn't have an associated DOM node). 

We have the virtual target bounds using `getBoundingClientRect` so we'll need to get an Axiom context component pointing at this virtual target.

Looking at the code - perhaps this is already supported in this PR? Maybe we could change the property from `targetElement` to `reference` (i.e. matching the Popper docs). 

image

t-sauer:
@jdelight That sounds totally reasonable and should already work with this PR. I will rename the prop and adjust the docs to make this more clear.

packages/axiom-components/src/Position/Position.js
```diff
@@ -47,14 +51,20 @@ export default class Position extends Component {
...
      * a valid position is found.
      */
     position: PropTypes.oneOf(['top', 'right', 'bottom', 'left']),
+    /**
+     * If Position should be attached to an element outside of the React context,
+     * either a HTMLElement reference or a Popper.js compatible referenceObject
+     * (see https://popper.js.org/popper-documentation.html#referenceObject) can
+     * be passed in.
+     */
+    reference: PropTypes.object,
```
tomru: Might it be worth to check if it's either a HTMLElement or something having the functions used by Popper.js ?

PR #675, opened 2018-10-03
Update to babel 7

stefanbuck:
I used `npx babel-upgrade --write --install` for the basic changes plus a few manual  changes were needed to make it working with this repo


package.json
```diff
@@ -18,18 +18,20 @@
...
+    "@babel/cli": "^7.0.0",
+    "@babel/core": "^7.0.0",
+    "@babel/plugin-proposal-class-properties": "^7.1.0",
+    "@babel/plugin-proposal-object-rest-spread": "^7.0.0",
+    "@babel/plugin-transform-classes": "^7.1.0",
+    "@babel/plugin-transform-modules-commonjs": "^7.1.0",
+    "@babel/polyfill": "^7.0.0",
+    "@babel/preset-env": "^7.1.0",
+    "@babel/preset-react": "^7.0.0",
+    "babel-core": "^7.0.0-bridge.0",
```
stefanbuck: `^7.0.0-bridge.0"` is needed for Jest see https://jestjs.io/docs/en/getting-started.html#using-babel

babel.config.js
```diff
@@ -0,0 +1,50 @@
+module.exports = {
```
stefanbuck: There is a breaking change for monorepos and renaming `.babelrc` to `babel.config.js` is the preferred way to solve it. 

https://github.com/facebook/jest/issues/6053#issuecomment-383632515

lpoulter: This file seems to have lost it's git history. It should stay if you
`git mv .babelrc babel.config.js`

lpoulter: Hmm i forget that git mv sucks with renames

package.json
mrtnrst: This can be set in a peerDependency rather than directly in your devDependencies.

stefanbuck: Thanks @mrtnrst, but I think this dependency is not needed in our case. This Pull Request is about updating babel in the repo root (it's a monorepo) and by that it doesn't affects our published packages just our tooling.

stefanbuck:
`yarn build:site` is failing with

```
  [651] ./site/index.ejs 1.28 kB {0} [depth 2] [built]
        [only some exports used: default]
        harmony import ./index.ejs [330] ./site/static.js 8:0-35

ERROR in ReferenceError: colors is not defined
    at Object.<anonymous> (evalmachine.<anonymous>:41:68527)
    at __webpack_require__ (evalmachine.<anonymous>:1:403)
    at Object.<anonymous> (evalmachine.<anonymous>:41:64218)
```

@HHogg do you have any idea what cause this issue?

stefanbuck:
I was able to narrow the issue down to this weird thing. When removing all Axiom related imports from [Documentation.js](https://github.com/BrandwatchLtd/axiom/blob/ba73be74959467cc50afaa5cdee8ee1c3529e98a/site/components/Documentation/Documentation.js) the build step works 😕

Binarytales:
@tomru Is this old PR superseded by the working you have been doing on ES Modules: https://github.com/BrandwatchLtd/axiom/pull/707

tomru:
@Binarytales nope, did not touch the babel version in there. just added a second build step for es modules

Binarytales:
I've not been able to resolve the colours issue so I'm going to close this as we'll be changing how the site publishing works soon and once that work has been completed we can revisit updating Babel.

PR #676, opened 2018-10-04
Add prop to make Position non-interactive

t-sauer:
This adds an option to make the Position component non-interactive for situations where you for example don't want mouseovers over a tooltip to trigger any events.

### DIFF
--- a/packages/axiom-components/src/Position/Position.js
+++ b/packages/axiom-components/src/Position/Position.js
@@ -39,6 +39,11 @@ export default class Position extends Component {
     flip: PropTypes.oneOf(['anticlockwise', 'clockwise', 'mirror']),
     /** Toggles the visibility of the PositionSource */
     isVisible: PropTypes.bool.isRequired,
+    /**
+     * Controls if the elements that are within the Position component
+     * are non-interactive
+     */
+    nonInteractive: PropTypes.bool,
     /** Controls the starting offset of the content */
     offset: PropTypes.oneOf(['start', 'middle', 'end']),
     /**
@@ -73,6 +78,7 @@ export default class Position extends Component {
     boundariesElement: 'viewport',
     enabled: true,
     flip: 'clockwise',
+    nonInteractive: false,
     offset: 'middle',
     position: 'top',
     showArrow:  false,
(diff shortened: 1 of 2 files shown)


packages/axiom-components/src/Position/Position.js
```diff
@@ -39,6 +39,11 @@ export default class Position extends Component {
     flip: PropTypes.oneOf(['anticlockwise', 'clockwise', 'mirror']),
     /** Toggles the visibility of the PositionSource */
     isVisible: PropTypes.bool.isRequired,
+    /**
+     * Controls if the elements that are within the Position component
+     * are non-interactive
+     */
+    nonInteractive: PropTypes.bool,
```
lpoulter: I think `isInteractive` is most consistent with other boolean variable names.

jdelight:
Hi @t-sauer - could you explain the situation this is fixing for you? I'm wondering if there are features which are so specific that they can just exist in "project" code. Instead of being part of the Axiom API. Perhaps if we had a class name you could bind to it would be easier to just have this CSS in the project code. Just a thought at this stage :) it's probably fine to merge this but would be interesting to know a bit more about the situation. Thanks :)

t-sauer:
Basically we want to have a tooltip, where we can mouse over without it interfering with the mouse events in the area behind it (i.e. we still want to capture the mouse events behind the tooltip as if the tooltip wasn't there).

I was under the impression that passing through class names through some components wasn't really desired, that's why I went with the custom prop. I am totally fine to change it so that the Position component makes use of a `className` prop.

packages/axiom-components/src/Position/Position.js
```diff
@@ -180,12 +186,13 @@ export default class Position extends Component {
   }
 
   render() {
-    const { children, enabled, isVisible, onMaskClick, showArrow, reference, ...rest } = this.props;
+    const { children, enabled, isVisible, nonInteractive, onMaskClick, showArrow, reference, ...rest } = this.props;
```
lpoulter: Is the intention is to just disable the tooltip or the target also? Perhaps we can add this prob lower down the tree instead?

t-sauer: Ah, we are using a custom target reference so that didn't come up for me. This PR is probably too specific for our use case and it makes more sense to go with a custom class as @jdelight proposed.

jdelight:
Cool thanks for explaining @t-sauer - I'm assuming this is for the situation where for the Volume over time component you have peak bubbles and tooltips on the same point?

>  I am totally fine to change it so that the Position component makes use of a className prop.

Yes I think this would be ideal in this scenario - it enables your feature to work but also unlocks future unique situations and helps keep the surface area of the Position component API relatively small.

PR #680, opened 2018-10-09
chore(build): Pre-process CSSNext style nesting

Binarytales:
[JIRA MC-202](https://jira.brandwatch.com/browse/MC-202) 

But do not pre-process custom properties as these are transpiled by consuming applications

### DIFF
--- a/postcss.prod.config.js
+++ b/postcss.prod.config.js
@@ -1,0 +1,16 @@
+module.exports = {
+  plugins: [
+    require('postcss-cssnext')({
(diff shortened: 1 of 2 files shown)


lpoulter:
Nice so is on `nesting` default in CSS-Next? Does this fix the issue with the `UserMenu`?

Binarytales:
I've been doing additional testing and this isn't working how I thought it would. :( Will do some more experimentation.

PR #683, opened 2018-10-16
feat(Icons): Adding new icons

Flawwles:
* format-bold
* format-italic
* format-title
* format-underline

[Demo](https://unruffled-murdock-4dfb36.netlify.com/docs/packages/axiom-components/icon)


Binarytales:
Thanks @Flawwles for your PR. Perhaps this should be addressed in a separate PR if it applies to existing SVGs in Axiom but the new SVGs have a lot of extra meta data that is added Sketch but not actually needed for display. 

You can run the SVG through a tool called SVGO that is roughly equivalent to ImageOptim. There is a nice Web UI for SVGO here: https://jakearchibald.github.io/svgomg/

Binarytales:
Having spoken to @Flawwles IRL we have decided to get this merged and we will raise image optimisation at the next WG meeting.

PR #685, opened 2018-10-17
fix(Button): set background on secondary button

tomru:
The secondary Button has been transparent until now along with tertiary and quaternary buttons.

For the tertiary and quaternary buttons I think it fine to blend in with another background, but for the secondary button this looks really weird.

This sets the background to the theme background.

### Before
image

### After

image

[Netlilfy Deploy](https://5bc6f960c6aed62d7719c1ed--brandwatch-axiom-tomru.netlify.com/)
 (I've change the background of documentation to shade-3 to make the change more obvious)

### DIFF
--- a/packages/axiom-components/src/Button/Button.css
+++ b/packages/axiom-components/src/Button/Button.css
@@ -168,6 +168,7 @@
 
 .ax-button--secondary {
   border-color: var(--color-theme-border);
+  background-color: var(--color-theme-background);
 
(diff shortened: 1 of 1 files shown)


lpoulter:
@mozzling

tomru:
Discussed in DM - transparent background is intended. So closing this for now.

For the use-case we're considering to switch to a tertiary button instead

PR #693, opened 2018-11-02
feat(UsageHint): Consume width on dropdown context.

bwilson-bw:
Currently there's no way to change the width of the UsageHint, it takes the default off of the dropdown  context which is `14.5rem`. In the frontend Analytics we would find the addition of this width very useful.

https://kind-hodgkin-6e93e5.netlify.com/docs/packages/axiom-components/usagehint/


jdelight:
@bwilson-bw would be good to have a netlify deploy of this and also to include this prop in the documentation. 
There are some notes on deploying with netlify here: https://github.com/BrandwatchLtd/axiom/blob/master/CONTRIBUTING.md#deployments

PR #697, opened 2018-11-23
AD: colored button icons

larister:
Netlify link: https://kind-booth-fc222a.netlify.com/docs/packages/axiom-components/icon/

This allows us to colour our icon buttons like this:

image

This is pretty much my first Axiom PR so be extra vigilant 👀 

The main issue I see with this is the confusion between the styling of the button (which has a colour component), and the icon colour itself.

Advice/criticism most welcome.


packages/axiom-components/src/Button/Button.js
```diff
@@ -29,7 +29,7 @@ export default class Button extends Component {
     shape: PropTypes.oneOf(['circle', 'rectangle', 'stadium']),
     /** Size of standard shape */
     size: PropTypes.oneOf(['small', 'medium', 'large', 'huge']),
-    /** Style of the Button, that affects it's coloring and sizing */
+    /** Style of the Button, which affects its coloring and sizing */
```
lpoulter: 🐝

packages/axiom-components/src/Icon/IconButton.js
```diff
@@ -16,18 +16,64 @@ export default class IconButton extends Component {
...
   }
 
   render() {
-    const { name, size, ...rest } = this.props;
+    const { name, size, buttonStyle, iconColor, ...rest } = this.props;
 
     return (
-      <Button style="secondary" { ...rest } shape="circle" size={ size }>
-        <ButtonIcon name={ name } size={ sizeMap[size] } />
+      <Button { ...rest } shape="circle" size={ size } style={ buttonStyle }>
```
lpoulter: I think we will need to set `style:secondary` in the defaultProps to avoid changing existing buttons.

larister: yes good catch 👍

larister:
Huh apologies, linting's not working for me for some reason. Will dig into it and fix...

larister:
Thumbs up from Matt Evans as well 👍

PR #701, opened 2018-12-17
fix(Validate): Make validation on form element dynamic

lpoulter:
If you change the `required` property of a form input the `Validate` component does not update to reflect the change.

Each form input has a `Validate` wrapper.
A forms `Validates` are managed by the `Validation` component.

A forms `Validates` are now added when in `componentDidUpdate` lifecycle method. Previously `Validation` stored a form `Validators` as an array and used the index to manage removal and edit. `Validators` is now a map with unique ID's.

Example use case, radio button makes `in new folder` input required.

screen shot 2018-12-18 at 01 12 30

### DIFF
--- a/packages/axiom-components/src/Validation/Validate.js
+++ b/packages/axiom-components/src/Validation/Validate.js
@@ -1,5 +1,6 @@
 import PropTypes from 'prop-types';
 import { Component } from 'react';
+import uuid from 'uuid';
 
 export default class Validate extends Component {
   static contextTypes = {
@@ -44,11 +45,8 @@ export default class Validate extends Component {
   constructor(props) {
     super(props);
     this.checkPatternMet = this.checkPatternMet.bind(this);
-    this.indexSetter = this.indexSetter.bind(this);
     this.validationGetter = this.validationGetter.bind(this);
-    this.state = {
-      index: -1,
-    };
+    this.id = uuid();
   }
 
   componentWillMount() {
@@ -56,7 +54,7 @@ export default class Validate extends Component {
 
     this.context.registerValidate(
       this.validationGetter,
-      this.indexSetter,
+      this.id,
     );
   }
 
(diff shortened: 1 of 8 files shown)


packages/axiom-components/src/Validation/Validate.js
```diff
@@ -69,6 +69,18 @@ export default class Validate extends Component {
     );
   }
 
+  componentDidUpdate(prevProps) {
+    if (prevProps.required !== this.props.required) {
+      if (this.shouldValidate()){
+        return this.context.registerValidate(
```
lpoulter: spotted a bug with this.

packages/axiom-components/src/Validation/Validate.js
```diff
@@ -65,16 +63,28 @@ export default class Validate extends Component {
...
+  componentDidUpdate(prevProps) {
+    if (prevProps.required !== this.props.required) {
+      if (this.props.required || this.props.patterns) {
+        return this.context.registerValidate(
+          this.validationGetter,
+          this.id,
+        );
+      }
+
+      return this.context.unregisterValidate(
```
MariamBashir: I am little confused here. 🐢

lpoulter: `Validation` has an array of all `Validate` components in a form. Each `Validate` component can have a `required` prop to say the form element it wraps is required. 

The two methods `registerValidate` and `unregisterValidate` are used to tell `Validation` to update it's list of `Validate` components. This used to only happen on component mount, now it happens when a `Validates` required prop updates.

MariamBashir: Okay, so its going to `unregisterValidate` on every update of the component, even if the `required` prop didn't change?

MariamBashir: No i see it. Makes sense. ignore my comment 🙈

PR #704, opened 2019-01-16
Adding EllipsisTooltip

tomru:
Adding a `EllipsisTooltip` that shows a tooltip when the text clipped due to not enough width available.

image

image

This is needed for DropdownMenuItems that is forced to have a fixed height (e.g. when used in combination with [react-window](https://github.com/bvaughn/react-window). Long content should be hidden with an ellipsis but but still be shown when hovered over.

[Demo EllipsisTooltip](https://brandwatch-axiom-tomru.netlify.com/docs/packages/axiom-components/ellispsistooltip)
[Demo inside Dropdown (lower right one)](https://brandwatch-axiom-tomru.netlify.com/docs/packages/axiom-components/dropdown)

### DIFF
--- a/packages/axiom-components/src/EllipsisTooltip/EllipsisTooltip.js
+++ b/packages/axiom-components/src/EllipsisTooltip/EllipsisTooltip.js
@@ -1,0 +1,82 @@
+import PropTypes from 'prop-types';
+import React, { Component } from 'react';
+
+import Base from '../Base/Base';
+import Tooltip from '../Tooltip/Tooltip';
+import TooltipContent from '../Tooltip/TooltipContent';
+import TooltipContext from '../Tooltip/TooltipContext';
+import TooltipSource from '../Tooltip/TooltipSource';
+import TooltipTarget from '../Tooltip/TooltipTarget';
+
+function elementHasEllipsis(ref = {}) {
+  const { current } = ref;
+  return current && current.scrollWidth > current.clientWidth;
+}
+
+export default class EllipsisTooltip extends Component {
+  static propTypes = {
+    children: PropTypes.node,
+    delay: PropTypes.bool,
+    theme: PropTypes.oneOf(['day', 'night']),
+    width: PropTypes.string,
+  };
+
+  static defaultProps = {
+    delay: true,
+    theme: 'night',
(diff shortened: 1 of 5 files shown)


packages/axiom-components/src/EllipsisTooltip/EllipsisTooltip.js
```diff
@@ -0,0 +1,82 @@
...
+    this.updateTooltipEnabled();
+  }
+
+  render() {
+    const { children, delay, ...rest } = this.props;
+    const { hasEllipsis } = this.state;
+
+    if (!hasEllipsis) {
+      return (
+        <Base baseRef={ this.container } textEllipsis { ...rest } onMouseOver={ this.updateTooltipEnabled }>
```
tomru: I tried to avoid to detect on `onMouseOver` but instead detect the ellipsis in `componentDidMount`.

Though when used inside a dropdown, the `scrollWidth` and `clientWidth` is still 0 as it's still hidden via popper.js.

packages/axiom-components/src/EllipsisTooltip/EllipsisTooltip.js
```diff
@@ -0,0 +1,82 @@
...
+import TooltipContext from '../Tooltip/TooltipContext';
+import TooltipSource from '../Tooltip/TooltipSource';
+import TooltipTarget from '../Tooltip/TooltipTarget';
+
+function elementHasEllipsis(ref) {
+  if (!ref || !ref.current) {
+    return false;
+  }
+
+  const { current } = ref;
```
t-sauer: Couldn't you remove the `if` above if you just do `const { current } = ref || {};` here since you check for a truthy value of `current` in the return statement anyway?

tomru: yeah, I think there is no win in keeping the if statement. I'll add a default parameter instead though...

tomru:
Design feedback: using the dark tooltip by default, but you can pass the theme if a white one is needed.

tomru:
Design feedback: width of the tooltip should be `auto` by default, but is also configurable.

PR #705, opened 2019-01-17
fix(ContextMenuItem): mouse events do not fire in firefox

tomru:
A [bug](https://bugzilla.mozilla.org/show_bug.cgi?id=1084196) in firefox cause mouse events to not fire below a button element causing the EllipseTooltip to not work inside a DropdownContextMenuItem in Firefox.

Generally it makes sense to use a button for a ContextMenuItem as long as it is a simple entry you can click.

But there are at least two use-cases that allow some sort of interaction inside the ContextMenuItem:

1. checkbox to activate ContextMenuItem
2. showing a tooltip on a truncated named (using the EllipseTooltip)

Though the definition of the button does [permit interactive content](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/button).

As there are use-cases where there is interaction in the ContextMenuItem it is reasonable to change the element of ContextMenuItem to a div and set the ARIA role to `Menuitem` instead to fix that bug.

### DIFF
--- a/packages/axiom-components/src/Context/ContextMenuItem.js
+++ b/packages/axiom-components/src/Context/ContextMenuItem.js
@@ -36,7 +36,7 @@ export default class ContextMenuItem extends Component {
 
     return (
       <Base { ...rest } { ...{ [contextMenuItemSelector]: true } }
-          Component="button"
+          aria-role="Menuitem"
           className={ classes }
           disabled={ disabled }
           onClick={ onClick }
--- a/packages/axiom-components/src/Context/__snapshots__/ContextMenu.test.js.snap
+++ b/packages/axiom-components/src/Context/__snapshots__/ContextMenu.test.js.snap
@@ -16,7 +16,8 @@ exports[`ContextMenu renders with defaultProps 1`] = `
     <div
       className="ax-context-menu ax-context-menu--padding-vertical-medium"
     >
-      <button
+      <div
+        aria-role="Menuitem"
         className="ax-context-menu__item"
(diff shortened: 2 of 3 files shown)


packages/axiom-components/src/Context/ContextMenuItem.js
```diff
@@ -36,7 +36,7 @@ export default class ContextMenuItem extends Component {
 
     return (
       <Base { ...rest } { ...{ [contextMenuItemSelector]: true } }
-          Component="button"
+          area-role="Menuitem"
```
siggysamson: Isn't that supposed to be `aria-role`?

tomru: oh man, for sure. Thanks for catching that!

packages/axiom-components/src/Context/__snapshots__/ContextMenu.test.js.snap
```diff
@@ -16,7 +16,8 @@ exports[`ContextMenu renders with defaultProps 1`] = `
     <div
       className="ax-context-menu ax-context-menu--padding-vertical-medium"
     >
-      <button
+      <div
+        area-role="Menuitem"
```
siggysamson: This snapshot needs to be updated 🙂

tomru:
For reference, this has been [fixed](https://bugzilla.mozilla.org/show_bug.cgi?id=1089326) in Firefox v66 which will be shipped soon. Verified on the the beta of v66

PR #707, opened 2019-01-22
AX-121 publish es modules

tomru:
👷 

This PR adds es modules along to the common js build. This enables bundlers like webpack, rollup etc. to do [tree-shaking](https://webpack.js.org/guides/tree-shaking/) and remove unused code from the build.

The approach is naive in the sense that it creates two directories, `dist-cjs` and `dist-esm` in which the relevant module format is built via babel separately. The `main` and `module` in the relevant `package.json` will point to the relevant `index.js`.

This is only applied to modules that are relevant to a consumer, i.e. `babel-plugin-axiom-imports`, `babel-plugin-transform-svg-axiom`, `eslint-config-axiom` and `stylelint-config-axiom` will still only have a common js build.

I'd be happy to get some feedback already

### TODO
- [ ] measure build improvements in analytics

### DIFF
--- a/.babelrc
+++ b/.babelrc
@@ -37,7 +37,13 @@
       plugins: [
         "@brandwatch/babel-plugin-transform-svg-axiom",
         "@brandwatch/babel-plugin-axiom-imports",
-        "transform-es2015-modules-commonjs"
+        "transform-es2015-modules-commonjs",
+      ]
+    },
+    "production-esm": {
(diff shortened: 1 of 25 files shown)


Binarytales:
This is great stuff @tomru. Will it change how applications consume Axiom at all?

tomru:
close for now

PR #709, opened 2019-01-29
Fix broken dist css for Tip Component

eliotstocker:


tomru:
👍 thanks for fixing the commit message

PR #712, opened 2019-02-08
docs(colors): clicking colour copies the variable name to clipboard

TomWija:


Flawwles:
This is a nice idea! 

Feels very hidden however, a quick win could be adding a tooltip when you hover the swatch saying `Copy variable name to clipboard`- also having the cursor change to a pointer would help make this feature easier to discover.

Would be nice if we defined just one click area, e.g. clicking the colour swatch copies to clipboard, but clicking on the text doesn't. The current implementation would actually stop me from selecting or copying the hex value text.

Would also be nice to make use of the `notification` component to let the user know the variable name has been copied.

Binarytales:
Thanks for the feedback @Flawwles - really useful and reminds me why I shouldn't gansta merge things last thing on a Friday.

@TomWija Are you able to incorporate these changes, I think at least the ability to highlight and copy the names and values should be restored.

PR #713, opened 2019-02-11
fix(ContextMenuItem): hide all overflow, not only overflow-x

tomru:
In https://github.com/BrandwatchLtd/axiom-react/pull/704 we added  `overflow-x: hidden` which unexpectetly causes `overflow-y` to compute to `auto` ( see note on [here](https://developer.mozilla.org/en-US/docs/Web/CSS/overflow-y) ) causing scroll bars to show everywhere, where a `ContextMenuItem` have filled with more content that there was space available.

This PR will also set `overflow-y: hidden` to guard against scrollbars in these cases.

### DIFF
--- a/packages/axiom-components/src/Context/Context.css
+++ b/packages/axiom-components/src/Context/Context.css
@@ -149,7 +149,7 @@
 
 .ax-context-menu__item-content {
   flex: 1 1 auto;
-  overflow-x: hidden;
+  overflow: hidden;
 }
 
 .ax-context-menu__item-checkbox {


tomru:
@larister maybe, still didn't understand 100% what's happening. I'm unsure about what actually makes the height of the context menu item to be smaller in those cases, so that the scrollbar even appears.

larister:
I haven't looked into it, and like I say happy for this to go in, but perhaps worth discussing in the WG if it needs further investigation?

tomru:
not sure what to discuss until I've not really understood it thb. I just feel stuck and another set of :eye: might help - @BrandwatchLtd/axiom anyone an idea on :point_up: ?

Binarytales:
@tomru Thanks for this. Let's get this in, and published, and fix the immediate problem.

PR #717, opened 2019-02-27
docs: Add "Better Together" colour

Binarytales:
screen shot 2019-02-27 at 2 31 45 pm

This also adds a little but of positive feedback to click on a color to get the variable name in your clipboard

### DIFF
--- a/packages/axiom-materials/src/colors.css
+++ b/packages/axiom-materials/src/colors.css
@@ -198,5 +198,7 @@
   --color-social-instagram--active: rgb(171, 58, 133);
   --color-social-instagram: rgb(183, 71, 146);
   --color-social-instagram--hover: rgb(196, 86, 160);
+
+  --color-better-together: rgb(240, 78, 78);
 }
 /* stylelint-enable */
--- a/packages/axiom-materials/src/colors.js
+++ b/packages/axiom-materials/src/colors.js
@@ -171,6 +171,8 @@ export const socialInstagramActive = { r: 171, g: 58, b: 133 };
 export const socialInstagram = { r: 183, g: 71, b: 146 };
 export const socialInstagramHover = { r: 196, g: 86, b: 160 };
 
+export const betterTogether = { r: 240, g: 78, b: 78 };
+
 export const externalTwiiterBlue = socialTwitter;
 
 export const themeDayMain = uiCarbon;
--- a/site/components/Documentation/Resources/Materials/Colors/ColorGrid.css
+++ b/site/components/Documentation/Resources/Materials/Colors/ColorGrid.css
@@ -8,6 +8,11 @@
   margin: 0 auto;
   border: 0.0625rem solid var(--color-theme-border);
   border-radius: 50%;
+  cursor: pointer;
+}
+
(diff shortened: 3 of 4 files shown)


packages/axiom-materials/src/colors.js
```diff
@@ -171,6 +171,8 @@ export const socialInstagramActive = { r: 171, g: 58, b: 133 };
 export const socialInstagram = { r: 183, g: 71, b: 146 };
 export const socialInstagramHover = { r: 196, g: 86, b: 160 };
 
+export const betterTogther = { r: 240, g: 78, b: 78 };
```
siggysamson: Looks like a small typo here - the `e` got lost! `betterTogether`

site/components/Documentation/Resources/Materials/Colors/ColorGrid.css
```diff
@@ -8,6 +8,10 @@
   margin: 0 auto;
   border: 0.0625rem solid var(--color-theme-border);
   border-radius: 50%;
+  cursor: pointer;
+}
+.si-color-dot:active {
+ box-shadow: 0 0 0 3px var(--color-theme-border);
```
siggysamson: If I'm remembering this correctly we didn't use `px` much in Axiom and usually went with `rem`. While I don't mind it much it's probably worth considering that for the sake of consistency

packages/axiom-materials/src/colors.css
```diff
@@ -198,5 +198,7 @@
   --color-social-instagram--active: rgb(171, 58, 133);
   --color-social-instagram: rgb(183, 71, 146);
   --color-social-instagram--hover: rgb(196, 86, 160);
+
+  --colour-better-together: rgb(240, 78, 78);
```
t-sauer: Shouldn't this be `color` to stay consistent with the other names?

packages/axiom-materials/src/colors.js
t-sauer: Small typo here (you are missing an `e` in `Together`)

site/components/Documentation/Resources/Materials/Colors/ColorGrid.css
t-sauer: Indentation seems to be off here.

roburidge: nice addition 👍

Binarytales:
Thanks for comments @t-sauer - This is why I shouldn't code with a cold 😰

site/components/Documentation/Resources/Materials/Colors/ColorGrid.css
Binarytales: Indeed. The linter picked this up to and I have switch to `rem`

Binarytales:
@siggysamson I wasn't sure about the `docs` vs `feat` thing. Katja asked for this and it's a colour that should _not_ ever be used in real UIs.

siggysamson:
Oohh interesting! Makes sense to me. In that case I'd leave it with `docs`! Thanks for clarifying

PR #718, opened 2019-03-04
feat(packaging): add es modules in build

tomru:
### Tasks
- set babel environment in scripts directly
- renamed commonjs babel env to `production-cjs`
- build commonjs in `dist-cjs` instead of `dist`
- add an es-module build step and build to `production-esm`

### TODO
- [ ] is `scripts/ci_build.sh` really needed?
- [ ] run `postcss` just once
- [ ] append stats for analytics, audiences and vizia builds
- [ ] reconsider leaving commonjs build in `dist` OR do a major version bump


lpoulter:
This PR looks really cool @tomru what's the current status with it? Is there anything I can help with?

tomru:
Thanks @lpoulter - I just need to work through the todo's. 

Do you have thoughts in keeping `dist` or rather go with `dist-cjs` for clarity?

Also I need to test our apps against the es build and collect the stats with webpack bundle analyzer.

lpoulter:
I think `dist-cjs` and `dist-esm` is clearer than `dist` and `dist-esm`. I can test this in audiences looking at our bundle size is what brung me here. Do you think it's strange that the css has to be transpiled by the end user of Axiom? Would it make sense to do that in the dist files?

tomru:
moving this to BrandwatchLtd/axiom instead ...

PR #720, opened 2019-03-12
chore: Bump Node version to LTS/Dubnium

Binarytales:


.nvmrc
```diff
@@ -1 +1 @@
-lts/carbon
+lts/dubnium
```
lpoulter: Is this the same as saying `10.x`? Do you think it's clearer in future if we use version numbers instead of codename?

lpoulter:
Looks like there was an issue with lerna on 10.x. Do you know if that's been resolved.
https://github.com/BrandwatchLtd/axiom-react/pull/641

Binarytales:
@lpoulter I was not aware and I do not know if it's been resolved. I will investigate.

tomru:
> Looks like there was an issue with lerna on 10.x. Do you know if that's been resolved.
> #641

that was related to the various things we tried to get publishing to work - which was causing issue locally with 10.x back then. But as publishing on Travis still has other issues, this shouldn't hold us back bumping node now I'd say...

PR #722, opened 2019-03-12
Flags

stefanpearson:
- Adds a build step to generate flag assets. This will process a dir of flags with a 1:1 aspect ratio from an `flag-svg-collection` npm module. It replaces the parsed colours with the nearest Axiom product / base colour from `axiom-materials` and dumps the result into the Axiom site to be statically hosted.
- Adds a flag component. This is just a wrapped image component for now that takes an alpha2 country code. If anyone wants to mend the visuals of this component, feel free (cc [@GrandeCapuccino](https://github.com/GrandeCapuccino), [@dpodboj1](https://github.com/dpodboj1)).
- The source flag SVGs can be swapped out at any point if design wants a new look (as long as the alpha2 codes match).

**NB** This is fairly unconventional because the output assets (a dir of SVGs and lookup JSON) will need to be generated before the axiom packages and site build steps. For the moment, flags will resolve from `https://axiom.brandwatch.com/assets/flags/<alpha2>.svg`. It would be a lot cleaner if the SVG markup is baked in to the component (like the icons), but unfeasible as the generated SVGs come in at ~4mb.


.babelrc
```diff
@@ -22,26 +22,29 @@
   "env": {
     "development": {
       "plugins": [
-        "@brandwatch/babel-plugin-transform-svg-axiom",
+        "./packages/babel-plugin-transform-svg-axiom",
```
stefanpearson: I don't know how this worked for everyone else, but webpack just would not resolve these unless I made a local path.

scripts/process_flags/index.js
```diff
@@ -0,0 +1,59 @@
...
+  const map = items.reduce((memo, item) => {
+    return Object.assign(memo, {
+      [item.id]: {
+        src: `${assetsBaseUrl}/${item.id}.svg`,
+        color: item.color,
+      },
+    });
+  }, {});
+  return Promise.all([
+    fs.writeFile(path.join(destMapPath, 'src', 'flags.json'), JSON.stringify(map)),
```
stefanpearson: Needs to write to `src` and `dist` because the tests run differently on local and CI (?)

stefanpearson:
Good spot. This is expected and I don't like it, but I'm not sure how to do make this work in every environment. Right now, the SVG URLs are hardcoded with a base URL that resolves to the Axiom docs assets URL. I'll have a think about how this could work, because it's not a good dev experience.

stefanpearson:
Fixed ^

PR #724, opened 2019-03-18
Duration Picker

stefanpearson:
Adds a Duration Picker component, for Control Circus.

image


packages/axiom-components/src/DurationPicker/DurationPicker.js
```diff
@@ -0,0 +1,219 @@
...
+
+const getIsoDurationValueFromState = (state) => {
+  const { selectedUnit, selectedValue } = state;
+
+  if (!validTimeUnits.includes(selectedUnit) || !selectedValue) {
+    return '';
+  }
+
+  const duration = Duration.fromObject({
+    [state.selectedUnit]: state.selectedValue,
```
lpoulter: we can use the destructed values here.

packages/axiom-components/src/DurationPicker/DurationPicker.js
```diff
@@ -0,0 +1,219 @@
...
+  }
+
+  render() {
+    const enabledTimeUnits = {
+      minutes: this.props.minutes,
+      hours: this.props.hours,
+      days: this.props.days,
+      weeks: this.props.weeks,
+      months: this.props.months,
+      years: this.props.years,
```
lpoulter: We could destructor this.props here or prehaps it would be better to pass an `enabledTimeUnits` object as one prop rather than infer it here?

packages/axiom-components/src/DurationPicker/DurationPicker.js
```diff
@@ -0,0 +1,219 @@
...
+    /** Enable months option */
+    months: PropTypes.bool,
+    /** Invoked with the ISO duration value when the it is computed */
+    onChange: PropTypes.func,
+    /** Valid ISO duration value (see https://en.wikipedia.org/wiki/ISO_8601#Durations) [DISABLED] */
+    value: PropTypes.string,
+    /** Enable weeks option */
+    weeks: PropTypes.bool,
+    /** Enable years option */
+    years: PropTypes.bool,
```
lpoulter: We could make this one object then validate it against `validTimeUnits`

packages/axiom-components/src/DurationPicker/DurationPicker.js
```diff
@@ -0,0 +1,219 @@
...
+  const selectedUnit = [].concat(validTimeUnits)
+    .reverse()
+    .find((timeUnit) => duration.values[timeUnit]);
+  const selectedValue = duration.values[selectedUnit];
+
+  return {
+    selectedUnit,
+    selectedValue,
+    value,
+  };
```
lpoulter: It looks like this will over-right the state. That may cause problems in the future. I think it would be better to remove the concept of state from this function and spread the return object into state elsewhere.

packages/axiom-components/src/DurationPicker/DurationPicker.js
lpoulter: Yes good point. Perhaps we could do it like this.
`removedUnits={["hours", "weeks"]}`

lpoulter:
```
render(){
    const { removedUnits } = this.props;

    const filteredTimeUnits = validTimeUnits.filter((timeUnit) => !removedUnits.includes(timeUnit));
...
```
```
    removedUnits:  PropTypes.arrayOf( PropTypes.oneOf([...validTimeUnits]) ),
```

packages/axiom-components/src/DurationPicker/DurationPicker.js
stefanpearson: Right I see what you mean, I thought setState assigns over original state, will fix

packages/axiom-components/src/DurationPicker/DurationPicker.js
```diff
@@ -0,0 +1,206 @@
...
+  onChangeValue(event) {
+    const inputValue = parseInt(event.target.value, 10);
+    const selectedValue = Number.isFinite(inputValue) && inputValue >= 1 ? inputValue : '';
+    const durationValue = getIsoDurationValueFromState({
+      selectedUnit: this.state.selectedUnit,
+      selectedValue,
+    });
+
+    this.setState({
+      ...this.state,
```
lpoulter: Sorry I worded my previous comment poorly. This is good but you don't need to spread this.state here. setState will merge the old state in for you.

stefanpearson: Ah right… so just remove `...this.state`?

packages/axiom-components/src/DurationPicker/DurationPicker.js
```diff
@@ -0,0 +1,206 @@
...
+    if (prevProps.value !== this.props.value) {
+      this.setState({
+        ...this.state,
+        ...getStateFromIsoDurationValue(this.props.value),
+      }, this.onChange);
+    }
+  }
+
+  render() {
+    const excludedTimeUnits = this.props.excludedTimeUnits.concat('seconds');
```
lpoulter: I'm not sure why we concat `seconds` here? won't this have the affect of always excluding `seconds`?

packages/axiom-components/src/DurationPicker/DurationPicker.js
```diff
@@ -0,0 +1,206 @@
...
+
+const formatTimeUnit = (timeUnit) => {
+  return `${timeUnit.slice(0, 1).toUpperCase()}${timeUnit.slice(1)}`;
+};
+
+export default class DurationPicker extends Component {
+  static propTypes = {
+    /** Exclude time units */
+    excludedTimeUnits: PropTypes.arrayOf(
+      PropTypes.oneOf(['minutes', 'hours', 'days', 'weeks', 'months', 'years'])
```
lpoulter: Should seconds be here? Can you just ...validTimeUnits. To ensure all valid units can also be turned off.

packages/axiom-components/src/DurationPicker/DurationPicker.js
stefanpearson: Yeah, it's still a 'valid time unit' according to an ISO duration, but seconds doesn't seem like a useful unit for most cases (e.g. show me the last 30 seconds of data). I could remove this for simplicity but I can see every developer adding this themselves, which creates a bit of boilerplate in project code.

stefanpearson:
I can explain some of the use-cases for Vizia, where we have to handle different data sources:

* For APIs that don't handle breakdowns smaller than a day (quite a lot of them), you would want to exclude ['hours', 'minutes'].
* For BW, or any other API that handles more granular data querying, you shouldn't need to exclude any.

So I figured it was a more common use-case to exclude a few, rather than specifying each time what units to include (and having to dig through the API to find out which ones).

If this isn't consistent with the general Axiom API, I'm happy to change it

packages/axiom-components/src/DurationPicker/DurationPicker.js
lpoulter: Yep setState does that for you.

packages/axiom-components/src/DurationPicker/DurationPicker.js
lpoulter: I see that makes sense. However this looks like it has the affect of never allowing seconds. Prehaps you could you set ` excludedTimeUnits: ['seconds'],` in the default props, I think it's fine it that get's over written if the user includes there own `excludedTimeUnits` or you could just remove seconds altogether if you think it's not gonna be used a alot.

stefanpearson: Ok, I've done this to keep the API cleaner.

PR #727, opened 2019-03-20
feat(Icon): add support for custom width/height

tomru:
Add the option to set the `height` and/or `width` instead of just `size` for an icon. 

Useful for non-square icons like the iris icon added in #726

### DIFF
--- a/packages/axiom-components/src/Icon/Icon.js
+++ b/packages/axiom-components/src/Icon/Icon.js
@@ -13,6 +13,8 @@ export default class Icon extends Component {
      * Inline styling that allows the Icon to placed next to other inline
      * elements or text.
      */
+    /** Height of icon (with unit). */
+    height: PropTypes.string,
     inline: PropTypes.bool,
     /** Name of the icon that will be displayed */
     name: PropTypes.oneOf([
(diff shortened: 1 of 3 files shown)


packages/axiom-components/src/Icon/Icon.js
```diff
@@ -13,6 +13,8 @@ export default class Icon extends Component {
      * Inline styling that allows the Icon to placed next to other inline
      * elements or text.
      */
+    height: PropTypes.string,
+    /** Height of icon (with unit). */
```
lpoulter: Should this be above the prop?

packages/axiom-components/src/Icon/Icon.js
```diff
@@ -110,21 +112,26 @@ export default class Icon extends Component {
       'warning',
       'warning-circle',
     ]).isRequired,
-    /** Size of icon (with unit) */
+    /** Size of icon (with unit). Not used when height or width is provided */
```
lpoulter: 👍

packages/axiom-components/src/Icon/Icon.js
tomru: oops, indeed. Thanks!

PR #730, opened 2019-03-21
fix: builds flags before publishing

stefanpearson:
The build dist is currently broken because a build step was missed before the last publish, this will ensure that happens. I think.

Also, is there a time frame on auto publishing from CI?

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -17,7 +17,6 @@ script:
 - yarn ci:test
 - yarn build:site
 after_success:
-- bash scripts/publish_packages.sh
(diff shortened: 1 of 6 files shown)


tomru:
Oh darn, I'm guilty in missing the new step when I've published last, sorry 😞 

@stefanpearson as it's mandatory to build the flags prior to building the packages, I'm wondering if `build_flags.sh` should just be ran inside `build:packages` maybe?

stefanpearson:
That's a fair point, I'll see if I can move it all around.

PR #731, opened 2019-03-25
fix: forces blur when unit option is selected

stefanpearson:
I noticed the blur handler was triggered before the change handler, so any parent components managing the state received blur before changes had occurred.

### DIFF
--- a/packages/axiom-components/src/DurationPicker/DurationPicker.js
+++ b/packages/axiom-components/src/DurationPicker/DurationPicker.js
@@ -1,5 +1,5 @@
 import PropTypes from 'prop-types';
-import React, { Component } from 'react';
+import React, { PureComponent } from 'react';
 import { Duration } from 'luxon';
 import omit from 'lodash.omit';
 import Dropdown from '../Dropdown/Dropdown';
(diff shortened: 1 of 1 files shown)


lpoulter:
Not sure about this doesn't feel right to trigger an onBlur when a value has changed. They way i understand it browsers are inconsistant in the order of onBlur and onChange, have you got an example of what sort bug this fixes?

stefanpearson:
My issue is that I need to trigger an expensive action only on blur, and I'm getting a race condition because of `setState` being async. The action is supposed to get the current state of the application and do something with it. The problem is that the `onBlur` is happening before the `onChange`. So the action being triggered does not have the desired state yet.

I figured selecting something in a dropdown counts as a blur, so the blur is triggered explicitly in the handler (after triggering `props.onChange`).

It might help if I explain it to you in person? I'm a bit of a React n00b and trying to get this to play nice with Control Circus is a bit of a learning curve.

stefanpearson:
Closed this as Control Circus has a unique use-case and it's better to handle the idiosyncrasies there instead.

PR #736, opened 2019-04-02
feat(Checkbox, RadioButton): change style of selected and active

tomru:
This updates the design of radio and checkboxes.

image

image

### DIFF
--- a/packages/axiom-components/src/Form/ChedioButtox.css
+++ b/packages/axiom-components/src/Form/ChedioButtox.css
@@ -90,14 +90,15 @@
   transition-timing-function: var(--transition-function);
 }
 
-.ax-checkbox__input:focus + .ax-checkbox__indicator,
-.ax-radio__input:focus + .ax-radio__indicator {
-  border-color: rgba(var(--rgb-ui-accent), var(--opacity-input-border--focused));
-}
-
 .ax-checkbox__input:disabled + .ax-checkbox__indicator,
 .ax-radio__input:disabled + .ax-radio__indicator {
   background-color: var(--color-theme-background--disabled);
+}
+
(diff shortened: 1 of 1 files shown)


packages/axiom-components/src/Form/ChedioButtox.css
```diff
@@ -80,6 +94,15 @@
...
     transition-timing-function: var(--transition-function);
   }
+
+  @nest .ax-checkbox__input:checked + & {
+    border-width: 0;
+    background-color: var(--color-ui-accent);
+  }
+  @nest .ax-checkbox__input:checked:focus + & {
+    border-width: var(--component-border-width);
+    background-color: var(--color-ui-white);
```
dpodboj1: Should this also be `--color-ui-white-noise`?

tomru: Good point, will verify with design

tomru:
missed stylelint errors...

packages/axiom-components/src/Form/ChedioButtox.css
lilybrandwatch: I confirm use the color `--color-ui-white-noise`.
About the selected radio button style, the proportion between the blue and the white should be slightly different, which means that the white circle should be of 6px so the exact opposite of the old style if that makes sense. :)

tomru:
There has been some confusion on this. TIL there seems to be a lingo issue between designers and devs. Designers use the term `focused` for elements that have the state `active` i.e. when a item is clicked, opposed to having the focus as outlined [here](https://developer.mozilla.org/en-US/docs/Mozilla/Tech/XUL/Tutorial/Focus_and_Selection#Focused_Elements)

PR #737, opened 2019-04-08
Sw 641 add range slider to axiom

clairebw:
This adds a Range slider with two handles.
Screen Shot 2019-04-08 at 15 47 50


packages/axiom-components/src/Slider/Range.js
```diff
@@ -0,0 +1,228 @@
...
+    if (values[0] > values[1]) {
+      switch (this.focusedHandleIndex) {
+      case 0:
+        values[0] = values[1];
+        break;
+      case 1:
+        values[1] = values[0];
+        break;
+      default:
+        return;
```
lpoulter: Might be nice to throw/log an error here.

packages/axiom-components/src/Slider/Range.js
```diff
@@ -0,0 +1,228 @@
...
+
+    this.onChange(this.getValues(clientX));
+  }
+
+  handleKeyDown(event) {
+    const { step } = this.props;
+    const { values } = this.state;
+
+    switch (event.key) {
+    case 'ArrowLeft':
```
lpoulter: Formatting looks strange here should `case` be indented?

lpoulter:
Hi @clairebw thanks for this. Looks like there are some lint and test issues on the build. Are you able to loop in design and link the new component in the docs once it build?

packages/axiom-components/src/Slider/Range.js
clairebw: The eslint rule explicitly requires that there is no indentation for `switch` cases: https://github.com/clairebw/axiom-react/blob/04ef2645fbce630b80043fd6539262c27ddf348a/packages/eslint-config-axiom/src/eslintrc.json#L18

clairebw:
@lpoulter @TomWija @MariamBashir Thanks for your reviews. I fixed the tests, addressed the comments and @lilybrandwatch also approved it. Would you mind having a second look?

PR #740, opened 2019-04-17
feat(Chip): Add chip component

cezarykubel:
A new chip component, wrapped in a auto-collapsing chip list component.  This was originally used for AD-4010, the topics in the details view work.  

Some examples of the different options:
image
image
image

Some things to note:
- if there are not enough chips to have a multiline list, the list will not be collapsible
- you can use the `size` prop on both individual Chip components and the whole ChipList component.  The latter will apply the size to all Chips within the container.  
- if `activeColor` is not set, the default is 'accent' and solid.  However, if an `activeColor` is set, the Chip has a "outline" style, as per other use-cases.  

Preview [here](https://deploy-preview-740--serene-shannon-8198f2.netlify.com/docs/packages/axiom-components/chip)

### DIFF
--- a/packages/axiom-components/src/Chip/Chip.css
+++ b/packages/axiom-components/src/Chip/Chip.css
@@ -5,7 +5,6 @@
   margin-right: var(--cmp-inline-spacing);
   margin-bottom: var(--cmp-inline-spacing);
   padding: 0;
-  float: left;
   outline: 0;
   border-width: var(--component-border-width);
   border-style: solid;
--- a/packages/axiom-components/src/Chip/Chip.css
+++ b/packages/axiom-components/src/Chip/Chip.css
@@ -1,0 +1,116 @@
+.ax-chip {
+  display: flex;
+  align-items: center;
+  justify-content: space-between;
+  margin-right: var(--cmp-inline-spacing);
+  margin-bottom: var(--cmp-inline-spacing);
+  padding: 0;
+  float: left;
+  outline: 0;
+  border-width: var(--component-border-width);
+  border-style: solid;
+  border-radius: var(--component-border-radius-huge);
+  border-color: transparent;
+  background-color: var(--color-ui-white-noise);
+  color: var(--color-ui-carbon);
+  letter-spacing: var(--letter-spacing-loose);
+  text-align: center;
+  line-height: var(--component-line-height);
+  white-space: nowrap;
+  vertical-align: middle;
+  transition-property: background-color, border-color, color;
+  transition-duration: var(--transition-time-base);
+  transition-timing-function: var(--transition-function);
+
+  &:hover {
+    background-color: var(--color-ui-white-noise--dark);
+    cursor: pointer;
+  }
+
+  &[disabled],
+  &[disabled]:hover,
+  &[disabled]:active {
+    cursor: default;
+  }
+
+  &[disabled]:hover {
+    background-color: var(--color-ui-white-noise);
+  }
+}
+
+.ax-chip-metric {
+  margin-right: var(--spacing-grid-size--x1);
+  margin-left: -var(--spacing-grid-size--x1);
+  color: var(--color-theme-text--subtle);
+}
+
+.ax-chip-metric--light {
+  color: var(--color-ui-white);
+}
+
+.ax-chip--active {
+  background-color: var(--color-ui-white);
+  color: var(--color-ui-carbon);
+
+  &:hover {
+    background-color: var(--color-ui-white);
+  }
+}
+
+.ax-chip--active.ax-chip-color--accent {
+  background-color: var(--color-ui-accent);
+  color: var(--color-ui-white);
+
+  &:hover {
+    background-color: var(--color-ui-accent--hover);
+  }
+}
+
+.ax-chip--active.ax-chip-color--highlight { border-color: var(--color-product-highlight); }
+.ax-chip--active.ax-chip-color--success { border-color: var(--color-product-success); }
+.ax-chip--active.ax-chip-color--error { border-color: var(--color-product-error); }
+.ax-chip--active.ax-chip-color--forbidden-planet { border-color: var(--color-product-forbidden-planet); }
+.ax-chip--active.ax-chip-color--tiny-clanger { border-color: var(--color-product-tiny-clanger); }
+.ax-chip--active.ax-chip-color--critical-mass { border-color: var(--color-product-critical-mass); }
+.ax-chip--active.ax-chip-color--fantastic-voyage { border-color: var(--color-product-fantastic-voyage); }
+.ax-chip--active.ax-chip-color--paradise-lost { border-color: var(--color-product-paradise-lost); }
+.ax-chip--active.ax-chip-color--serene-sea { border-color: var(--color-product-serene-sea); }
+.ax-chip--active.ax-chip-color--event-horizon { border-color: var(--color-product-event-horizon); }
+.ax-chip--active.ax-chip-color--electric-dreams { border-color: var(--color-product-electric-dreams); }
+.ax-chip--active.ax-chip-color--outer-limits { border-color: var(--color-product-outer-limits); }
+.ax-chip--active.ax-chip-color--giant-leap { border-color: var(--color-product-giant-leap); }
+.ax-chip--active.ax-chip-color--moon-lagoon { border-color: var(--color-product-moon-lagoon); }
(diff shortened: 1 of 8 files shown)


packages/axiom-components/src/Chip/ChipList.css
```diff
@@ -0,0 +1,72 @@
+.ax-chiplist {
+  margin-top: var(--spacing-grid-size--x2);
+  margin-bottom: var(--spacing-grid-size--x2);
+  overflow: hidden;
+  &::after {
```
lpoulter: If we remove the floats and use flex on the list can we also remove the clearfix? 

```
.ax-chiplist {
		display: flex;
    	flex-wrap: wrap;
}
```

packages/axiom-components/src/Chip/ChipList.css
```diff
@@ -0,0 +1,72 @@
...
+    clear: both;
+  }
+}
+
+.ax-chiplist-label {
+  display: flex;
+  align-items: center;
+  justify-content: center;
+  margin-right: var(--spacing-grid-size--x1);
+  float: left;
```
lpoulter: I'm not sure we need to float this if we use flex?

larister:
Congrats on sneaking in some Celtics players into Axiom 😂

packages/axiom-components/src/Chip/ChipList.css
```diff
@@ -0,0 +1,72 @@
...
+  font-weight: var(--font-weight-bold);
+  line-height: var(--component-line-height);
+}
+
+.ax-chiplist--small,
+.ax-chiplist--medium,
+.ax-chiplist--large {
+  transition-property: max-height, opacity, margin;
+  transition-duration: var(--transition-time-slow);
+  transition-timing-function: var(--transition-function);
```
larister: Can't quite figure out why but it looks to me that on the demo at least the transition is only happening on "collapse", not on "expand"

packages/axiom-components/src/Chip/ChipList.js
```diff
@@ -0,0 +1,86 @@
...
+      label,
+      size,
+    } = this.props;
+
+    const {
+      isExpanded,
+      isExpandable,
+    } = this.state;
+
+    const childrenWithSize = React.Children.map(children, child => React.cloneElement(child, { size }));
```
larister: Cool, didn't know you could do that. Part of me wonders if it's worth checking the child is actually a chip but not sure it's worth it

packages/axiom-components/src/Chip/ChipList.js
```diff
@@ -0,0 +1,86 @@
...
+      `ax-chiplist-label--${size}`,
+      { 'ax-chiplist--expandable': isExpandable }
+    );
+
+    return (
+      <Base baseRef={ this.chipList } className={ classes }>
+
+        <div className={ labelClasses } onClick={ this.handleExpand }>
+          { label }
+          { isExpandable && (<Icon inline name={ isExpanded ? 'chevron-up' : 'chevron-down' } spaceLeft="x1" />) }
```
larister: would slightly prefer this was abstracted out into a fn rather than having multiple conditions inline like this but nabd

packages/axiom-components/src/Chip/ChipList.js
```diff
@@ -0,0 +1,86 @@
...
+    };
+  }
+
+  componentDidMount() {
+    const {
+      clientHeight,
+      scrollHeight,
+    } = this.chipList.current;
+
+    if (scrollHeight > (clientHeight * 2)) {
```
larister: Could we use `getBoundingClientRect()` instead? Feels like the multiplication by 2 is a little inexact; I might have misunderstood though

packages/axiom-components/src/Chip/ChipList.css
cezarykubel: fixed.  not quite sure why i used floats.  old habits? :/

packages/axiom-components/src/Chip/ChipList.css
cezarykubel: fixed

packages/axiom-components/src/Chip/Chip.js
```diff
@@ -0,0 +1,101 @@
...
+import Icon from '../Icon/Icon';
+import classnames from 'classnames';
+import './Chip.css';
+
+export default class Chip extends Component {
+  static propTypes = {
+    /** Apply active styling */
+    active: PropTypes.bool,
+    /** Colour of the Chip when active */
+    activeColor: PropTypes.oneOf([
```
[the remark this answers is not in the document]
cezarykubel: you'll never convince me 😄

packages/axiom-components/src/Chip/Chip.css
```diff
@@ -0,0 +1,116 @@
+.ax-chip {
+  display: flex;
+  align-items: center;
+  justify-content: space-between;
+  margin-right: var(--cmp-inline-spacing);
+  margin-bottom: var(--cmp-inline-spacing);
+  padding: 0;
+  float: left;
```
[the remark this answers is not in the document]
cezarykubel: yeeeup, not quite sure why i used floats initially.  but its fixed now

packages/axiom-components/src/Chip/ChipList.css
cezarykubel: its because we use `cubic-bezier`, specifically `cubic-bezier(0.25, 0.46, 0.45, 0.94)` for all transitions.  when the list is large, the transition is smoother on both the collapse and expand

packages/axiom-components/src/Chip/ChipList.js
cezarykubel: I don't think i've needed to use it before but its super useful, especially in this use-case when with one line of code, you can set the size of all the Chips with one prop on the container.  As far as checking the child, technically, it should accept any list of components, Chip or not, so I don't know if thats necessary.  I can be easily swayed otherwise if everyone thinks otherwise.

packages/axiom-components/src/Chip/ChipList.js
cezarykubel: hrmm...would that just cause more jumbled code, trying to find the functions relating to each bit of the render method?

packages/axiom-components/src/Chip/ChipList.js
cezarykubel: I don't think I've used getBoundingClientRect() before but it seems to get the same values as `clientHeight`.  The reasoning behind using 2x the client height is because we want to make sure that the ChipList is only expandable IF theres more than one line of chips.  The client height represents one line of chips

packages/axiom-components/src/Chip/Chip.js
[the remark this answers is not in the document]
larister: Not sure I'll ever convince myself...

packages/axiom-components/src/Chip/ChipList.css
larister: hmm interesting, thanks. Still a bit surprised it only happens on "collapse" not "expand" for the demo but nabd

packages/axiom-components/src/Chip/ChipList.js
larister: Nah I don't think it's worth it personally

packages/axiom-components/src/Chip/ChipList.js
larister: yeah maybe you're right - nabd anyway

lpoulter: I've gone back and forth on this but generally prefer the way Cezary has done it here as theres really not much logic.

packages/axiom-components/src/Chip/ChipList.js
larister: Huh, it's meant to take into account padding/margin etc AFAIK and it did seem to give a different value to `clientHeight` when I tried it.

> The reasoning behind using 2x the client height is because we want to make sure that the ChipList is only expandable IF theres more than one line of chips. The client height represents one line of chips

Sorry I'm being slow here, but in that case shouldn't it be` >=`?

Don't worry about answering if you don't want, this code works and is fine, it's just for my own curiosity really :)

cezarykubel:
made a few adjustments:  removed the colored chips and fixed the transition based on the design review.  also, fixed the disabled state for active chips.

PR #745, opened 2019-05-07
chore(React): Make use of react versions consistent 

lpoulter:
We are using two versions of React. 1 > 2.

### DIFF
--- a/packages/axiom-charts/package.json
+++ b/packages/axiom-charts/package.json
@@ -12,6 +12,6 @@
     "@brandwatch/axiom-components": "^5.21.0",
     "classnames": "^2.2.1",
     "prop-types": "^15.6.0",
-    "react": "^16.4.1"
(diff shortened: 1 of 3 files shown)


lpoulter:
> Fine by me but will it cause issues for anyone on older versions of React?

Hmm I'm not certain but if you where on an older version of React wouldn't that get installed along side 16.8.6?

larister:
Yeah that's what I'm worried about considering we've had bugs with different React versions being installed alongside each other....

When you say "We are already 16.8.6 elsewhere", does that mean all the other apps consuming Axiom? Or just Audiences?

lpoulter:
I mean other axiom packages are already using 16.8.

lpoulter:
@larister. I've done some testing with project with older versions and can't see any problems. I'm gonna merge this as I think it will fix the issues I'm seeing in audiences.

PR #747, opened 2019-05-09
feat(Dialog): Add new prop "hasOverlay" to allow no overlay to be shown.

lpoulter:
This will allow modals to be kept open when clicking outside of the Dialog.
Default of true will mean the behaviour of existing Dialogs will not change.

### DIFF
--- a/packages/axiom-components/src/Dialog/Dialog.js
+++ b/packages/axiom-components/src/Dialog/Dialog.js
@@ -9,6 +9,8 @@ export default class Dialog extends Component {
   static propTypes = {
     /** Content inside the Dialog */
     children: PropTypes.node,
+    /** Stops the dialog closing when the mask is clicked */
+    closeOnOverlayClick: PropTypes.bool,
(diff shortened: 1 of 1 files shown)


lpoulter:
That sounds better i like the prob `closeOnOverlayClick` better. I think i'd still like to pass the prob down to model thou. Might be clearer?

larister:
hmm personally I think that since the Modal's prop is `handleOverlayClick` it's pretty clear that if that prop is not specified then clicking it should do nothing

lpoulter:
@larister Ok i think I agree on reflection. thoughts on the edit?

packages/axiom-components/src/Dialog/Dialog.js
```diff
@@ -68,7 +72,8 @@ export default class Dialog extends Component {
 
     return (
       <Modal { ...rest }
-          onOverlayClick={ onRequestClose }
+          closeOnOverlayClick={ closeOnOverlayClick }
```
larister: I think this is redundant now?

PR #748, opened 2019-05-10
Textinput icon tooltip

larister:
Adds the ability to specify a tooltip for TextInputIcon.

Note that as per the commit message, I fleshed out some tests for TextInput as well

### DIFF
--- a/packages/axiom-components/src/Form/TextInputIcon.js
+++ b/packages/axiom-components/src/Form/TextInputIcon.js
@@ -1,8 +1,14 @@
 import classnames from 'classnames';
 import PropTypes from 'prop-types';
 import React, { Component } from 'react';
+import { isComponent } from '@brandwatch/axiom-utils';
 import Icon from '../Icon/Icon';
 import Link from '../Typography/Link';
+import Tooltip from '../Tooltip/Tooltip';
+import TooltipContent from '../Tooltip/TooltipContent';
+import TooltipContext from '../Tooltip/TooltipContext';
+import TooltipSource from '../Tooltip/TooltipSource';
+import TooltipTarget from '../Tooltip/TooltipTarget';
 
 export const TextInputIconRef = 'TextInputIcon';
 
@@ -16,6 +22,8 @@ export default class TextInputIcon extends Component {
     name: PropTypes.string.isRequired,
     /** Click handler, applies styling to indicate it is clickable. */
     onClick: PropTypes.func,
+    /** Optional tooltip for the icon. */
+    tooltip: PropTypes.oneOfType([PropTypes.string, PropTypes.element]),
   };
 
   static defaultProps = {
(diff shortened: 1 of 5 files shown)


packages/axiom-components/src/Form/TextInputIcon.js
```diff
@@ -35,11 +59,9 @@ export default class TextInputIcon extends Component {
         {
           onClick ? (
             <Link { ...rest } onClick={ onClick } style="subtle">
-              <Icon name={ name } size="1rem" textColor={ iconColor } />
+              { this.getIcon(name, iconColor, tooltip) }
```
larister: Decided on this route but it does mean that the tooltip is created inside the link. At an HTML level this isn't an issue, but happy to change if people think it's fugly semantically.

packages/axiom-components/src/Form/TextInputIcon.test.js
```diff
@@ -42,4 +44,17 @@ describe('TextInputIcon', () => {
...
+  it('renders with tooltip', () => {
+    props.tooltip = 'I am a tooltip';
+
+    const component = getComponent(props);
+    const tree = component.toJSON();
+
+    const wrapper = shallow(<TextInputIcon { ...props } />);
+    wrapper.simulate('mouseenter');
+
+    expect(tree).toMatchSnapshot();
```
larister: I decided to keep a snapshot test in here anyway, and have used the standard `react-test-renderer` to generate the snapshot still as enzyme uses a different format. I could perhaps use something like [this](https://www.npmjs.com/package/enzyme-to-json) but seems a bit overkill.

lpoulter:
The build is failing because of the commit message. You can run `yarn lint:commits` locally to debug.

larister:
Cheers @lpoulter 👍

lpoulter:
What's the use case for this? It feels a little close to a `TextInput` usage hint.

larister:
Yes good question @lpoulter sorry I should have included that in the description. This arose from the "Add Multiple Keywords" feature in Audiences where you can click an icon in the text box to open a dialog, allowing you to add multiple keywords at once:

image

image

@Flawwles requested that the icon has a tooltip explaining what it does. I know what you mean about the usage hint but that applies to the input as a whole; the tooltip just explains the icon.

packages/axiom-components/src/Form/TextInputIcon.js
```diff
@@ -24,8 +31,25 @@ export default class TextInputIcon extends Component {
...
 
+  getIcon(name, iconColor, tooltip) {
+    return tooltip ? (
+      <Tooltip isVisible>
+        <TooltipTarget>
+          <Icon name={ name } size="1rem" textColor={ iconColor } />
+        </TooltipTarget>
+        <TooltipSource width="auto">
+          <TooltipContext>
+            <TooltipContent size="tiny">
```
lpoulter: I'm a little torn on this. On one hand it makes it easy to use on the other it's a bit restrictive. Did you consider letting `TextInputIcon` take a child of `TooltipContent` and rendering that. Either was I'm happy with what you think is best i think i would rename `tooltip` to `tooltipText` thou.

larister: Hmm yeah good shout about the restrictiveness - I'm going to allow both if that's ok.

larister: i.e. 
```js
    tooltip: PropTypes.oneOfType([PropTypes.string, PropTypes.instanceOf(TooltipContent)]),
```

larister: Gah turns out `instanceOf` doesn't work for react elements, and there's no other way of checking that an element is of a given type unless you write a custom matcher which seems overkill. Have left it with just checking `PropTypes.element` instead

lpoulter: Cool seems fine.

packages/axiom-components/src/Form/TextInputIcon.js
```diff
@@ -24,8 +31,27 @@ export default class TextInputIcon extends Component {
...
 
+  getIcon(name, iconColor, tooltip) {
+    return tooltip ? (
+      <Tooltip isVisible>
+        <TooltipTarget>
+          <Icon name={ name } size="1rem" textColor={ iconColor } />
+        </TooltipTarget>
+        <TooltipSource width="auto">
+          <TooltipContext>
+            { typeof(tooltip) === 'string' ? (
```
lpoulter: There is a [helper](https://github.com/BrandwatchLtd/axiom-react/blob/fb3f7ecbc83e6cfd1ba7bc596705333eaac6e917/packages/axiom-utils/src/isComponent.js) method that uses [isValidElement](https://reactjs.org/docs/react-api.html#isvalidelement)

larister: Thanks, TIL. Updated...

PR #753, opened 2019-06-04
fix(DatePicker): pass on initialDate prop

tomru:
In https://github.com/BrandwatchLtd/axiom-react/pull/614 `initialDate` was missed to be passed down to the new `DatePickerSelection`.

With `initialDate` you can define which month should be shown, when a user opens the dialog, without setting any date/date range as selected. So IMO it's still useful and should be kept.

This PR fixes `initialDate` to work again.

### DIFF
--- a/packages/axiom-components/src/DatePicker/DatePicker.js
+++ b/packages/axiom-components/src/DatePicker/DatePicker.js
@@ -7,15 +7,16 @@ import DatePickerContext from './DatePickerContext';
 
 export default class DatePicker extends Component {
   static propTypes = {
+    /**
+     * The date that should be shown to the user when the picker opens.
+     * When a selected date or range is given, this supersedes this
+     * property. Defaults to 'today'.
+     */
+    calendarOpenDate: PropTypes.instanceOf(Date),
     /** The element to position the DatePicker around */
     children: PropTypes.node.isRequired,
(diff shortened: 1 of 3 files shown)


lpoulter:
Can you provide an example Tom, I can't seem to get initial date to work on this branch.

tomru:
sure @lpoulter [demo on netlify branch](https://drive.google.com/file/d/16ooNDimQuvtZFbbyu1JewTSjBnCC42I_/view?usp=sharing)

lpoulter:
Ahh I see. I don't think i noticed it was working because it does not actually appear to set the date. Don't you think it should select/highligh jan 1st in your example?

tomru:
No, as `intialDate` it's just a way to set the month when the picker opens, but has nothing to do with the selected date/range.

Otherwise I would not see a benefit in keeping it, cause you could just pass `selectedDate` or `selectedRange` to achive that.

I'm unsure if the docs might be not clear enough :/

tomru:
I think it's useful when the context where the date picker is used, makes `today` not a great initial thing to show, but rather e.g. last year (e.g. enter you income from last year)

tomru:
@lpoulter as discussed, rename to `calendarOpenDate` to make it clearer for what it should be used for.

PR #754, opened 2019-06-04
Allow size tiny on Card

Jafferwaffer:
The size tiny works fine and is causing errors in prop validation.

Example screenshot:
image

For context we use this in the controls for admin in Vizia


lpoulter:
Code looks good. Is the need for Tiny based on some designs? Do you think this PR will need a design review?

Jafferwaffer:
Thanks for the 👍 yes this is based on a design from Aliyah in our team so think we're good without a design review

Jafferwaffer:
Closing this due to design review

PR #756, opened 2019-06-05
fix(Position): Fixes the margin depending on where the arrow is shown

siggysamson:
I've improved the position of the Dropdown component when using `showArrow` and `offset`. I've noticed the dropdown is a little misplaced:
Screenshot 2019-06-05 at 10 43 34

When inspecting you can see it has a margin applied on every side:
Screenshot 2019-06-05 at 10 44 14

I've updated the CSS to only apply the margin on the side where the arrow is being shown, resulting in this:
Screenshot 2019-06-05 at 10 43 58

### DIFF
--- a/packages/axiom-components/src/Position/Position.css
+++ b/packages/axiom-components/src/Position/Position.css
@@ -3,8 +3,20 @@
   z-index: var(--z-index-position);
 }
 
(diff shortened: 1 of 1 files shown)


Flawwles:
👍 Looks good

PR #757, opened 2019-06-05
feat(DatePicker): Exporting DatePickerSelection

ogandose:
Exporting DatePickerSelection so that the DatePicker can be used without a Dropdown.

### DIFF
--- a/packages/axiom-components/src/DatePicker/DatePickerSelection.js
+++ b/packages/axiom-components/src/DatePicker/DatePickerSelection.js
@@ -4,7 +4,6 @@ import Grid from '../Grid/Grid';
 import GridCell from '../Grid/GridCell';
 import DatePickerHeaderControl from './DatePickerHeaderControl';
 import DatePickerViewMonth from './DatePickerViewMonth';
-import SelectedDateOrRangePropTypes from './SelectedDateOrRangePropTypes';
 import {
   addMonths,
   dateOrNow,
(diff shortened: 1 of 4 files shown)


packages/axiom-components/src/index.js
```diff
@@ -50,6 +50,7 @@ export { default as DataPicker } from './DataPicker/DataPicker';
 export { default as DataPickerDropdown } from './DataPicker/DataPickerDropdown';
 export { default as DataPickerMeta } from './DataPicker/DataPickerMeta';
 export { default as DatePicker } from './DatePicker/DatePicker';
+export { default as DatePickerSelection } from './DatePicker/DatePickerSelection';
```
tomru: `DatePickerSelection` seems to be the right level to export for the use-case we discussed in the WG. It's not the best naming that I chose for it back then. Now that we export it, maybe it's time for a better name... Any good ideas?

packages/axiom-components/src/DatePicker/DatePickerSelection.js
```diff
@@ -17,12 +16,46 @@ import './DatePicker.css';
 
 export default class DatePickerSelection extends Component {
   static propTypes = {
+    /**
+     * The date that should be shown to the user when the picker opens.
+     * When a selected date or range is given, this supersedes this
```
MariamBashir:
```suggestion
     * When a selected date or range is given, that supersedes this
```

packages/axiom-components/src/index.js
lpoulter: `DatePickerSelection` is ok. Other options I can think off;
*`DatePickerCalendar`
*`DatePickerSelectionCalendar`

tomru: if you are fine maybe `DatePickerSelection` is just fine. My second choice would be `DatePickerCalendar` ....

tomru: Let's leave it for now as `DatePickerSelection`...

PR #758, opened 2019-06-07
fix(InputWrapper): Apply margin to hint when label is inline

lpoulter:
Fix the label hint margin when inline. 

Preview [here](https://deploy-preview-758--serene-shannon-8198f2.netlify.com/docs/packages/axiom-components/form)

Screen Shot 2019-06-07 at 12 43 17


Flawwles:
Looks good 👍

PR #491, opened 2018-03-23
New UI Colours, Shadows and Shading pattern. 

HHogg:
Introduces a better way to handle Cards, Shadows and general variable background colours across the day and night themes. 

Notable breaking changes is mainly around Cards. 

* Configurable elevation is now not a thing. Simply use the flag `shadow` and see the Card matrix in the new documentation. 
* `color` property has been removed as the values did not work for the night theme. Use the new `shade` pattern. 

Other breaking changes that are unlikely to effect anyone updating include, some CSS variable changes and a prop removal from Modal. Please see commit messages and CHANGELOGs for more information. 

[Netlify Deploy](https://new-ui-colours--hhogg-axiom.netlify.com/)

PR #759, opened 2019-06-07
fix(color): Fix uiCarbon color

Flawwles:
When the UICarbon color was updated by Harry (#491) he only updated the CSS file and missed the JS.
This will bring the UICarbon color inline with the CSS file and also the color used across the Sketch file.

Artboard


lpoulter:
Cheers Matt. Sounds like this might be a breaking change?

lpoulter:
Discussed IRL. Breaking change likely not the best course of action. This won't break any code and we want design changes to propagate.

PR #761, opened 2019-06-12
feat(CardComponents): Add classname prop to all Card Components

bwilson-bw:
Allows a custom class to be passed to Card Components.
This means that

```javascript
<div className="dialog__sidebar">
  <CardList>...</CardList>
</div>
// is Now
<CardList className="dialog__sidebar">...</CardList>
```

### DIFF
--- a/packages/axiom-components/src/Card/Card.js
+++ b/packages/axiom-components/src/Card/Card.js
@@ -32,6 +32,8 @@ export default class Card extends Component {
     borderRadius: PropTypes.oneOf(['small', 'large']),
     /** Content to be inserted inside the Card */
     children: PropTypes.node.isRequired,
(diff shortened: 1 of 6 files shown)


roburidge:
Looks good @bwilson-bw thanks for adding this. I'm thinking as this is a common pattern we seem to be adding to components we should take the time in this PR to add the `className` prop to the other "Card" components.
- `<Card>`
- `<CardCaption>`
- `<CardContent>`
- `<CardImage>`
- `<CardImages>`

lpoulter:
Yes I think that's a good idea @roburidge

jdelight:
Agree with @roburidge / @lpoulter - lets update all Card* components with className.

PR #763, opened 2019-06-14
feat(colorUtils): added colorUtils with rgb2hex converter

tomru:
Axiom provides colors as a `{ r, g, b [,a]}` object to consumers. 

In [AX-107](https://office.brandwatch.com/jira/browse/AX-107) it was raised that consumers might  need to corresponding hex values, like e.g. the charts in analytics.

There is no need to convert `a` (alpha) as it's a number between 0 and 1, like `opacity` uses. So ignoring this in here and leave it up to the consumer to deal with alpha.

This PR adds `colorUtils#rgb2hex` so consumers don't need to do that themselves.


lpoulter:
@jdelight Does this fit your use case?

jdelight:
This looks great. Thanks.

PR #764, opened 2019-06-25
Update logo

lpoulter:


packages/axiom-components/src/Logo/Logo.css
```diff
@@ -1,3 +1,35 @@
 .ax-logo {
   max-width: 100%;
 }
+
+.cls-1 {
```
lpoulter: I'll update these to the new logo vars when and if we add them.

PR #766, opened 2019-07-09
fix(Range): range sliders crossing over on mouse up

clairebw:
The code ensuring that the handles were not crossing over was only called on mousemove but it also needs to be called on mouseup.

### DIFF
--- a/packages/axiom-components/src/Slider/Range.js
+++ b/packages/axiom-components/src/Slider/Range.js
@@ -71,6 +71,24 @@ export default class Range extends Component {
     return step * Math.round(valueInRange / step);
   }
 
+  ensureValuesDontCrossover(values) {
+    const orderedValues = [...values];
+
+    if (orderedValues[0] > orderedValues[1]) {
+      switch (this.focusedHandleIndex) {
+      case 0:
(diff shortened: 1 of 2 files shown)


packages/axiom-components/src/Slider/Range.js
```diff
@@ -71,6 +71,24 @@ export default class Range extends Component {
     return step * Math.round(valueInRange / step);
   }
 
+  ensureValuesDontCrossover(values) {
+    const orderedValues = values;
```
lpoulter: Not sure this assignment serves any purpose?

clairebw: Just wanted to make the function pure but failed 🙈 . Fixed now

packages/axiom-components/src/Slider/Range.js
```diff
@@ -71,6 +71,24 @@ export default class Range extends Component {
     return step * Math.round(valueInRange / step);
   }
 
+  ensureValuesDontCrossover(values) {
+    const orderedValues = [].concat(values);
```
tomru: NABD `[...values]` serves the same purpose, but looks a bit more elegant to me 😄

PR #771, opened 2019-07-11
fix(TextInput): respect space prop on TextInput with multiple children

tomru:
`TextInput` takes a `space` property, but does not pass it on correctly to the `InputWrapper`. E.g.
```jsx
<TextInput space="x0">
    <TextInputIcon align="left" name="magnify-glass" />
    <TextInputButton>Search</TextInputButton>
</TextInput>
```
will still use `x4` on the on an internal element (`ax-input__wrapper`).  [Example](https://codesandbox.io/embed/green-firefly-z5q8q)

The issue is not visible when you don't add multiple children to the `TextInput` like in the example as the last child does never have a bottom margin.

This PR fixes this issue.

### DIFF
--- a/packages/axiom-components/src/Form/TextInput.js
+++ b/packages/axiom-components/src/Form/TextInput.js
@@ -150,6 +150,7 @@ export default class TextInput extends Component {
                 isValid={ isValid }
                 label={ label }
                 size={ size }
+                space={ space }
                 style={ style }
                 usageHint={ usageHint }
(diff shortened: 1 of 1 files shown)


tomru:
> Wondering if perhaps it also makes sense to update propTypes for this component for space.

Thanks, I think the idea, why `PropTypes.string` was used here, is that you would still see the the prop type warning from `Base` as it just get's passed along. This way we wouldn't have to update 2 places if the space values change anytime...

PR #775, opened 2019-07-12
chore(axiom-materials): add rgb variable for all colors

eliotstocker:
they can now be used to construct an RGBA with opacity variables

### DIFF
--- a/packages/axiom-materials/src/colors.css
+++ b/packages/axiom-materials/src/colors.css
@@ -27,6 +27,8 @@
   --rgb-ui-error: 252, 72, 112;
   --rgb-ui-error--active: 240, 58, 97;
 
+  --rgb-ui-highlight: 215, 255, 0;
+
   --rgb-ui-info--hover: var(--rgb-ui-accent--hover);
   --rgb-ui-info: var(--rgb-ui-accent);
   --rgb-ui-info--active: var(--rgb-ui-accent--active);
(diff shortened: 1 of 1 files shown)


lpoulter:
Hi Elliot thanks for the PR. Are you able to provide an example use case to help my understanding?

eliotstocker:
Hey Laurie,

im currently working on slipstream UI, which previously was using hard coded copies of the axiom colors, there is a number of CSS animations etc where we require using axiom colors with a set opacity for a color.

Example:
```
@keyframes stageProgress {
  0% {
    box-shadow: 0 0 0 0 rgba(var(--rgb-product-paradise-lost), var(--opacity-input-border--focused));
  }
  70% {
    box-shadow: 0 0 5px 10px rgba(var(--rgb-product-paradise-lost), 0);
  }
  100% {
    box-shadow: 0 0 0 0 rgba(var(--rgb-product-paradise-lost), 0);
  }
}
```

this way if i just need the color i can use var(--color-product-paradise-lost),
but if i need the color with a set opacity i can then do something like: `rgba(var(--rgb-product-paradise-lost), var(--opacity-input-border--focused));`

lpoulter:
Ok good stuff that makes sense to me. 
We are unpacking
`--color-product-tiny-clanger: rgb(var(--rgb-product-tiny-clanger));`  
to 
```
--rgb-product-tiny-clanger: 244, 152, 183;
--color-product-tiny-clanger: rgb(var(--rgb-product-tiny-clanger));
```

lpoulter:
Looks like your build is failing due to our commit message lint rules. You can run `yarn lint:commits` to debug locally.

eliotstocker:
all seems to be passing locally :(

PR #776, opened 2019-07-17
chore(docs): fix Node.js link in docs

roburidge:


lpoulter:
Actually it looks like we are not.

roburidge:
`We're using the latest stable` Good point Laurie, to be accurate we're using the latest LTS version, not "latests stable"

PR #784, opened 2019-08-08
Flag color fix

stefanpearson:
It looks like Facebook's brand change messed up our flags. The two should not be related but Axiom doesn't have a navy blue in the palette, so I've had to switch to Tumblr for now. Let's hope they don't rebrand…

Ideally I would like to make the colour matching smarter and prioritise hue over the full colour value, but this is a quick and easy fix.


Flawwles:
LGTM 👍

PR #785, opened 2019-08-12
fix(animicon): Changed Cross animation

Flawwles:
The cross icon has been created with two paths, this is to enable the animicon component to animate each path separately.

Because we created the cross with two overlapping paths, it means any time we used the cross icon with opacity, you saw the overlap..

Look at overlapping parts of the cross 🕵️‍♂️
cross-before-after

This PR changes the animation of the cross, but I dont think its used much if at all?

This might not be the best solution, but hopefully we can get the ball rolling with this.

### DIFF
--- a/packages/axiom-components/src/Icon/Animicon.css
+++ b/packages/axiom-components/src/Icon/Animicon.css
@@ -21,9 +21,9 @@
 .ax-animicon--cross {
   & .ax-animicon__path-1,
   & .ax-animicon__path-2 {
-    stroke-dasharray: 14;
-    stroke-dashoffset: 14;
-    animation-duration: var(--transition-time-base);
+    stroke-dasharray: 25;
+    stroke-dashoffset: 25;
+    animation-duration: var(--transition-time-slow);
     animation-fill-mode: forwards;
     animation-timing-function: var(--transition-function);
   }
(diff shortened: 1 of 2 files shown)


lpoulter:
Just did some digging it looks like the icon was added in [this](https://github.com/BrandwatchLtd/axiom-react/pull/355) PR which makes me suspect it was added for Viza but I can't find any usage of <Animicon> in any BW repo.

Flawwles:
> Just did some digging it looks like the icon was added in [this](https://github.com/BrandwatchLtd/axiom-react/pull/355) PR which makes me suspect it was added for Viza but I can't find any usage of `<Animicon>` in any BW repo.

Thanks for checking @lpoulter.

packages/axiom-components/src/Icon/Animicon.css
```diff
@@ -21,9 +21,9 @@
 .ax-animicon--cross {
   & .ax-animicon__path-1,
   & .ax-animicon__path-2 {
```
jdelight: Do we still need `& .ax-animicon__path-2` now?

Flawwles: Good point - Guess we don't need any of the paths thinking about it.

PR #787, opened 2019-08-15
Sn 100 add responsiveness to slider

MariamBashir:
Add responsiveness to the slider in Axiom. This was creating problems for the Bubble chart component as the slider did adjust its width with the window resizing. 

Before

Screen Shot 2019-08-15 at 14 39 53

After
Screen Shot 2019-08-15 at 14 40 11


packages/axiom-components/src/Slider/Range.js
```diff
@@ -197,55 +219,82 @@ export default class Range extends Component {
...
 
+  handleResize() {
+    const track = this.trackRef.current;
+
+    if (!track){
+      return;
+    }
+
+    const { width: trackWidth } = track.getBoundingClientRect();
+    this.setState({ trackWidth });
```
clairebw: When we move a component to have the dashboard split/unsplit the Customize Bar is rerendered and so is the slider. We end up having a new `trackWidth` after the render with no window 'resize' event being triggered. 

Is there a way to handle this? or is this case already addressed? I'm not sure.

lilybrandwatch:
@MariamBashir It's hard to judge with the screenshots only but in your "After", the inputs on the left needs to be resized too which will allowed more space to the slider like it in my design but I guess this is because you were only focusing on the sliders . For the rest if the handle size works instead of the input I'm fine with this as solution but I'd like to test it first of course. 
Thanks. :)

packages/axiom-components/src/Slider/Range.js
lpoulter: Not sure I fully understand @clairebw 's use case but it sounds like you should be able to use [ResizeObserver](https://developer.mozilla.org/en-US/docs/Web/API/ResizeObserver). This will trigger when the element changes size, it's also more efficient as it does not trigger on every window resize. You will need to use the polyfill for edge support.

packages/axiom-components/src/Slider/Range.js
```diff
@@ -133,6 +150,11 @@ export default class Range extends Component {
     }, 0);
   }
 
+  componentDidMount(){
+    this.handleResize();
+    window.addEventListener('resize', this.handleResize);
```
lpoulter: Not sure this Listener is removed, but you might want to use ResizeObserver anyway, see below.

packages/axiom-components/src/Slider/Range.js
```diff
@@ -197,55 +219,82 @@ export default class Range extends Component {
...
+          <Text>{ minLabel }</Text>
+        </GridCell>
+        <GridCell width={ 70 }>
+          <Base { ...omit(rest, ['onSlideEnd', 'min', 'max', 'step']) }
+              className={ classes }
+              onBlur={ this.handleBlur }
+              onFocus={ disabled ? null : this.handleFocus }
+              onMouseLeave={ () => this.setState({ isMouseOver: false }) }
+              onMouseOver={ () => this.setState({ isMouseOver: true }) }
+              style={ { maxWidth: MAX_SLIDER_WIDTH } }
```
lpoulter: This max width seems to be causing issues when it's parent, in this case a `GridCell`, is larger than this max width. A gap is created between the label and the track.

lpoulter: Screen Shot 2019-08-28 at 21 05 25

MariamBashir: Interesting, ill dig into that.

MariamBashir:
@lpoulter I will do some research on the `ResizeObserver` and replace the window resize event with that. Thank you for the feedback.

@lilybrandwatch I will implement the changes and then set up a meeting with you again so you can test it with me and we can take it from there. Thanks for taking a look.

MariamBashir:
@lpoulter , I have added the resizeObserver which also covers the case @clairebw mentioned. Also removed the `MAX_WIDTH` which was causing the min label to appear away from the slider track. 

@lilybrandwatch as discussed yesterday with you, i will create a new ticket for resizing the drop downs in the customize bar.  This ticket focused on the sliders and makes its responsive based on the width of the customize bar. I have created another [ticket](https://office.brandwatch.com/jira/browse/SN-192) to add responsiveness to the drop downs based on the width of customize bar.

PR #789, opened 2019-08-28
feat(Dropdown/Select): Changed selected color to grey

Flawwles:
Small change to the hover colour of the Dropdown & Select boxes.

Before-after

### DIFF
--- a/packages/axiom-components/src/Context/Context.css
+++ b/packages/axiom-components/src/Context/Context.css
@@ -169,8 +169,7 @@
 
 .ax-context-menu__item:focus {
   outline: 0;
-  background-color: var(--color-ui-accent--hover);
-  color: var(--color-ui-white-noise);
+  background-color: var(--color-theme-background--active);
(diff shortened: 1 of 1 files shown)


packages/axiom-components/src/Context/Context.css
```diff
@@ -169,8 +169,7 @@
 
 .ax-context-menu__item:focus {
   outline: 0;
-  background-color: var(--color-ui-accent--hover);
-  color: var(--color-ui-white-noise);
+  background-color: var(--color-theme-background--active);
```
jdelight: @Flawwles Only issue for me here is semantically should we be changing the colour associated with color-ui-accent--hover? i.e. is there now an inconsistency with other parts of the UI that still use color-ui-accent--hover. 
Also kind of makes sense that this dropdowns are using the background active colour but I wouldn't naturally associate color-theme-background--active with dropdowns. Hope that makes sense :)

PR #794, opened 2019-09-18
MYB-2 update user avatar

tommillward:
This PR does the following:

- Updates the Candytar component to accept and show the User's initials. If no initials are supplied then it shows the silhouette by default.
Screen Shot 2019-09-18 at 10 01 01

- Updates the UserMenu component to use the initials prop for the Candytar component so the User's initials are shown.
Screen Shot 2019-09-18 at 10 00 04

- Adds the initials prop to two of the Candytar components in the Avatar documentation example.

### DIFF
--- a/packages/axiom-components/src/Avatar/Candytar.js
+++ b/packages/axiom-components/src/Avatar/Candytar.js
@@ -27,6 +27,11 @@ export default class Candytar extends Component {
       'luna-dust',
     ]),
     /**
+     * The User's initials to appear in the candytar.
+     * If no initials are supplied the silhouette will be shown by default.
+     */
+    initials: PropTypes.string,
+    /**
      * A function that can be passed to determine the colour of the candytar.
      * By default if no colour is provided it will randomly picker a colour
      * from the available options. Given function is called with an array of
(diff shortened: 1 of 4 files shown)


packages/axiom-components/src/Avatar/Candytar.js
```diff
@@ -39,26 +44,36 @@ export default class Candytar extends Component {
 
   static defaultProps = {
     picker: defaultPickerFn,
+    initials: '',
   };
 
   UNSAFE_componentWillMount = renderFilter
```
lpoulter: Would be good to replace this with `componentDidMount` or calling renderFilter in a constructor whilst you are here.

packages/axiom-components/src/Avatar/Candytar.js
```diff
@@ -47,7 +47,7 @@ export default class Candytar extends Component {
     initials: '',
   };
 
-  UNSAFE_componentWillMount = renderFilter
+  componentDidMount = renderFilter
```
lpoulter: 🍠

Flawwles:
Looks good to me 🎨

PR #798, opened 2019-10-10
feat(Dialog): Adds new large and fullscreen options.

lpoulter:
Add size prop with 2 options to allow dialog to cover w…hole screen and cover with padding.

BREAKING CHANGE: Usage of the fullscreen property on the Dialog component will now have no effect. You can recreate the previous fullscreen mode by passing `size=large` e.g.
```
<Dialog size="large" />
```
Size fullscreen will remove the padding previously found in fullscreen mode.

large size
Screen Shot 2019-10-10 at 10 05 48

fullscreen size
Screen Shot 2019-10-10 at 10 05 27

### DIFF
--- a/packages/axiom-components/src/Dialog/Dialog.css
+++ b/packages/axiom-components/src/Dialog/Dialog.css
@@ -36,7 +36,12 @@
   flex: 0 1 auto;
 }
 
-.ax-dialog--fullscreen {
+.ax-dialog--fullscreen,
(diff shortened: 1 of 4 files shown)


Flawwles:
Nice, this is looking good.

I was wondering if the dialog is in fullscreen mode should we remove the border radius, as it just looks a bit broken at the moment. 

When the dialog is fullscreen - 
image

PR #800, opened 2019-10-15
fix(colors): fix white noise color in color.js

tomru:
The values for the color `white-noise` in `color.js` are not inline with [`color.css`](https://github.com/BrandwatchLtd/axiom-react/blob/7dc34bf730dec0a71c668bda3689ccc07c541f47/packages/axiom-materials/src/colors.css#L7-L9)

This PR aligns them.


Flawwles:
LGTM - Thanks for fixing this so quickly too!

PR #815, opened 2019-11-18
fix(popper): downgrade to fix browser crashing

TomWija:


lpoulter:
Merging this for now to fix a critical production issue. Looks like the root of the problem might be a popper memory leak. More [here](https://github.com/reactstrap/reactstrap/pull/1669/files).

PR #816, opened 2019-11-18
Popper fix

lpoulter:
Breaking Change.

A recent upgrade of `Popper.js` introduced a [bug](814) which created an infinite loop, crashing the browser. This was caused by and highlighted a flaw in our implementation, we had two sources of truth for `placement` local state and Popper.js. 

Position's local state was needed to update our component Arrow when the popper changed it's orientation as a user scrolled or the window size changed. This state was updated by hooking into Poppers.js callback. Attempting to remove this local state caused a few difficulties and would mean we would have to continue to write more imperative code to manipulate the arrow DOM.

This proposed solution replaces `Popper.js` with [react-popper](https://github.com/FezVrasta/react-popper). `react-poppper` provides a declarative API wrapper around poppper. This allows us to hook into the `placement` change with eas, using the provided render prop. Another benefit of this change is that we can delete our code that bridge the gap between DOM and React.

Removes the `onPositionChange` which I can see [no use](https://github.com/search?q=org%3ABrandwatchLtd+onPositionChange&type=Code) of in all of Brandwatch.

Things that are broke
* DataPicker - input/base ref issue
* usageHint - misaligned
* Tip - arrow styles?

### DIFF
--- a/packages/axiom-components/package.json
+++ b/packages/axiom-components/package.json
@@ -16,10 +16,10 @@
     "element-closest": "^2.0.2",
     "lodash.omit": "^4.5.0",
     "luxon": "^1.12.0",
-    "popper.js": "1.12.5",
     "prop-types": "^15.6.0",
     "react": "^16.8.6",
     "react-dom": "^16.8.6",
+    "react-popper": "^1.3.6",
     "react-transition-group": "^2.3.0",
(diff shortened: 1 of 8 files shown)


lpoulter:
Closing this for now as popper 2 is out and popper React is to be upgraded also.

PR #819, opened 2019-12-03
feat(Table): added ability to specify space prop

TomWija:


packages/axiom-components/src/Table/Table.js
```diff
@@ -7,13 +7,18 @@ export default class Table extends Component {
...
+    space: PropTypes.string,
   };
 
+  static defaultProps = {
+    space: 'x6',
+  }
+
   render() {
-    const { children, ...rest } = this.props;
+    const { children, space, ...rest } = this.props;
```
MariamBashir: NABD but wondering if we need to destructure `space` from the rest of the props, we could potentially just use `...rest` for that.

TomWija: It's more a documentation thing to ensure that people are aware that they are able to make this change. I'm not a fan of our current obfuscation of base props because no one can tell that they actually work. Happy to change it though

MariamBashir: Lets keep it :) I like the visibility of it too.

PR #826, opened 2020-01-30
⚙️ Improve Axiom build size

lpoulter:
Currently Axiom [ships as commonJs](https://github.com/BrandwatchLtd/axiom-react/blob/master/.babelrc#L44) file format. This makes it impossible for Axiom consumers to tree shake unused parts of Axiom with Webpack. 

This PR is the result of an exploration comparing 3 different bundle options, webpack/rollup and no bundle (just babel). It's [not possible](https://github.com/webpack/webpack/issues/2933) to bundle to Es6 modules with webpack. Rollup and no bundle produced simular file sizes but as we don't need all of the features of Rollup it seemed unnecessary to use it, it might still be something we add in the future.

Key Changes.
* Enables Axiom end users to tree shake Axiom which reduces initial Axiom weight from 1.22Mb to 143.kb.
* Drops ie11 support.
* Updates build dependencies, Babel 7, post-css-env, Jest.

Before
Screen Shot 2020-02-04 at 18 01 15

After
Screen Shot 2020-02-04 at 18 08 31


larister:
Wow nice one Laurie, very exciting!

axiom-playground/package.json
```diff
@@ -0,0 +1,36 @@
...
+    "@babel/preset-react": "^7.0.0",
+    "babel-loader": "^8.0.6",
+    "css-loader": "^3.2.0",
+    "html-loader": "^0.5.5",
+    "html-webpack-plugin": "^3.2.0",
+    "prettier": "^1.18.2",
+    "prop-types": "^15.7.2",
+    "style-loader": "^1.0.0",
+    "webpack": "^4.39.1",
+    "webpack-bundle-analyzer": "^3.6.0",
```
lpoulter: I didn't use Create React App for the playground as it's not compatible with `webpack-bundle-analyzer`

andrewru: In explore-mode I've been using; https://create-react-app.dev/docs/analyzing-the-bundle-size/, does that have shortcomings in the context of Axiom?

lpoulter: I've not used `source-map-explorer` I did see you could use it with CRE. I wanted to use `webpack-bundle-analyzer` mostly because I knew had used it before. I disliked being restricted in which tool I could use by CRE and for this example app I wanted to have full and transparent control over the webpack config so I could be certain what was happening with the build.

andrewru: Yeah, if you can't accept being constrained then I wouldn't use create-react-app.

webpack.config.js
```diff
@@ -1,56 +1,56 @@
-const webpack = require('webpack');
-const HtmlWebpackPlugin = require('html-webpack-plugin');
+const webpack = require("webpack");
```
stefanbuck: 💄I don't care much about single vs double quotes, but according to https://github.com/BrandwatchLtd/axiom-react/blob/master/packages/eslint-config-axiom/src/eslintrc.json#L33 it should be single quotes. Looks like files in the root are not take into account

scripts/build_flags.sh
```diff
@@ -1,4 +1,4 @@
 #!/bin/bash
 set -e
 
-npx babel-node ./scripts/process_flags --presets env
+npx babel-node ./scripts/process_flags --presets @babel/preset-env
```
stefanbuck: Not related to this PR, but from a quick glance at `./scripts/process_flags` I don't see a need for `babel-node`. Node.js 10 support quite a few ES2015-2017 features so we might be able to remove it.

babel.config.js
```diff
@@ -1,8 +1,6 @@
 module.exports = {
   plugins: [
     '@babel/plugin-proposal-class-properties',
-    '@babel/plugin-transform-classes',
-    '@babel/plugin-proposal-object-rest-spread',
```
stefanbuck: 😍 fewer dependencies

babel.config.js
```diff
@@ -14,7 +14,6 @@ module.exports = {
             'last 2 Chrome versions',
             'last 2 Edge versions',
             'last 2 Safari versions',
-            'IE 11',
```
stefanbuck: Just out of interested, does this reduces the build size too? Probably just a small percentage.

babel.config.js
```diff
@@ -37,11 +37,7 @@ module.exports = {
       plugins: [
         '@brandwatch/babel-plugin-transform-svg-axiom',
         '@brandwatch/babel-plugin-axiom-imports',
-        '@babel/plugin-transform-modules-commonjs',
```
stefanbuck: I thought webpack is compiling to cjs by default, at least this was the case in the past. So does this mean that webpack 4 compiles to ES6 modules by default?

babel.config.js
```diff
@@ -1,7 +1,5 @@
 module.exports = {
-  plugins: [
-    '@babel/plugin-proposal-class-properties',
-  ],
+  plugins: ['@babel/plugin-proposal-class-properties'],
```
stefanbuck: Again out of curiosity, do you know how much this change affected the build size?

packages/axiom-components/package.json
```diff
@@ -3,6 +3,10 @@
   "version": "8.7.1",
   "description": "Brandwatch Axiom - Components",
   "main": "dist/index.js",
+  "sideEffects": [
+    "./src/Base/Base.js",
+    "./**/*.css"
+  ],
```
stefanbuck: Small change, huge impact 🏇

webpack.config.js
lpoulter: Yeah this was my bad I had prettier running by mistake. The project doesn't use Prettier, yet....

babel.config.js
lpoulter: Yes it does. I recorded the sizes as I went.
```
Base
1.22Mb
Babel 7
1.43Mb
Remove ie11 support
1.11MB
Remove ie grid post css support
1.07MB
Tree Shake
Side effect false in components
143.83kb
```

babel.config.js
lpoulter: Hmm I'm not sure I follow. For the dist files we serve there is no webpack config, just babel. But you are right webpack does not compile to ES6.

babel.config.js
lpoulter: I didn't record it but from memory it was a decent amount ~100KB.

babel.config.js
stefanbuck: 🤦‍♂ I had not payed attention to the filename. I thought I'm looking at webpack.config.js 😆

PR #828, opened 2020-02-04
feat(playground): Basic Axiom dev playground

lpoulter:
As part of the build work I added a playground to dev in. It's not perfect and likely to be replaced by storybook but I thought it useful for now.


lpoulter:
Storybook will supersede this.

PR #829, opened 2020-02-05
fix: updates luxon to fix chrome 80 bug

stefanpearson:
In Chrome 80, there was a bug (either in Chrome or Luxon) that caused constructed dates to move forwards by 24 hours which was causing big issues for Vizia. We've updated on our side of things but here's a safety fix.

https://github.com/moment/luxon/issues/619


lpoulter:
Looks like a test needs a quick update.

PR #830, opened 2020-02-06
feat(Classnames): Add classNames to AlertIcon & Link

lpoulter:


packages/axiom-components/src/AlertIcon/AlertIcon.js
```diff
@@ -20,6 +20,8 @@ const SIZE_MAP = {
 
 export default class AlertIcon extends Component {
   static propTypes = {
+    /** SKIP */
```
TomWija: Do we definitely want to skip this out of the documentation? I normally flip a coin to know whether classname is supported so knowing for certain that it is would be nice

lpoulter: Good question. I did this last week and have forgotten why I skipped it, I thought perhaps it wasn't working in the docs but it is. I think we should include if in the Docs especially as not all components support classname yet.

PR #833, opened 2020-02-10
build: update build to generate both esm and cjs

eliotstocker:
also update package jsons to point modules at esm build


lpoulter:
Replicated in #834

PR #835, opened 2020-02-15
fix: PostCSS was not being run on build files.

lpoulter:
Glob misses some files when not in a string. Think its a postcss cli bug 🤷


lpoulter:
Going to merge this as it fixes an issue in the @latest build.

PR #840, opened 2020-02-21
Move parts of globals.css into component specific CSS

siggysamson:
I came across an issue where the CSS from the `globals.css` of Axiom would _leak_ into the rest of the app (if that app isn't fully axiom-based).
As the `ax-theme--[day|night]` CSS class is _required_  anyways to get Axiom to apply styles correctly I figured we could just scope the `globals.css` by that. This works for `html`, `body` and `*` selectors but unfortunately it increases the [CSS Specificity](https://developer.mozilla.org/en-US/docs/Web/CSS/Specificity) of the selectors and results in the styles for elements like `button`, `input` and `textarea` to not be applied (technically they are applied, they are just _overwritten_).

I ended up removing some CSS from the `globals.css` and add it to their respective components. This is obviously duplicating some CSS but I think this is way more explicity anyways.

### DIFF
--- a/packages/axiom-components/src/Form/TextInput.css
+++ b/packages/axiom-components/src/Form/TextInput.css
@@ -1,4 +1,8 @@
 .ax-input {
+  /* reset */
+  overflow: visible; /*  Show the overflow in Edge. */
+  font: inherit;
+
   flex: 1 1 auto;
   border: 0;
   background-color: transparent;
--- a/packages/axiom-components/src/Base/globals.css
+++ b/packages/axiom-components/src/Base/globals.css
@@ -30,12 +30,3 @@ body[class*="ax-theme"],
   text-rendering: optimizeLegibility;
   line-height: var(--line-height-body);
 }
-
-textarea {
-  margin: 0; /* Remove the margin in Firefox and Safari. */
-  font: inherit;
-}
-
-textarea {
-  overflow: auto; /* Remove the default vertical scrollbar in IE. */
-}
--- a/packages/axiom-components/src/Base/globals.css
+++ b/packages/axiom-components/src/Base/globals.css
@@ -1,24 +1,26 @@
 /* stylelint-disable property-no-vendor-prefix, selector-max-type */
 /* stylelint-disable selector-max-universal */
-*,
-*::before,
-*::after {
+[class*="ax-theme"]*,
+[class*="ax-theme"]*::before,
+[class*="ax-theme"]*::after {
   box-sizing: inherit;
 }
 /* stylelint-enable selector-max-universal */
 
-html,
-body {
+html[class*="ax-theme"],
+body[class*="ax-theme"],
+[class*="ax-theme"] body {
   box-sizing: border-box;
 }
 
-html {
+html[class*="ax-theme"] {
   font-size: 100%;
   -ms-text-size-adjust: 100%;
   -webkit-text-size-adjust: 100%;
 }
 
-body {
+body[class*="ax-theme"],
+[class*="ax-theme"] body {
   margin: 0;
   font-family: var(--font-family-body);
   font-size: var(--font-size-body);
--- a/packages/axiom-components/src/Base/globals.css
+++ b/packages/axiom-components/src/Base/globals.css
@@ -31,7 +31,6 @@ body[class*="ax-theme"],
   line-height: var(--line-height-body);
 }
 
-button,
 input {
   overflow: visible; /*  Show the overflow in Edge. */
 }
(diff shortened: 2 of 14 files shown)


packages/axiom-components/src/Form/TextInput.css
```diff
@@ -1,4 +1,8 @@
 .ax-input {
+  /* reset */
+  overflow: visible; /*  Show the overflow in Edge. */
+  font: inherit;
+
```
lpoulter: Does this need `margin: 0` also?

lpoulter:
Now I'm seeing this I do have some concerns. 
I think this makes it a requirement to have `ax-theme` on the <html> tag as opposed to any part of the dom? I'm not sure how much of an issue in practice but there might be scenarios where you don't readily have access to the body tag. There may also be current implementations that add the theme to other part of the DOM, I know i've made a lot of sandboxes that use the theme tag on a div. Also [this example ](https://github.com/BrandwatchLtd/axiom-react/pull/841/files#diff-41e445b678933963d77632d4206c90b8R4)from my storybook pr.

However I think my main concerns is that avoiding global styles does feel like we are fighting against how CSS is intended to work. I mostly worry that this might be error prone espically when adding new components that feature any of the HTML elements targeted, `button/hr/input/textarea`.  

I'll have a think about other approaches, did you explore any? Two things that might help would be to document the need for manual resets when adding any of the affected html tags or increasing the specificity of the component specific overrides.

siggysamson:
Yeah I keep going back and forth on it and after further adjusting it I'm less and less happy about the solution actually.

This change would require a `ax-theme--[day|night]` class to be available on a DOM element that is a parent of where Axiom is going to be used. It doesn't necessarily have to be `body` or `html`. In BCR this has been working without having the theme name on `body` or `html`.
I don't think it would break the use you are describing. It would rather support that.

I'm thinking my problem might be very specific and I should use a work-around where I'm consuming Axiom as I don't see a way to confidently solve this in Axiom.

> Document the need for manual resets when adding any of the affected html tags.

Looking at what we are trying to "reset" globally my feeling is that when creating a new component we'd "solve" browser discrepancies specific to a certain element anyways. It looks like some of the fixes in `globals.css` are intended to fix issues in specific browsers (e.g. Edge).

> Increasing the specificity of the component specific overrides.

I was indeed considering that for a moment but I'm afraid this is not a "long-term" solution either

> Making the resets opt-in, you have to import them into the target project.

Currently, in order for Axiom to look as it's supposed to, you *must* include `globals.css`. 

-------

Thinking about it further I think one issue is that `globals.css` tries to solve to cover two cases at once:
1) Apply styles to **all** elements as some kind of "base styling". E.g. `box-sizing` or `font-family`
2) Fix browser discrepancies. E.g. "Remove the margin in Firefox and Safari."

Unfortunately what happened is that other styles expect the reset (2)) to be applied in order for the other style definitions to work. So there's actually a dependency between them.

Generally "resetting" CSS has always been a bigger back and forth in terms of how/where this should be applied. In a way my feeling is that for a Component Library there shouldn't be any "global" styles, especially if the selectors are written in a very generic way. Unfortunately I don't see a good way in Axiom to solve this though :/

lpoulter:
>This change would require a ax-theme--[day|night] class to be available on a DOM element that is a parent of where Axiom is going to be used. It doesn't necessarily have to be body or html. In BCR this has been working without having the theme name on body or html.
I don't think it would break the use you are describing. It would rather support that.

Ok good stuff.

>Increasing the specificity of the component specific overrides.
>>I was indeed considering that for a moment but I'm afraid this is not a "long-term" solution either

Yeah it's not ideal but maybe it's clear or rather easier for a future person to add a double class than to know which styles to reset. Alhou I kinda dislike like my suggestions more now i've thought about it.

>In a way my feeling is that for a Component Library there shouldn't be any "global" styles, especially if the selectors are written in a very generic way.

I've seen other designs systems do this. So perhaps there is some inspiration there.

>I'm thinking my problem might be very specific and I should use a work-around where I'm consuming Axiom as I don't see a way to confidently solve this in Axiom.

Do you see this as a problem specific to BCR? It seemed like it would effect any app that wanted to mix Axiom with other sytles?

>Making the resets opt-in, you have to import them into the target project.
>>Currently, in order for Axiom to look as it's supposed to, you must include globals.css.

I was thinking that perhaps BCR could provide it's own "global.css" excluding the bits it that conflicted. I'm not sure if this would work nor if it's nicer than the [postCSS option](https://github.com/BrandwatchLtd/frontend/blob/master/reactApps/postcss.config.js#L22) you have now.

lpoulter:
Some example of an opt-in reset.
[material-ui](https://material-ui.com/components/css-baseline/)
[Braid](https://github.com/seek-oss/braid-design-system/blob/ff6409b7aa2f42742acd88b07ff62bfbc89a2271/README.md)

siggysamson:
Closing this as it doesn't seem worth it because we only have one application that is affected by this so we should use workarounds in there. Thanks!

PR #841, opened 2020-02-21
Storybook

lpoulter:
This PR puts in place the minimal set-up needed to document and develop components in storybook.


package.json
```diff
@@ -15,7 +15,9 @@
     "publish:site": "bash ./scripts/publish_site.sh",
     "start": "bash ./scripts/start_development.sh",
     "postinstall": "lerna run prepublish && lerna run prepare",
-    "test": "TZ=utc BABEL_ENV=test jest ./packages --config .jest.json"
+    "test": "TZ=utc BABEL_ENV=test jest ./packages --config .jest.json",
+    "storybook": "bash ./scripts/start_storybook.sh",
+    "build-storybook": "build-storybook"
```
Binarytales: Could we make this `build:storybook` to be consistent with the other static build command?

Binarytales: Also I think we should add the resulting `storybook-static` folder to `.gitignore`

packages/axiomDocs.stories.mdx
```diff
@@ -0,0 +1,3 @@
+<Meta title="Axiom | Intro" />
+
+# Axiom is Great
```
Binarytales: It really is.

package.json
```diff
@@ -78,5 +82,6 @@
     "validate-commit": "^3.4.0",
     "webpack": "^3.6.0",
     "webpack-dev-server": "^2.9.1"
-  }
+  },
+  "dependencies": {}
```
Binarytales: Looks like some `yarn` shenanigans probably added this. Left over after forgetting to do a `-D`? :stuck_out_tongue_closed_eyes:

scripts/start_storybook.sh
```diff
@@ -0,0 +1,6 @@
+#!/bin/bash
+set -e
+
+npx lerna exec --parallel -- "rm -rf dist && mkdir dist && ln -s ../src dist/cjs"
+npm run build:flags
+yarn start-storybook -p 6006
```
Binarytales: Missing newline?

scripts/start_storybook.sh
```diff
@@ -0,0 +1,6 @@
+#!/bin/bash
+set -e
+
+npx lerna exec --parallel -- "rm -rf dist && mkdir dist && ln -s ../src dist/cjs"
+npm run build:flags
```
Binarytales: Do you want to add `npm run build:packages` here also to ensure that it's run? Is there a way you can do it only if it hasn't already been run? Or does line 4 mean that's not needed?

package.json
lpoulter: Hmm not sure about the name change. This is the default on given by the storybook cli, so it's what you'd expect if you have used storybook before. I guess it's better to be consistent with our repo thou.

scripts/start_storybook.sh
lpoulter: > Or does line 4 mean that's not needed?
This.

PR #846, opened 2020-03-03
Platform context

lpoulter:
Removes legacy Context from `Platform` and children components also took the opportunity to add an example story.

### DIFF
--- a/packages/axiom-components/src/Platform/Console.js
+++ b/packages/axiom-components/src/Platform/Console.js
@@ -1,50 +1,43 @@
 import PropTypes from 'prop-types';
-import React, { Component } from 'react';
+import React, { useContext } from 'react';
 import classnames from 'classnames';
 import Base from '../Base/Base';
+import ConsoleContext from './ConsoleContext';
+import PlaformContext from './PlatformContext';
 
-export default class Console extends Component {
-  static propTypes = {
-    children: PropTypes.node,
-    position: PropTypes.oneOf(['left', 'right']).isRequired,
-    shade: PropTypes.oneOf(['shade-2', 'shade-3', 'shade-4']),
-    width: PropTypes.string.isRequired,
-  };
+export default function Console({
+  children,
+  position,
+  shade = 'shade-3',
+  width,
+  ...rest
+}) {
+  const { openConsolePosition } = useContext(PlaformContext);
+  const style = { width };
+  const classes = classnames(
+    'ax-platform__console',
+    `ax-platform__console--${position}`,
+    `ax-platform__console--${shade}`,
+    {
+      'ax-platform__console--open': openConsolePosition === position,
+    }
+  );
 
-  static defaultProps = {
-    shade: 'shade-3',
-  };
-
-  static contextTypes = {
-    openConsolePosition: PropTypes.string,
-  };
-
-  static childContextTypes = {
-    consolePosition: PropTypes.string,
-  };
-
-  getChildContext() {
-    return {
-      consolePosition: this.props.position,
-    };
-  }
-
-  render() {
-    const { children, position, shade, width, ...rest } = this.props;
-    const { openConsolePosition } = this.context;
-    const style = { width };
-    const classes = classnames(
-      'ax-platform__console',
-      `ax-platform__console--${position}`,
-      `ax-platform__console--${shade}`, {
-        'ax-platform__console--open': openConsolePosition === position,
-      }
-    );
-
-    return (
+  return (
+    <ConsoleContext.Provider
+        value={ {
+          consolePosition: position,
+        } }>
       <Base { ...rest } className={ classes } style={ style }>
         { children }
       </Base>
-    );
-  }
+    </ConsoleContext.Provider>
+  );
 }
+
+Console.propTypes = {
+  children: PropTypes.node,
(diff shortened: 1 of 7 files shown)


packages/axiom-components/src/Platform/Platform.stories.js
```diff
@@ -0,0 +1,122 @@
...
+      );
+    }
+  }
+
+  return (
+    <Platform
+        onConsoleClose={ closeConsole }
+        openConsolePosition={ openConsolePosition }
+        openConsoleWidth="26.25rem"
+        platform-context>
```
larister: Where's `platform-context` defined/used?

packages/axiom-components/src/Platform/ConsoleContext.js
```diff
@@ -0,0 +1,5 @@
+import React from 'react';
+
+const ConsoleContext = React.createContext();
+
+export default ConsoleContext;
```
larister: Could we add these new contexts to `/src/index.js`? In BCR I'm not using Platform but I am using ConsoleHeader, so I'll need to include the ConsoleContext provider manually. This wasn't a problem with the old Context API as you just needed to specify whatever in `getChildContext`

packages/axiom-components/src/Platform/ConsoleHeader.js
```diff
@@ -1,65 +1,60 @@
...
-      'ax-platform__console-header',
-      `ax-platform__console-header--${shade}`,
-      `ax-platform__console-header--${size}`, {
-        'ax-platform__console-header--separator': separator,
-      },
-    );
+  return (
+    <Base { ...rest } className={ classes }>
+      <Grid gutters="small" responsive={ false } verticalAlign="middle">
+        { onConsoleClose && size === 'large' && consolePosition === 'right' && (
```
larister: I'd take this opportunity to extract these out into `const showCloseOn[Left|Right]` but up to you

packages/axiom-components/src/Platform/ConsoleHeader.js
```diff
@@ -1,65 +1,60 @@
...
+          <GridCell shrink>
+            <Link onClick={ onConsoleClose } style="subtle">
+              <Icon name="cross" size="1.5rem" />
+            </Link>
+          </GridCell>
```
larister: nabd but this is exactly the same as below, perhaps extract into a `CloseCell` var or something? Existing issue I know...

packages/axiom-components/src/Platform/Platform.stories.js
lpoulter: ha this is a typo. 🍝

packages/axiom-components/src/Platform/Console.js
```diff
@@ -1,50 +1,43 @@
...
-        'ax-platform__console--open': openConsolePosition === position,
-      }
-    );
-
-    return (
+  return (
+    <ConsoleContext.Provider
+        value={ {
+          consolePosition: position,
+        } }>
```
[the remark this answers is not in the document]
lpoulter: Good catch. I set up a codesandbox [example](https://codesandbox.io/s/context-unnecessary-render-59ps5) to explore this more. Looks like this is a problem when a child uses React.Memo to save renders off an expensive operation. I've used React.useMemo to stop the object being newly created on every render of Console.

packages/axiom-components/src/Platform/ConsoleContext.js
lpoulter: To check I understand. You are just rendering `ConsoleHeader` none of the other Platform components? As I understand it you only needed to provide the `consolePosition` to avoid the warning about it not being provided? I think you can just render ConsoleHeader without a context and I can just mock out `PlatformContext` to prevent errors. I guess you could also provide a mock context in fe but that sounds worse.

lpoulter: I'll push a commit with a story that I think covers your usecase and the mocked Context.

larister: Cool thanks, yeah that seems better to me :)

packages/axiom-components/src/Platform/Platform.js
```diff
@@ -20,13 +20,15 @@ export default function Platform({
     'ax-platform--console-open': openConsolePosition,
   });
 
+  const providerValue = useMemo(() => ({
+    onConsoleClose,
+    openConsolePosition,
+    openConsoleWidth,
+  } ), [onConsoleClose, openConsolePosition, openConsoleWidth]);
```
larister: Neat trick. Is this an established pattern?

lpoulter: Most of the example I can find on storing an object in a Provider use the state of the containing component, note they all pre-date the release of useMemo, in this case Platform. Another common approach is to define the Provider object out of function declaration, not possible here because it depends on props.   
I think using useMemo() is a good solution as it's very explicit that we are  cacheing the object.

larister: yeah nice one, I'm happy with this anyway

larister:
Lol I didn't make you do anything (last commit message)! Just suggestions ;) Nice one anyway, LGTM

PR #848, opened 2020-03-18
Tooltip add Storybook / remove context

lpoulter:
Added Tooltip to Storybook.
Removed React's Legacy Context API from Tooltip.

### DIFF
--- a/packages/axiom-components/src/Tooltip/Tooltip.js
+++ b/packages/axiom-components/src/Tooltip/Tooltip.js
@@ -1,5 +1,5 @@
 import PropTypes from 'prop-types';
-import React, { Component } from 'react';
+import React, { useState, cloneElement } from 'react';
 import { findComponent } from '@brandwatch/axiom-utils';
 import Position from '../Position/Position';
 import PositionSource from '../Position/PositionSource';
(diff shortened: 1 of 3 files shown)


packages/axiom-components/src/Tooltip/Tooltip.js
```diff
@@ -1,86 +1,73 @@
...
-
-  show() {
-    const { enabled } = this.props;
-    const { isVisible } = this.state;
+export default function Tooltip({
+  children,
+  delay = false,
+  onClick,
+  position = 'top',
+  enabled = true,
```
SeaOfTea: So changing the enabled prop will destroy the popper right? So you can theoretically declaratively close a pop up?

packages/axiom-components/src/Tooltip/Tooltip.js
```diff
@@ -1,86 +1,73 @@
...
+  ...rest
+}) {
+  const [isVisible, setIsVisible] = useState(false);
 
+  function showTooltip() {
```
SeaOfTea: Dunno if it's worth adding a `useImperativeHandle` for these methods.

That would look like this:
```
function Tooltip(props, ref) {
    React.useImperativeHandle(ref, () => {
        showTooltip,
        hideTooltip
    });

    function showTooltip() {

    }

    function hideTooltip() {

    }
}

export default React.forwardRef(Tooltip)
```

Because then one day if someone wanted to close the tooltip they could theoretically do this.

```
<Button onClick={() => tooltipRef.current.hideTooltip()}/>
<Tooltip ref={tooltipRef}>
...
</Tooltip>

Comment (lpoulter) [reply packages/axiom-components/src/Tooltip/Tooltip.js]: That's very cool. I've not used `useImperativeHandle` before. This seems like an easy add for a lot of power on the user side. Ref's a mess in Axiom and it's on my list to better and more consistently support them.

File: packages/axiom-components/src/Tooltip/Tooltip.js [hunk 3]
```diff
@@ -7,16 +7,23 @@ import PositionTarget from '../Position/PositionTarget';
...
   delay = false,
   onClick,
   position = 'top',
   enabled = true,
   ...rest
-}) {
+}, ref) {
   const [isVisible, setIsVisible] = useState(false);
 
+  React.useImperativeHandle(ref, () => ({
```
SeaOfTea: You don't need the `React.` you can just import it from React. I was just using it in the example to be clear where it came from.

PR #849, opened 2020-03-20
Popper 2

lpoulter:
Upgrades popper 1.x to 2.x following the [migration guide](https://popper.js.org/docs/v2/migration-guide/).

Breaking change that I think will affect no body. If anyone was using `<Tip>` with their own custom positioning then they will now need to use `<ContextTip>`

I've done some manual testing of this in BCR and all seems well. I also can't recreate the bug below. @TomWija Would you able to test in Viza and validate [this](https://github.com/BrandwatchLtd/axiom-react/pull/814) bug hasn't returned?

Fixed BCR bug.
ezgif-1-71408f25c6e7


packages/axiom-components/src/Position/Position.js
```diff
@@ -118,28 +119,50 @@ export default class Position extends Component {
...
         },
-        offset: {
-          enabled: false,
+      },
+      {
+        name: 'offset',
+        options: {
+          offset: () => {
+            if (showArrow) {
+              return [0, 15];
```
jdelight: interesting, what do these numbers mean?

lpoulter: popper offset to allow for arrow. Replaces margins in popper 2. I can name them.

jdelight:
Thanks for the heads up on the breaking change and detailed descriptions in the comments 👍 
We are using the CSS directly in Explore so we'll make sure to update this. 

I guess people are important `Tip` directly - is that why we're not deleting it?

lpoulter:
Yes Tip is used [in the component-picker](https://github.com/BrandwatchLtd/frontend/blob/master/reactApps/src/apps/component-picker/components/detailview/DetailView.js#L19).

PR #850, opened 2020-03-23
Docs improvements

lpoulter:
Adds the [knobs add on](https://www.npmjs.com/package/@storybook/addon-knobs) for storybook which will allow interactive props.
Add docs for,

- ColorPicker
- Dropdown
- Tip


package.json
```diff
@@ -31,8 +31,9 @@
     "@babel/polyfill": "^7.0.0",
     "@babel/preset-env": "^7.0.0",
     "@babel/preset-react": "^7.0.0",
-    "@storybook/addon-docs": "^5.3.13",
-    "@storybook/react": "^6.0.0-alpha.14",
+    "@storybook/addon-docs": "next",
+    "@storybook/addon-knobs": "next",
+    "@storybook/react": "next",
```
siggysamson: Are you on purpose using `next` as the version here? It's probably fine because the version is locked in `yarn.lock` anyways but just curious to know why you chose `next` here

lpoulter: We are using the alpha version of @storybook/react as it comes with support for monorepos and some other features we want. Releases are coming out fast and we will be wanting thoose changes as they happen.

PR #857, opened 2020-04-02
Remove unsafe componentWillMount

lpoulter:
`componentWillMount` is now marked as legacy and will be dropped completely in React 17. 

There is no direct replacement. `componentWillMount` methods will be moved to either the constructor or `componentDidMount` depending if they need to happen before or after first `render`.

Screenshot 2020-04-02 at 14 11 49

### DIFF
--- a/packages/axiom-components/src/Portal/Portal.js
+++ b/packages/axiom-components/src/Portal/Portal.js
@@ -14,7 +14,8 @@ export default class Portal extends Component {
     axiomPositionParentNode: PropTypes.object,
   };
 
-  UNSAFE_componentWillMount() {
+  constructor(props) {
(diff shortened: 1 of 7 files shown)


packages/axiom-components/src/Portal/Portal.js
```diff
@@ -14,7 +14,8 @@ export default class Portal extends Component {
     axiomPositionParentNode: PropTypes.object,
   };
 
-  UNSAFE_componentWillMount() {
+  constructor(props) {
```
SeaOfTea, line 17: It probably should've been in the constructor originally anyway!

lpoulter:
Cheers for the 👀 people.

PR #870, opened 2020-05-04
Don't generate flags on every build

lpoulter:
Currently Flags svgs are built on every build using the current axiom
color pallete. This isn't really nessacary we only need to update the flag SVGS when Axiom colors change. 

The script consisted of two parts.
1. Build a json map containing flag colours and names.
2. Build an SVG for each flag and place in /site to be deployed with the docs site. This SVG is then loaded from `https://axiom.brandwatch.com/assets/` when the Flag component is rendered. 

This PR removes the build flags step from the ci process. This fixes the flags as they are. I've committed the previously generated flags.json so it can be copied into the dist folders. The SVGs will live on the web bucket untouched.

I've left the script build_flags.sh to be run again if and when the axiom colors change. At some point we will also want to move the flag svgs and host on their own to a new bucket likey a CDN.


.storybook/static_assets/flags copy/ad.svg
```diff
@@ -1,152 +0,0 @@
-<svg xmlns="http://www.w3.org/2000/svg" xmlns:xlink="http://www.w3.org/1999/xlink" height="512" width="512" viewBox="0 0 512 512">
```
lpoulter, line 1: This is not nessacary as we can pull the svgs from the web-bucket.

lpoulter:
'flags copy' was added in error in a previous PR. I've deleted those assets all together.

PR #922, opened 2020-07-21
Word wrap table headers

larister:
This PR allows us to have wrappable table headers. After consultation with @Flawwles, it was decided to limit the number of possible lines to 2.

As noted in the commit message, to achieve this I've used the `line-clamp` property.

image


lpoulter:
@larister looks like the preview build has not worked. I guess that's because the PR is on your fork are you able to try moving it to upstream/https://github.com/BrandwatchLtd/axiom-react?

larister:
ah sure, didn't realise that. Thanks Laurie, will reopen

lpoulter:
I hand't considered it either will see if there's a fix or add some messaging. Thanks for reopening.

PR #923, opened 2020-07-21
Word wrap table headers

larister:
This PR allows us to have wrappable table headers. After consultation with @Flawwles, it was decided to limit the number of possible lines to 2.

As noted in the commit message, to achieve this I've used the `line-clamp` property.

image


lpoulter:
Screenshot 2020-07-21 at 22 08 03

Looks like the asc/desc indicator is shrinking due to the wrapping of the long header. You should be able to fix that by applying `flex-shrink: 0` to stop it shrinking.

larister:
Good catch @lpoulter, thanks. Fix pished.

PR #937, opened 2020-08-05
style: 💄 Adds secondary style to ProgressButton

dpodboj1:
I like progress buttons. (https://github.com/BrandwatchLtd/axiom-react/pull/935)

Since Progress already accepts a color prop of subtle (was using white which matched primary well) I thought I'd add support for the ProgressButton to be secondary as well. Wouldn't really make sense to enable other variants as they essentially don't have backgrounds.

Let me know what you think.


Flawwles:
👍  Looks good to me - Thanks for adding this

PR #947, opened 2020-08-14
docs: ✏️ Improve Form Validation docs

### LINKED ISSUE #918: Can we add more docs for validation
Currently it looks like the docs are a little light on how to implement text input validation. A few more examples / code sandbox or within the docs would be good. Typical use cases would be (required, email pattern, and custom validation).

lpoulter:
Closes #918


jdelight (approved):
Really easy to read - love it! thanks @lpoulter 

Am I right in thinking if you have a required prop on a text input - you also must have a requiredError prop on the form? Basically I think this is where I might have got stuck before because required is a [valid HTML attribute](https://developer.mozilla.org/en-US/docs/Web/HTML/Attributes/required) and it doesn't require anything additional to work natively. So this is where there's a difference which may need to be called out for a developer. I.e. it looks like you've added a required prop - you also need prop x for this to work. Alternatively, can we put in a default prop? This is all dependent on whether they are both needed - if not please ignore this comment :D 

Regardless of the above I don't think anything needs to change in Axiom validation right now - the docs are pretty clear about how it works. Just wondered if there's a hidden dependency there which differs from native HTML.

lpoulter:
A good question. Yes it looks like requiredError necessarily for required to work. Don't see why that is as you say we could include an default error message. It looks like there are some design changes coming to validation so I'll look at this as part of that work.

PR #950, opened 2020-08-17
feat: 🎸 Add source icon

lpoulter:


lilybrandwatch:
Hey @lpoulter I just realized that the color used is different than in my sketch file. Is it something that you could look at for this source icon to have the same color that the previous one created. Thanks

PR #951, opened 2020-08-18
fix(EditableLine): Call onChange on every change

josa42:
This PR changes the `EditableLine` from holding local state, to passing through the `onChange` handler and `value` property to the input.

**BREAKING CHANGE:** onChange of the `EditableLine` component is called on every change to the input now and receives the original event as an argument

---

Previously the `onChange` handler was only called on blur, if the value has changed. This caused an issue, where the value was reseted if the component was re-rendered while editing, before the blur event.

The issue can be produced in [this code sandbox](https://codesandbox.io/s/axiom-editable-line-issue-4j7fl).

### DIFF
--- a/packages/axiom-components/src/Editable/Editable.stories.js
+++ b/packages/axiom-components/src/Editable/Editable.stories.js
@@ -1,4 +1,4 @@
-import React from "react";
+import React, { useState } from "react";
 import EditableTitle from "./EditableTitle";
 import EditableLine from "./EditableLine";
 import Heading from "../Typography/Heading";
@@ -9,13 +9,15 @@ export default {
 };
 
 export function Default() {
+  const [value, setValue] = useState("Editable text...");
+
   return (
     <EditableTitle>
       <Heading textSize="headline">
         <EditableLine
-          onChange={function () {}}
+          onChange={(event) => setValue(event.target.value)}
           placeholder="Editable text here"
-          value="Editable text..."
+          value={value}
         />
       </Heading>
     </EditableTitle>
--- a/packages/axiom-components/src/Editable/EditableLine.js
+++ b/packages/axiom-components/src/Editable/EditableLine.js
@@ -1,14 +1,15 @@
 import PropTypes from "prop-types";
 import React, { Component } from "react";
+import classnames from "classnames";
 import "./EditableLine.css";
 
 export default class EditableLine extends Component {
   static propTypes = {
-    /** SKIP */
-    onBlur: PropTypes.func,
-    /** Callback with the value when the text has finished being edited */
+    /** Class name to be appended to the element */
+    className: PropTypes.string,
+    /** Handler for when the input field is changed */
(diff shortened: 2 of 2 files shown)


packages/axiom-components/src/Editable/EditableLine.js
```diff
@@ -60,20 +36,16 @@ export default class EditableLine extends Component {
   }
 
   render() {
-    const { value } = this.state;
-    const { placeholder, ...rest } = this.props;
+    const { className, placeholder, value, ...rest } = this.props;
 
     return (
-      <div className="ax-editable-line">
+      <div className={classnames("ax-editable-line", className)}>
```
lpoulter, line 42: Thanks for adding the classname.

packages/axiom-components/src/Editable/EditableLine.js
```diff
@@ -1,57 +1,33 @@
...
+import classnames from "classnames";
 import "./EditableLine.css";
 
 export default class EditableLine extends Component {
   static propTypes = {
-    /** SKIP */
+    /** Class name to be appended to the element */
+    className: PropTypes.string,
+    /** Handler for when the input field is blurred */
     onBlur: PropTypes.func,
```
lpoulter, line 11: So onBlur is now just spread to the html input? I wonder if we still need this prop might be able to remove it?

packages/axiom-components/src/Editable/Editable.stories.js
```diff
@@ -9,13 +9,15 @@ export default {
...
 
 export function Default() {
+  const [value, setValue] = useState("Editable text...");
+
   return (
     <EditableTitle>
       <Heading textSize="headline">
         <EditableLine
-          onChange={function () {}}
+          onChange={(event) => setValue(event.target.value)}
```
lpoulter, line 18: 👍

lpoulter:
To signal a breaking change to learna and in line with [conventionalcommits](BREAKING CHANGE: onChange of the EditableLine component is called on every change to the input now and receives the original event as an argument

) spec you need to add the breaking change message to the footer of your commit message e.g.
```
fix(EditableLine): Call onChange on every change

BREAKING CHANGE: onChange of the EditableLine component is called on every change to the input now and receives the original event as an argument.
```
You can also use `npx git-cz` to make the process a little easier.

josa42:
@lpoulter
> To signal a breaking change to learna and in line with conventionalcommits spec you need to add the breaking change message to the footer of your commit message

I must have lost that in a rebase 🙈  I'll fix that.

packages/axiom-components/src/Editable/EditableLine.js
josa42: We do not need it, I just kept it so it shows up in the documentation. But if you prefer that, I'll remove it.

josa42: @lpoulter That makes a lot of sense. I'll remove all props that are not explicitly used

PR #953, opened 2020-08-27
feat(Icons): Add arrow left and right

Flawwles:
We were missing the arrow left icon which is being used in a few upcoming designs. To complete the set, I also added the arrow right, which doesn't have an immediate use, but made sense to add at the same time. 

image


lpoulter (approved):
Thanks Matt.

PR #954, opened 2020-08-28
fix: editable text sets value on input, adds minimum width for hit area

stefanpearson:
#951 Added a change that removed `value` from the input, which seemed to break things for two reasons:

1. The value was ignored on the input
2. The `ax-editable-line__structure` element represented a different string, so the layout was broken

I've added this back in, and added a minimum width (of ~1 character), to make the hit-area more usable.

See sandbox for bug reproduction: https://codesandbox.io/s/axiom-boilerplate-forked-283h2?file=/src/components/App/index.js

It may not have been picked up because the storybook example had a similar string for both the value and placeholder, so the bug could have been missed in development.

cc @lpoulter @josa42

### DIFF
--- a/packages/axiom-components/src/Editable/EditableLine.css
+++ b/packages/axiom-components/src/Editable/EditableLine.css
@@ -6,6 +6,7 @@
 
 .ax-editable-line__structure {
   visibility: hidden;
+  min-width: 1em;
 }
 
 .ax-editable-line__input {
--- a/packages/axiom-components/src/Editable/EditableLine.css
+++ b/packages/axiom-components/src/Editable/EditableLine.css
@@ -5,8 +5,8 @@
 }
 
 .ax-editable-line__structure {
+  min-width: 1em; /* stylelint-disable-line unit-blacklist */
   visibility: hidden;
-  min-width: 1em;
 }
 
 .ax-editable-line__input {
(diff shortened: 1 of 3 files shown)


packages/axiom-components/src/Editable/EditableLine.css
```diff
@@ -5,6 +5,7 @@
 }
 
 .ax-editable-line__structure {
+  min-width: 1em; /* stylelint-disable-line unit-blacklist */
```
lpoulter, line 8: I guess this was added to take care of the case where there is no value or placeholder? Any reason to not use rem here? Axiom only uses rem I think mostly for the simplicity it brings over em.

lpoulter: Maybe this is the perfect use case for `ch`?

stefanpearson: Yeah it was quite fiddly to click on the editable text when it was either empty, or had a single narrow character (e.g. 'I')/ Usually I would agree about the em, but in this case, the editable text is agnostic of text size. This would represent around 1 character for whatever text size is currently inherited (i.e. one 'M').

lpoulter (approved):
Thanks muchly for this fix. I'll publish a new patch version now.

PR #955, opened 2020-09-07
Splits out TextInputIconWrapper from TextInputIcon

stefanpearson:
* Splits out a new `TextInputIconWrapper` from `TextInputIcon`. This allows for injecting a custom child into the space that would usually only be allowed for an `Icon`, whilst respecting the same layout.
* Remains backwards compatible, so that `TextInputIcon` can be used in the same way as before.
* Allows `style` to be applied to `Icon`, for edge-case CSS styles (Vizia requires a dot icon, with a dynamic colour applied that is not part of the Axiom product colour subset).
* Removes snapshots that seemed to be broken anyway?


packages/axiom-components/src/DurationPicker/DurationPicker.js
```diff
@@ -178,7 +178,7 @@ export default class DurationPicker extends Component {
                 readOnly
                 value={formatTimeUnit(selectedUnit)}
               >
-                <TextInputIcon name="chevron-down" />
+                <TextInputIcon align="right" name="chevron-down" />
```
stefanpearson, line 181: this was out of sync with the snapshot for some reason?

packages/axiom-components/src/Form/TextInputIcon.test.js
```diff
@@ -42,34 +42,4 @@ describe("TextInputIcon", () => {
     const tree = component.toJSON();
     expect(tree).toMatchSnapshot();
   });
-
-  it("renders with tooltip when string passed", () => {
```
lpoulter, line 46: Do we need to copy this test over to TextInputIconWrapper? I have started copying over some tests using react-testing-libary if you've used that before.

lpoulter: Could also add a story to input to cover this usecase whilst we are at it. Something like,
```

export function WithTooltipIcon() {
  return (
    <TextInput aria-label="Search">
      <TextInputIcon
        align="right"
        iconColor="error"
        name="information-circle"
        tooltip="This username exists please select a different one"
      />
    </TextInput>
  );
}
```

lpoulter (commented):
Thanks for this. One comment about a test that might have been lost.

packages/axiom-components/src/Form/TextInputIcon.test.js
stefanpearson: Ok will do, think I got confused by the test because the snapshot didn't show anything to do with a tooltip.

lpoulter (approved):
Looks good. The wrapper approach feels right, thanks for the contribution 🙇

PR #960, opened 2020-09-11
fix(TextInput): Use error colors for invalid states, even if focused

josa42:
Currently the `TextInput` which has `invalid={true}` and is focused, shows a blue border and normal text color.

textinput--pre

This PR makes sure the error color is used, even if the input is focused.

textinput-post


lpoulter (approved):
Looks good to me thanks for the pr. I'll cc in @claudiodiliberto who is taking a look at the overall validation designs.

PR #973, opened 2020-09-25
docs: ✏️ Context Docs

### LINKED ISSUE #970: Documentation of Context component
I'm implementing a tooltip that requires quite an amount of control over the style of its contents. It seems like this requires using the Position component and looking at the examples on that page it seems as if I ought to be using Context family components inside that. I've sort of figured out what I need to do by reading the source but I think it might be nice to have some documentation of the Context components and some (excuse the pun) context around what they are and how to use them to make it easier to understand, particularly as uses of the Position component are likely to be more complex and difficult to implement.

lpoulter:
Closes #970

### DIFF
--- a/packages/axiom-components/src/Context/Context.js
+++ b/packages/axiom-components/src/Context/Context.js
@@ -41,12 +41,8 @@ export default class Context extends Component {
       ...rest
     } = this.props;
 
-    const classes = classnames(
-      className,
-      "ax-context",
-      `ax-context--${position}`,
-      {
-        [`ax-context--${color}`]: color,
+    const classes = classnames(className, "ax-context", {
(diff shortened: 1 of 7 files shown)


packages/axiom-components/src/Context/Context.js
```diff
@@ -5,6 +5,12 @@ import Base from "../Base/Base";
 import ContextTip from "./ContextTip";
 import "./Context.css";
 
+/**
+ * Context provides styling elevating it's self from the background.
```
andrewru, line 9: A small typo here:

```suggestion
 * Context provides styling elevating itself from the background.
```

andrewru (approved):
Thanks for this Laurie :+1:

PR #984, opened 2020-10-06
wip document css variables

lpoulter:


lpoulter:
Super seeded by #1043

PR #986, opened 2020-10-08
feat: 🎸 Pass Popper Modifiers to Position

### LINKED ISSUE #985: Popper version / update
Just wondering which version of Popper Axiom uses for Position and if it's possible to switch to the latest / react-popper implementation? I.e. do we need our own version of Position anymore? Probably one to discuss more broadly.

One reason is documentation but also it's been updated and it's not clear if we have access to the latest improvements. For example, I'm working on an announcement feature which needs the ability to put a padding around the context so it doesn't hit the edge of the browser window.

image

lpoulter:
This will allow setting and overriding of Popper modifiers via the Axiom Position component. Other components that spread props to Position such as Dropdown and Tooltip will also support this feature.

Closes #985

### DIFF
--- a/packages/axiom-components/src/Position/Position.stories.js
+++ b/packages/axiom-components/src/Position/Position.stories.js
@@ -3,10 +3,19 @@ import Position from "./Position";
 import PositionTarget from "./PositionTarget";
 import PositionSource from "./PositionSource";
 import Context from "../Context/Context";
+import ContextContent from "../Context/ContextContent";
 import ContextMenu from "../Context/ContextMenu";
 import Icon from "../Icon/Icon";
+import Button from "../Button/Button";
+import IconButton from "../Icon/IconButton";
+import ButtonGroup from "../Button/ButtonGroup";
+import ButtonIcon from "../Button/ButtonIcon";
 import Heading from "../Typography/Heading";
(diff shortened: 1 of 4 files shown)


packages/axiom-components/src/Position/Position.stories.js
```diff
@@ -39,6 +48,70 @@ export function Default(props) {
...
+                  <Paragraph>
+                    Using powerful AI, Brandwatch Search identifies what you’re
+                    looking for. This includes brands, organizations and people.
+                    It then creates an accurate search automatically.
+                  </Paragraph>
+                  <Paragraph>
+                    And there’s no need for writing long, complex boolean. Just
+                    start typing.
+                  </Paragraph>
+                  <ButtonGroup>
```
jdelight, line 103: Some nice improvements here - thanks.

jdelight:
Nice work @lpoulter - I think this should help a lot! 👍 do you think it's going to cause any issues elsewhere? looks like it gets added to the list of modifiers with the existing props still present. I guess it's possible for someone to override an existing modifier but you'd probably not do it by accident and would probably know what you're doing. So this solution seems fine to me.

ogandose (approved):

PR #994, opened 2020-11-17
fix(axiom-components): animated logo data bg colour

lcapel:


packages/axiom-components/src/AnimatedLogo/logos/DataLogo.js
```diff
@@ -3,7 +3,7 @@ import React from "react";
 function DataLogo() {
   return (
     <>
-      <rect fill="#57b7dd" x="0" y="0" width="28" height="28" />
```
josa42, line 6: Shouldn't the rect be inside the svg tag?

lcapel: I just used one of the other logos as a template without thinking too much about it 😅 I can see that this is a nested SVG that gets injected here https://github.com/BrandwatchLtd/axiom-react/blob/master/packages/axiom-components/src/AnimatedLogo/AnimatedLogo.js#L103 so technically it is inside the svg tag

josa42: Oh I see, thanks  🙂

PR #1003, opened 2020-11-24
feat(Chip): updating design

ogandose:
Updating chip to new designs.

### DIFF
--- a/packages/axiom-components/src/Chip/Chip.css
+++ b/packages/axiom-components/src/Chip/Chip.css
@@ -1,4 +1,7 @@
 .ax-chip {
+  --chip-color-opacity: 0.2;
+  --chip-background-opacity: 0.08;
+  --chip-background-opacity-hover: 0.15;
   display: flex;
   align-items: center;
   justify-content: space-between;
@@ -10,7 +13,7 @@
   border-style: solid;
   border-radius: var(--component-border-radius-huge);
   border-color: transparent;
-  background-color: var(--color-ui-white-noise);
+  background-color: rgba(var(--rgb-ui-carbon), var(--chip-background-opacity));
   color: var(--color-ui-carbon);
   letter-spacing: var(--letter-spacing-loose);
   text-align: center;
@@ -22,7 +25,7 @@
   transition-timing-function: var(--transition-function);
 
   &:hover {
-    background-color: var(--color-ui-white-noise--dark);
+    background-color: rgba(var(--rgb-ui-carbon), var(--chip-background-opacity-hover));
     cursor: pointer;
   }
 
@@ -32,6 +35,7 @@
     background-color: var(--color-theme-background--disabled);
     color: var(--color-theme-text--disabled);
     cursor: default;
+    pointer-events: none;
   }
 
   &[disabled]:hover {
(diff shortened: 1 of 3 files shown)


packages/axiom-components/src/Chip/Chip.stories.js
```diff
@@ -17,9 +17,55 @@ export function Default() {
...
+InputChips.parameters = {
+  docs: {
+    description: {
+      story:
+        "Input chips represent a user entry in a compact form. The text entered by the user is validated and converted into chips",
+    },
+  },
+};
+
+export function ChoiceChips(props) {
```
lpoulter, line 53: It would be great to make these examples interactive. The Storybook storys run in React so you can useState etc.

packages/axiom-components/src/Chip/Chip.stories.js
```diff
@@ -17,9 +17,55 @@ export function Default() {
...
       <Chip disabled>Jayson Tatum</Chip>
-      <Chip active metric="(1%)" rightIcon="cross">
+      <Chip active metric="(1%)">
         Al Horford
       </Chip>
     </ChipList>
   );
 }
+
+export function InputChips(props) {
```
lpoulter, line 27: It looks like we aren't using the props here. The props here are populated by storybook based on the state of the probs in the props table. I've been using them to make the default stories interactive but that doesn't really make as much sense for more complex examples.

packages/axiom-components/src/Chip/Chip.stories.js
lpoulter: Generally a story should be a complete example of a how a component is to be used.

packages/axiom-components/src/Chip/Chip.css
```diff
@@ -47,15 +49,25 @@
...
-  background-color: var(--color-ui-accent);
-  color: var(--color-ui-white);
+  border-width: var(--component-border-width);
+  border-style: solid;
+  border-radius: var(--component-border-radius-huge);
+  border-color: var(--color-ui-accent);
+}
+
+.ax-chip--valid {
+  background-color: rgba(var(--rgb-ui-success), 0.2);
```
lpoulter, line 59: you could make the opacity here a css var scoped to the chip `ax-chip` class. Something like
```
.ax-chip {
  --chip-hover-opacity: 0.2;

...

background-color: rgba(var(--rgb-ui-success), var(--chip-hover-opacity));
```

packages/axiom-components/src/Chip/Chip.js
```diff
@@ -32,26 +36,32 @@ export default class Chip extends Component {
...
     });
 
     return (
       <Base className={classes} disabled={disabled} {...rest}>
-        {leftIcon && <Icon className={"ax-chip-icon"} name={leftIcon} />}
+        {leftIcon && (
+          <Icon
+            className={"ax-chip-icon"}
+            name={leftIcon}
+            color={active ? "ui-accent" : ""}
```
lpoulter, line 60: "" will trigger a prop type warning I think, better to use null.

lilybrandwatch:
Hey @ogandose the default grey background looks a bit light. I'm using the active color which is #3F3F3F in 8% opacity which you can find here:
https://axiom-react.netlify.app/docs/packages/axiom-materials/themes
The same default color should be use for the input one. This is for make sure there is enough contrast. 

The rest looks good to me. Thanks.

lpoulter (approved):
Good stuff, thanks.

PR #1009, opened 2021-01-05
[psi-300] feat: add topics wheel svg pictogram

Jafferwaffer:
The topics wheel component is being added to vizia.
We need to add this SVG to pictograms component so it can be used in the vizia platform like other visualisation.


lpoulter:
You will need to also import the svg into [Pictogram](https://github.com/BrandwatchLtd/axiom-react/blob/master/packages/axiom-components/src/Pictogram/Pictogram.js#L6). Also could you add a reference to it in the story, [here](https://github.com/BrandwatchLtd/axiom-react/blob/master/packages/axiom-components/src/Pictogram/Pictogram.stories.js#L11) so it shows up in storybook?

Jafferwaffer:
Thanks will the review, will make changes now

lpoulter:
Preview looks good, will merge. cheers.

lpoulter (approved):

PR #1010, opened 2021-01-05
fix: fill colour on topics-wheel pictogram svg

Jafferwaffer:
Fixes the fill colour on topics wheel svg added here https://github.com/BrandwatchLtd/axiom-react/pull/1009
If its not #000000 then `fill: currentColor` doesn't seem to work properly


lpoulter:
seems a little strange that current color works with #000000 and not #3F3F3F do you know why?

Jafferwaffer:
Seems to be a 'gotcha' with SVGs: https://css-tricks.com/gotchas-on-getting-svg-into-production/#gotcha-two-working-with-a-designer

lpoulter (approved):

PR #1021, opened 2021-03-05
fix: 🐛 Update css rule to handle disabled for TableHeaderLabel

PTihomir:
Css selector for handling cursor and color for TableHeaderLabel was only checking `:disabled` pseudo class, but the pseudo class `:disabled` is only available for input elements. TableHeaderLabel is a div, and to make css selector target it, I also added the attribute selector to the css rule.

### DIFF
--- a/packages/axiom-components/src/Table/Table.css
+++ b/packages/axiom-components/src/Table/Table.css
@@ -87,7 +87,7 @@
   color: var(--color-theme-text);
 }
 
-.ax-table__header-button:disabled {
+.ax-table__header-button--disabled {
   cursor: auto;
 }
 
@@ -114,7 +114,7 @@
   transform: scaleX(1) translateY(var(--spacing-grid-size--x2));
 }
 
-.ax-table__header-button:hover:not(:disabled) {
+.ax-table__header-button:hover:not(.ax-table__header-button--disabled) {
   color: var(--color-ui-accent--hover);
 }
 
--- a/packages/axiom-components/src/Table/Table.stories.js
+++ b/packages/axiom-components/src/Table/Table.stories.js
@@ -22,7 +22,9 @@ export function Default() {
   return (
     <Table>
       <TableHeader>
-        <TableHeaderLabel sortDirection="ascending">Column A</TableHeaderLabel>
+        <TableHeaderLabel sortDirection="ascending" onClick={() => {}}>
+          Column A
+        </TableHeaderLabel>
         <TableHeaderLabel>lower case</TableHeaderLabel>
         <TableHeaderLabel wrap={true}>
           Super long Column C label with wrapping Lorem ipsum
--- a/packages/axiom-components/src/Table/TableHeaderLabel.js
+++ b/packages/axiom-components/src/Table/TableHeaderLabel.js
@@ -65,13 +65,17 @@ export default class TableHeaderLabel extends Component {
       width: `${Math.max(0, Math.min(width, 100))}%`,
     };
 
+    const buttonClasses = classnames("ax-table__header-button", {
+      "ax-table__header-button--disabled": !onClick,
+    });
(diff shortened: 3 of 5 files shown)


packages/axiom-components/src/Table/Table.css
```diff
@@ -87,7 +87,8 @@
   color: var(--color-theme-text);
 }
 
-.ax-table__header-button:disabled {
+.ax-table__header-button:disabled,
+.ax-table__header-button[disabled] {
```
jdelight, line 91: Could be a nice opportunity to use the `:is()` selector.

```CSS
.ax-table__header-button:is(:disabled, [disabled]) {...}
```

Allows a more concise selector but also is slightly more robust. https://developer.mozilla.org/en-US/docs/Web/CSS/:is#forgiving_selector_parsing

packages/axiom-components/src/Table/Table.css
```diff
@@ -114,7 +115,7 @@
   transform: scaleX(1) translateY(var(--spacing-grid-size--x2));
 }
 
-.ax-table__header-button:hover:not(:disabled) {
+.ax-table__header-button:hover:not(:disabled):not([disabled]) {
```
jdelight, line 118: :is() could also be used here perhaps?

jdelight:
Thanks @PTihomir - minor comment but looks good 👍

josa42:
@PTihomir 

> TableHeaderLabel is a div, and to make css selector target it, I also added the attribute selector to the css rule.

I am a little confused by this. A div should not have a disabled attribute. Or am I missing something?

See: https://developer.mozilla.org/en-US/docs/Web/HTML/Attributes/disabled

packages/axiom-components/src/Table/Table.css
PTihomir: I am not using safari, so not sure if safari implements this selector in the prev 2 version. https://caniuse.com/css-matches-pseudo
If You say it is safe to use, I will change it.

PTihomir:
> @PTihomir
> 
> > TableHeaderLabel is a div, and to make css selector target it, I also added the attribute selector to the css rule.
> 
> I am a little confused by this. A div should not have a disabled attribute. Or am I missing something?
> 
> See: https://developer.mozilla.org/en-US/docs/Web/HTML/Attributes/disabled

Well You are right,  it can't be disabled as it is not an element from the list of elements which can be disabled. But still can get the attribute.
https://github.com/BrandwatchLtd/axiom-react/blob/master/packages/axiom-components/src/Table/TableHeaderLabel.js#L72

I tought about actually replacing in the css selectors the `:disabled` pseudoselector and to just use the attribute selector, but at the time I stumbled on this issue, I was not sure if I break that way something. So I ended up just adding the attribute selector, this way this PR has very little breaking change potential.

Now that I try to explain this, I am more confident, that as the table label text would never be a button or similar, I should probably just replace :disabled with [disabled].

packages/axiom-components/src/Table/Table.css
josa42: According to [MDN](https://developer.mozilla.org/en-US/docs/Web/CSS/:is#forgiving_selector_parsing) Safari does not support `:is()` for a list of selectors, and Safari 13 does not support `:is()` at all, so we cannot use it. (At least not without pre-processing.)

josa42:
@PTihomir IMHO `TableHeaderLabel` should add a css class (`ax-table__header-button--disabled`) instead of using the disabled attribute in that case. But I am not entirely sure if using `role="button"` allows the usage of `disabled`. So I'll leave this up to you 🙂

josa42 (approved):

josa42 (approved):
👍🏻 Thanks for the update! Looks good to me!

PR #1023, opened 2021-03-12
Allows Dropdowns to be nested

stefanpearson:
I ran into a problem where I had a Dropdown menu, which had a Select menu nested inside it. Clicking on a select option would obliterate the parent dropdown, because the Dropdown tears down when a node is clicked outside of its current DOM tree (the subtree portals were the issue here). Without rewriting all of this nested context / ref handling, managing these global event listeners seemed like the most sensible way to achieve this.

### DIFF
--- a/packages/axiom-components/src/Dropdown/DropdownContext.js
+++ b/packages/axiom-components/src/Dropdown/DropdownContext.js
@@ -10,6 +10,35 @@ if (typeof window !== "undefined") {
   require("element-closest");
 }
 
+/**
+ * Manage global document listeners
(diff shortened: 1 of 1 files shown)


TomWija (approved):

stefanpearson:
I'm going to close this for now, whilst it might work for my edge-case, it doesn't actually fix the underlying issue (it will break  when nesting more than 1 nested dropdown) and should be addressed properly later. For my use-case, I'm going to use a custom Position/Context component and bypass the Dropdown component.

PR #1030, opened 2021-05-20
fix(axiom-components): fix horizontalPadding in TableCell

tomru:
While experimenting with nesting tables I've stumbled over an issue when trying to set `horizontalPadding="none"` to a table cell, that did not work as expected.

See this [sandbox](https://codesandbox.io/s/axiom-tablecell-horizontalpadding-none-bug-sc5n0). This PR fixes the wrong usage of css properties for both the `none` case and `medium` case (which is the default anyways).

### DIFF
--- a/packages/axiom-components/src/Table/Table.css
+++ b/packages/axiom-components/src/Table/Table.css
@@ -41,13 +41,13 @@
 }
 
 .ax-table__cell--horizontal-padding-none {
-  padding-top: 0;
-  padding-bottom: 0;
+  padding-right: 0;
+  padding-left: 0;
 }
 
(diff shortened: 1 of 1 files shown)


jdelight (approved):
Thanks @tomru - surprised this wasn't noticed before - although there are probably missing tests around this.

tomru:
> although there are probably missing tests around this.

the existing tests would not have helped here, as they just check if the right css class is set :fried_shrimp:

tomru:
@BrandwatchLtd/axiom can I please have a 2nd review on this one?

josa42 (approved):

lpoulter:
Thanks for the fix.

PR #1042, opened 2021-07-28
NOTICKET Adds format as an option to fromNow

dpodboj1:


packages/axiom-formatting/src/fromNow.test.js
```diff
@@ -6,9 +6,13 @@ const MINUTE = 60000;
 describe("fromNow", () => {
   describe("fixed date in the past", () => {
     it("formats date correctly", () => {
-      expect(fromNow(new Date(2010, 10, 10).toISOString())).toBe(
-        "Wed 10 Nov 2010"
-      );
+      expect(fromNow(new Date(2010, 10, 10).toISOString())).toBe("10 Nov 2010");
+    });
+
+    it("formats date in long correctly", () => {
```
bwilson-bw, line 12: NABD, not a fan of writing `correctly` in test names. Think it would suffice without.

bwilson-bw (approved):

PR #1048, opened 2021-08-13
refactor:💡 Rescope CSS vars from global to component

lpoulter:
## Problem
We want each component's CSS file to be tree-shakable, that is without side effects.

As Axiom tree shakes unused components away we cannot have components depending on other component's CSS files else we will get data loss. This data loss has been observed in [this PR](https://github.com/BrandwatchLtd/consumer-research/pull/1028#issuecomment-885447037). There are a few places where this happens in the form of CSS variables. 

## Solution
Remove all instance of global CSS declarations, those using `:root`. The only place Axiom exposes global CSS vars should be in axiom-materials. Internally Axiom components should only use variables in axiom-materials or create new one's scoped only to that component, avoid the `:root` selector.  

## Assurance
Each CSS var usage was searched across the brandwatch org. Most are not used some have a few instances of usage marked as a breaking change for that reason.

BREAKING CHANGE: 🧨 --cmp-grid-* vars are now no longer on the global scope.

Please use var(--spacing-grid-size--x1) or equivalent instead.
```
--cmp-grid-gutter-size-tiny now --spacing-grid-size--x2
--cmp-grid-gutter-size-small now --spacing-grid-size--x4
--cmp-grid-gutter-size-medium now --spacing-grid-size--x6
--cmp-grid-gutter-size-large now --spacing-grid-size--x8

--cmp-context-padding-large now --spacing-grid-size--x6
--cmp-context-menu-item-padding-vertical now --spacing-grid-size--x3

--cmp-context-padding-medium now --spacing-grid-size--x4
```


larister (approved):
LGTM, thanks Laurie 👍

PR #1055, opened 2021-09-22
Monospace

stefanpearson:
Adds Roboto Mono for when monospace is required.

I feel like we should figure out a sensible way to import these fonts. At the moment, it is implicitly left up to the application to declare them as dependencies and import these fonts; I've seen quite a few inconsistencies and especially when including weights/variants. This can lead to the browser adding 'faux' transforms to the characters because the actual font data isn't available. Should the type-face ever change, all consumers will need updating.

Perhaps the Base component can write these Google links to the head? cc @lpoulter


packages/axiom-components/src/Typography/Typography.stories.mdx
```diff
@@ -124,6 +124,12 @@ Error is used exclusively to indicate an invalid state
   <Text textColor="error">error text</Text>
 </Preview>
 
+Monospace can be used when characters need a consistent width (e.g. code).
```
lpoulter, line 127: 👍

.storybook/preview-head.html
```diff
@@ -1 +1,3 @@
-<link href="https://fonts.googleapis.com/css?family=Roboto:100,300,400,500,700" rel="stylesheet">
\ No newline at end of file
+<link rel="preconnect" href="https://fonts.googleapis.com">
+<link rel="preconnect" href="https://fonts.gstatic.com" crossorigin>
+<link href="https://fonts.googleapis.com/css2?family=Roboto+Mono:ital,wght@0,100;0,300;0,400;0,500;0,700;1,100;1,300;1,400;1,500;1,700&family=Roboto:ital,wght@0,100;0,300;0,400;0,500;0,700;1,100;1,300;1,400;1,500;1,700&display=swap" rel="stylesheet">
```
lpoulter, line 3: Nice not seen `rel="preconnect"` before. Why do we need too if we are only calling one URL?

packages/axiom-components/src/Base/Base.js
```diff
@@ -136,6 +136,8 @@ export default class Base extends Component {
     textStrong: PropTypes.bool,
     /** Text underlined styling */
     textUnderline: PropTypes.bool,
+    /** Text mono-space */
+    textMono: PropTypes.bool,
```
lpoulter, line 140: I wonder if we might add more font styles. We might want to guard against that by making this something like `fontVariant` with a default of `body` and also accepting `monospace`. What do you think?

stefanpearson: Yeah possibly, the naming should probably be `textFace` (with `default` and `monospace` options), because variant often refers to bold/italic (which can also be used in combination with monospace). `body` seems to be used for either colour and text size so might end up mixing up our namespaces a bit.

.storybook/preview-head.html
stefanpearson: This was ripped straight from the current google fonts snippet. Must be a mirror or something googleapis redirects to? The preconnect bit I think just makes it a priority so it does the DNS lookup asap.

packages/axiom-components/src/Base/Base.js
lpoulter: Good points. Maybe `fontFace` as that is pretty much what it is but then `textFace` is does group it with our other text style props. `textFace` just wins for me I think.

lpoulter:
Very point about users having to require out fonts. I'll have about the best way we could inject them without also injecting them twice and keeping performant, IIRC `@font-face` has performance issues. Think that work should be in a separate PR.

lpoulter (approved):
Thanks

PR #1056, opened 2021-09-23
feat: adds notification with loading (ProgressInfinite) icon

tommillward:
This PR adds a 'Loading notification' to axiom-components 

The use case for this type of notification can be found here: https://gallery.io/files/300e2996746b4d1c9cf1984626b136a8


lpoulter:
This looks great. What do you think about adding a story that show's how you might transition from loading to success? I think that would be a nice example to show. Something like
```
export function Submit() {
  const [loadingState, setLoadingState] = useState("idle");

  useEffect(() => {
    let timer;
    if (loadingState === "loading") {
      timer = setTimeout(() => {
        setLoadingState("success");
      }, 1000);
    }
    return () => clearTimeout(timer);
  }, [loadingState]);

  return (
    <>
      <Button onClick={() => setLoadingState("loading")}>Submit</Button>

      {loadingState === "loading" && (
        <Notification type="loading">Loading notification</Notification>
      )}
      {loadingState === "success" && (
        <Notification type="success">Success notification</Notification>
      )}
    </>
  );
}
```

tommillward:
That's awesome cheers @lpoulter! I've added that to the story and it looks great

lpoulter (approved):
Looks great thanks for the addition.

PR #1069, opened 2021-11-03
fix(tip): Fix setting className on Tip component

josa42:
This PR fixes setting the className property in the Tip component.


packages/axiom-components/src/Tip/Tip.js
```diff
@@ -61,10 +62,11 @@ export default class Tip extends Component {
       children,
       shadow,
       size,
+      className,
       ...rest
```
josa42, line 66: Because `rest` is applied to the Base component like this

```
<Base className={classes} {...rest}>
```

`className` could be overridden.

lpoulter (approved):
Thanks.

PR #1074, opened 2021-12-17
feat(chip): label tweaks and allowing className

jonathanmiles:
ℹ️ <strike>TODO: Double check the impact of this on existing chip list usage.</strike>

<hr>

### What this does

* Allows `className` to be set on `<ChipList>`.
* Doesn't render the `.ax-chiplist-label` div when no label has been specified.

### Why

When no label has been specified there will still be space taken up by it, creating an unwanted indentation on the left of the chip list:

chiplist-empty-label

Not rendering the label, when it isn't needed, means that the chiplist has a flush left edge - this makes it easier to work with in terms of aligning it with the rest of the UI:

chiplist-no-label

The label div is still rendered when the list needs to be expandable:

chiplist-no-label-expandable

I've added `className` support because that tends to be a useful thing to have (to work around the chip list's current label behaviour, I had to wrap it in a div with a class so that I could target it).


lpoulter:
Thanks for list makes a lot of sense to me. Would be nice to add a story for expandable something like
```
export function Expandable(){
  return <div style={{width: '200px'}}><InputChips></InputChips></div>
}

Expandable.parameters = {
  docs: {
    description: {
      story:
        "When the list container is not wide enough to contain the whole ChipList the list is truncated. A chevron is added which will expand the list on click."
    },
  },
};
```

lpoulter:
Looking at the way chiplist is implemented it seems we could also improve it by by checking `isExpandable` on every render with something like `document.onresize` but a task for another day.

lpoulter:
In terms of the affect on existing chipLists we could mark this as a breaking change but I think as the change is quite minimal and the component is not used widely we can check some in place examples. I use this search to find axiom components across the Brandwatch org  https://github.com/search?q=org%3ABrandwatchLtd+ChipList+extension%3Ajsx+extension%3Ajs+org%3ABrandwatchLtd+-repo%3ABrandwatchLtd%2Faxiom-react+-repo%3ABrandwatchLtd%2Faxiom&type=code

packages/axiom-components/src/Chip/Chip.stories.js
```diff
@@ -100,3 +100,20 @@ ChoiceChips.parameters = {
...
+
+export function Expandable() {
+  return (
+    <div style={{ width: "200px" }}>
+      <InputChips></InputChips>
+    </div>
+  );
+}
+
+Expandable.parameters = {
```
lpoulter, line 112: 🍬  thanks for adding this.

lpoulter (approved):
Much nicer not having that mark-up. Thanks.

lpoulter:
I reviewed ChipList usage in BCR and found several places where ChipList is used without a label. They are as follows.

- https://github.com/BrandwatchLtd/consumer-research/blob/main/apps/administration-client-user-table/components/user-table/visible-user-controls.js#L9
- https://github.com/BrandwatchLtd/consumer-research/blob/a3e638a8a0d14160c67a84b1df9f29e84935a400/apps/social-panels/components/SocialPanelPage/SocialPanelsChipList.js#L10
- https://github.com/BrandwatchLtd/consumer-research/blob/main/apps/dashboards-page/components/sharing-filters/SharingFilters.js#L23
- https://github.com/BrandwatchLtd/consumer-research/blob/main/packages/segmentation/components/filters/FilterDropdown.js#L48
- https://github.com/BrandwatchLtd/consumer-research/blob/main/apps/administration-data-source-overview/components/datasourceoverviewtable/DataSourceOverviewTableContainer.js#L135
- https://github.com/BrandwatchLtd/consumer-research/blob/main/apps/query-builder/components/body/AudienceListsDropdownChipList.js#L116
- https://github.com/BrandwatchLtd/consumer-research/blob/main/packages/segmentation/components/filters/LocationFilterDropdown.js#L103
- https://github.com/BrandwatchLtd/consumer-research/blob/main/apps/social-panels/components/ViewPanel/ViewPanel.js#L380

Many of these already set some variation of 
```
.ax-chiplist-label {
    display: none;
}
```
Which can be removed after this merge. 2 make no style adjustments when the label is empty so I will test this doesn't not introduce any visual bugs there.

PR #1078, opened 2022-01-19
fix(axiom-composites): new password input non-alphanumeric character …

lcapel:
…validation

QA have reported a bug where although we request the user to include at least 1 non-alphanumeric character in their password the character `:` is not recognised as one.

I have found that this is explicitly been included in a regex negated here (https://github.com/BrandwatchLtd/axiom-react/blob/master/packages/axiom-composites/src/FormInputs/NewPasswordInput.js#L85), this is 4 year old code though and I’m not sure why this is.

Having had a look at the Java end I can see that we use this rule here (https://github.com/BrandwatchLtd/bw-app/blob/aea7dbb6d608c918babf2e1b19b32adbf492f[…]/java/com/brandwatch/dal/validation/password/PasswordUtils.java) and by looking at the vt-password docs (https://github.com/dfish3r/vt-password/blob/master/src/main/java/edu/vt/middleware/password/NonAlphanumericCharacterRule.java#L49) it seems like it should be ok to remove the colon from the regex’s negated set.


lpoulter (approved):
Makes sense, thanks.

PR #1080, opened 2022-01-21
feat(TableHeaderLabel): allow inline styles prop on TableHeaderLabel

SeaOfTea:
Pretty self explanatory, but currently the TableHeaderLabel will ignore use of the style prop. This will interfere with libraries that help with table management (and their layout to an extent) like `react-table`.

I'm using `rest.style` because I don't want it to appear as an explicit prop, or something that needs to be in the proptypes, just a value that will pass through to be spread on the underlying `th` element.


lpoulter (approved):
Thanks for the addition.

PR #1085, opened 2022-01-25
fix(Dropdown): Preventing dropdown to open if it's not enabled

josa42:
This PR prevents toggling the internal isVisible state of a Dropdown component, while it is not enabled. This could cause the dropdown to open, when it is re-enabled:

https://user-images.githubusercontent.com/423234/150944854-785c27ad-11eb-4c9f-a018-4d3507ca801e.mp4

_Best reviewed without whitepsace changes: [link](https://github.com/BrandwatchLtd/axiom-react/pull/1085/files?w=1)_


packages/axiom-components/src/Dropdown/Dropdown.js
```diff
@@ -59,6 +59,17 @@ export default class Dropdown extends Component {
...
   }
 
+  componentDidUpdate() {
+    const { enabled, onRequestClose } = this.props;
+    const { isVisible } = this.state;
+
+    if (!enabled && isVisible) {
+      this.setState({ isVisible: false });
+
+      if (onRequestClose) onRequestClose();
```
stefanpearson, line 69: I'm not sure if there is a linter in place to check but I think this is against the code-style
```suggestion
      if (onRequestClose) {
        onRequestClose();
      }
```

packages/axiom-components/src/Dropdown/Dropdown.js
```diff
@@ -76,19 +87,23 @@ export default class Dropdown extends Component {
...
   open() {
-    const { onRequestOpen } = this.props;
+    const { enabled, onRequestOpen } = this.props;
 
-    this.setState({ isVisible: true });
+    if (enabled) {
+      this.setState({ isVisible: true });
 
-    if (onRequestOpen) onRequestOpen();
+      if (onRequestOpen) onRequestOpen();
```
stefanpearson, line 95: as above ^

packages/axiom-components/src/Dropdown/Dropdown.js
```diff
@@ -76,19 +87,23 @@ export default class Dropdown extends Component {
...
   close() {
-    const { onRequestClose } = this.props;
+    const { enabled, onRequestClose } = this.props;
 
-    this.setState({ isVisible: false });
+    if (enabled) {
+      this.setState({ isVisible: false });
 
-    if (onRequestClose) onRequestClose();
+      if (onRequestClose) onRequestClose();
```
stefanpearson, line 105: as above ^

stefanpearson (approved):
All looks good to me. There are some minor issues with the code-style in that I typically don't see brace-less conditions anywhere else in Axiom/other BW codebases. Although, prettier doesn't seem to mind!

packages/axiom-components/src/Dropdown/Dropdown.js
josa42: Fair enough, I thought I get way with it since the linter allows it 😄

josa42: @stefanpearson `¯\_(ツ)_/¯`

image

PR #1093, opened 2022-03-29
Ba 725 text limit indicators

tommillward:
This PR introduces a 'limit indicator' for both the TextInput and TextArea Axiom components.

Why?
Design have mocked up a '_x_ characters left' pattern for the text input for the dashboards page and labels modals. 
Screenshot 2022-03-29 at 14 23 17
This had been implemented only for these specific modals only. It's a pattern the design intend to use more so adding it to Axiom seems sensible.

Also, it says it already exists in the new axiom documentation for the text area... but it doesn't yet - https://axiom.brandwatch.com/7dde06551/p/080af8-text-area/t/50bcd0


stefanpearson:
> I was thinking the same, but I think there are two main reasons to keep the input component stateless: First one being, that it is an established pattern since the invalid state is handled the same way. Secondly, it is very likely that you need the invalid state outside of the input component.

Perhaps a bit of both then? Pass `limit` and `limitInvalid` and keep it stateless, but allow the component to do the consistent formatting.

stefanpearson:
> > Perhaps a bit of both then? Pass limit and limitInvalid and keep it stateless, but allow the component to do the consistent formatting.
> 
> Ya, sounds good to me. I could also see `limitInvalid` being optional and defaulting to `value.length > limit`.

Yep! Although, on second thoughts, why would we need `limitInvalid` at all? The manager of state knows what the limit is and would know if it's invalid. The component can still be stateless and just render the limit invalid as you said `(value.length > limit)`

tommillward:
@stefanpearson @josa42 Thank you for these comments!

> Perhaps it doesn't need an invalid state at all, and stops firing the onChange callback when the limit has been reached?

We actually experimented with this when implementing this pattern for the dashboard labels work, but we decided having the text input unresponsive when invalid and when the user is trying to type something could be quite jarring/weird for them. ’Tis a matter of option though.


One of the reasons I decided to strip it right back and make the component not worry about any actual limits was to make it really flexible.
Because (for the text area at least) a character limit [value.length] isn’t the only limit that could be used or displayed. 
For example, when inviting new users in BCR there’s a text area that creates chips for each email address.
Screenshot 2022-03-30 at 15 41 10
So I was trying to cater for this type of use case of counting/limiting the chips. Or whatever content can be used

tommillward:
Which actually brings me to think maybe is it too flexible, to the point of why is it called ‘limit’ indicator. So your second point, Stefan, could definitely be an option.

> 2. Alternatively, if the logic has to be computed outside of this component, this looks like it could be a more generic slot than being reserved specifically for character length. Could it just be secondaryLabel and secondaryLabelInvalid?


So making it a more generic slot could make sense, but then again what else would people want to put there and want do the Axiom gods want to be there too? So maybe we’re treading into some UX ground but i don't know whether it's good to cater for any unknown use cases?

stefanpearson:
Definitely good points. The invite thing is an interesting use-case and probably confirms that this should be as generic as possible. Maybe it could even be:

```js
const MyComponent = () => {
  const [value, setValue] = '';
  return (
    <Input value={value}>
      <InputHint>
        Anything! Limit indicator, hints, supportive information; state could change the color of the text here
      </InputHint>
    </Input>
  )
};
```

tommillward:
I love that ^ 
I'll give that a go now

tommillward:
I've made some amendments @stefanpearson working from your idea. How does it look? (tagging you @josa42 in case you're curious too)

stefanpearson:
Yeah that looks really cool, nice to see it follows the same pattern as the `usageHint` as well!

Out of interest, what happens if `inlineLabel={true}`?

PR #1094, opened 2022-04-08
AFA-264. Some initial type def updates.

andrewru:
We're investigating using Typescript in Consumer Research and as part of this some of the Axiom types need some changes. From the history of the type file it looks like updating this manually (since the original auto-generated version) is the way that people have gone so have gone that way.

I think that the changes here are in keeping with the agreed interfaces.

It's likely that there will be more PRs like this over time as we port more code to Typescript. 

## Link type change

This change allows the following:

```javascript
<Link href="whatevertext" />
```

## ElipsisTooltip

The purpose of this change is to allow us to do this:

```javascript
<ElipsisTooltip className="myLovelyClass" />
```


packages/axiom-components/src/index.d.ts
```diff
@@ -1782,11 +1782,11 @@ declare module '@brandwatch/axiom-components' {
     name: string
   }
 
-  class Link extends React.Component<React.PropsWithChildren<LinkProps>, any> {
+  class Link<Component extends Element | ReactElement> extends React.Component<React.PropsWithChildren<LinkProps<Component>> & React.HTMLProps<Component>> {
```
andrewru, line 1785: This is a generic to take into account the `Component` prop. It means that you can do either of:

```javascript
<Link<HTMLAnchorElement> href="something">Text</Link>
```

or:

```javascript
<Link<Button> Component={<Button />}>Text</Link>
```

josa42 (approved):

PR #1098, opened 2022-04-25
improve a11y color contrast lozenge and badges

tomru:
Many colored `Lozenge` and `Badges` have had a bad color contrast. 

- 27 Lozenge colors scored below WCAG 2.0 Level AA
- Badges suffered from the same, although they support less colors

This PR sets the font color to either `--ui-color-carbon`, `--ui-color-carbon--dark` or `--ui-color-carbon--darker` or `--color-ui-black` whichever meets the Level AA color contrast first or improves it at all.

With this PR, both Lozenge and Badges pass Level AA for all colors, but following exceptions:
- `twitter`, `facebook` and `reddit` - font color is part of the brand, so I did not change them
- `sentiment-positive`, `sentiment-negative`, `paradise-lost`, `electric-dreams` - any dark font will make contrast worse, but still below 4.5
- `space-invaders` - with black better contrast than before, but still below 4.5

To easily see the changes side-by-side see [before](https://1vcz3h.csb.app/) and [after](https://3qers5.csb.app/) the change


packages/axiom-components/src/Badge/Badge.js
```diff
@@ -73,7 +73,6 @@ export default class Badge extends Component {
     } = this.props;
 
     const isFaded = color === "faded";
-    const textColor = isFaded || o < 1 ? undefined : "night";
```
tomru, line 76: `night` seems wrong to me, as regardless which theme is active the font color needs to accommodate to the badge background color, so defaulted to `white-noise` in css instead.

packages/axiom-components/src/Badge/Badge.js
```diff
@@ -89,10 +88,10 @@ export default class Badge extends Component {
         {...rest}
         Component="span"
         className={classes}
+        style={style}
```
tomru, line 91: Not entirely sure what the faded use-case is, but if a Badge is faded, it seems to make most sense to fade both background and text color.

Flawwles:
Nice! Loads better. 
Thanks for fixing this. Shame about a few still missing the mark, but I think we need to really rethink the badges/lozenge and text colour as a larger project, so this is a great quick improvement.
One comment about the fading the text, but other than that, looks good to me. 👍

packages/axiom-components/src/Badge/Badge.js
Flawwles: This is the example of where faded badges are used: image

tomru: Ah, I got very confused, why the text color is set to `night` in case it's faded and thought this might be an error.

Makes sense to leave that as is, will revert the related code changes then.

tomru: Removed related commits from this PR now :heavy_check_mark:

josa42 (approved):

PR #1105, opened 2022-05-26
feat(Dialog): add ability to set aria attributes.

jrobind:


PTihomir:
Thanks, keep this up so I can learn :D
Two questions. Seeing that You updated AlertDialog, should You also add the new props to the [ChangePasswordDialog](https://github.com/BrandwatchLtd/axiom-react/blob/1dedc2c1d4e734b25a9a24757a94ab115978f7fa/packages/axiom-composites/src/ChangePasswordDialog/ChangePasswordDialog.js)
Will these props appear in the storybook page? I think that would be super useful. Not sure how much others use that, but I almost always check https://axiom-storybook.brandwatch.com/ when want to use an axiom component.

There is also this page, maybe worth expanding it with what you added https://axiom.brandwatch.com/7dde06551/p/644a03-dialog/t/39fe6e

cdrapes (approved):
👍  nice!

jrobind:
> Thanks, keep this up so I can learn :D Two questions. Seeing that You updated AlertDialog, should You also add the new props to the [ChangePasswordDialog](https://github.com/BrandwatchLtd/axiom-react/blob/1dedc2c1d4e734b25a9a24757a94ab115978f7fa/packages/axiom-composites/src/ChangePasswordDialog/ChangePasswordDialog.js) Will these props appear in the storybook page? I think that would be super useful. Not sure how much others use that, but I almost always check https://axiom-storybook.brandwatch.com/ when want to use an axiom component.
> 
> There is also this page, maybe worth expanding it with what you added https://axiom.brandwatch.com/7dde06551/p/644a03-dialog/t/39fe6e

Thanks @PTihomir! I've just updated the `ChangePasswordDialog` to include the new props – good spot! And yes, these new props will appear on the storybook pages for `AlertDialog` & `Dialog`. They are visible on the latest PR deploy: https://629507a55886d908d78be9c4--axiom-react.netlify.app/.

Good shout regarding the docs updates, I'll raise a separate PR for that 👍

PTihomir (approved):
Nice work @jrobind !

josa42 (approved):
Thanks for the Updates!

